PR #14, opened 2018-09-23
MAKE-470: pail implementations for GRIDFS and LOCAL

tychoish:


pail/util.go
```diff
@@ -0,0 +1,69 @@
...
+func sha1sum(path string) (string, error) {
+	out, err := checksum(sha1.New(), path)
+	return out, errors.WithStack(err)
+}
+
+func walkLocalTree(ctx context.Context, prefix string) ([]string, error) {
+	var out []string
+	err := filepath.Walk(prefix, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return nil
```
bsamek: Seems like this should return an error.

tychoish: It does, but that doesn't work because listing empty buckets shouldn't be an error, and it's fine to just skip bad files, I think.

PR #16, opened 2018-10-11
MAKE-468: s3 backed implementation of the pail bucket interface

julianedwards:
Auth is still missing (going to create a separate ticket for that), and I need to fix the Copy method as we discussed offline (should that also be a separate ticket?).

### DIFF
--- a/pail/bucket_test.go
+++ b/pail/bucket_test.go
@@ -13,6 +13,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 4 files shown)


pail/bucket_test.go
```diff
@@ -25,6 +29,67 @@ func newUUID() string {
...
+func createS3Client(region string) (*s3.S3, error) {
+	sess, err := session.NewSession(&aws.Config{Region: aws.String(region)})
+	if err != nil {
+		return nil, errors.Wrap(err, "problem connecting to AWS")
+	}
+	svc := s3.New(sess)
+	return svc, nil
+}
+
+func createS3Bucket(name, region string) error {
```
tychoish: I think it would be reasonable/good to not create new buckets for the test and instead just use existing buckets. I don't think its reasonable to think that our tests would have keys that could create buckets.

pail/interface.go
```diff
@@ -77,15 +77,11 @@ type Bucket interface {
...
-
-	Auth   string
-	Region string
-	Name   string
-	Access string // for puts only, presumably.
+type CopyOptions struct {
+	SrcKey     string
+	DestKey    string
+	DestBucket Bucket
+	Dest       bool
```
tychoish: I would spell out these key names entirely.

pail/interface.go
```diff
@@ -123,6 +121,7 @@ type bucketItemImpl struct {
 }
 
 func (bi *bucketItemImpl) Name() string   { return bi.key }
+func (bi *bucketItemImpl) Hash() string   { return bi.hash }
```
tychoish: What populates hash values for local and gridfs impls?

pail/s3_bucket.go
```diff
@@ -0,0 +1,551 @@
...
+	svc    *s3.S3
+}
+
+type S3Options struct {
+	// If do not provide credentials, fall back to AWS default provider chain.
+	Credentials *credentials.Credentials
+	Region      string
+	Name        string
+	Prefix      string
+	Access      string // for puts only, presumably.
```
tychoish: the comments on these structs should probably removed.

pail/s3_bucket.go
```diff
@@ -0,0 +1,551 @@
...
+}
+
+type s3Bucket struct {
+	name   string
+	prefix string
+	sess   *session.Session
+	svc    *s3.S3
+}
+
+type S3Options struct {
```
tychoish: It's a good practice on public structs and functions to write a docstring.

pail/s3_bucket.go
```diff
@@ -0,0 +1,551 @@
...
+	}
+	sess, err := session.NewSession(config)
+	if err != nil {
+		return &s3Bucket{}, errors.Wrap(err, "problem connecting to AWS")
+	}
+	svc := s3.New(sess)
+	return &s3Bucket{name: s3Options.Name, prefix: s3Options.Prefix, sess: sess, svc: svc}, nil
+}
+
+func NewS3BucketSmall(s3Options S3Options) (Bucket, error) {
```
tychoish: can these functions have docstrings, and perhaps name them "S3Bucket" and "S3MultiPartBucket"

pail/s3_bucket.go
```diff
@@ -0,0 +1,551 @@
...
+		name: s.name,
+		svc:  s.svc,
+		ctx:  ctx,
+		key:  s.normalizeKey(key),
+	}, nil
+}
+func (s *s3BucketLarge) Writer(ctx context.Context, key string) (io.WriteCloser, error) {
+	// 5MB is the minimum size for a multipart upload, so buffer needs to be at least that big.
+	return &largeWriteCloser{
+		maxSize: 5000000,
```
tychoish: maybe make this a named constant? or use a constant from the aws package?

pail/s3_bucket.go
```diff
@@ -0,0 +1,551 @@
...
+			return errors.Wrapf(err, "problem checksumming '%s'", file)
+		}
+		input := &s3.HeadObjectInput{
+			Bucket:  aws.String(s.name),
+			Key:     aws.String(target),
+			IfMatch: aws.String(localmd5),
+		}
+		_, err = s.svc.HeadObjectWithContext(ctx, input)
+		if aerr, ok := err.(awserr.Error); ok {
+			if aerr.Code() == "PreconditionFailed" || aerr.Code() == "NotFound" {
```
tychoish: are these strings constants?

pail/interface.go
julianedwards: They aren't using `hash`, I added it for s3 in order to avoid making two separate API calls since we get the `ETag` for free when we iterate with the List function. The hashes are calculated using `md5sum` and `sha1sum` helper functions you created.

pail/s3_bucket.go
julianedwards: Unfortunately, there are only two error code constants in the s3 go sdk, the rest are generic strings:

> In this example the S3 GetObject API operation is used to request the contents of an object in S3. The example handles the NoSuchBucket and NoSuchKey error codes, printing custom messages to stderr. If any other error is received, a generic message is printed.

https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/handling-errors.html

pail/s3_bucket.go
julianedwards: Added.

evergreen.yaml
```diff
@@ -46,6 +46,22 @@ variables:
...
+    type: setup
+    params:
+       silent: true
+       script: |
+         mkdir ~/.aws
+         cat <<EOF > ~/.aws/config
+         [default]
+         region = us-east-1
+         EOF
+         cat <<EOF > ~/.aws/credentials
```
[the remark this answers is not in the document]
julianedwards: added.

PR #17, opened 2018-11-02
MAKE-499 implement new rollup model

ablack12:
@tychoish


model/perf_data.go
```diff
@@ -36,18 +38,163 @@ type PerfRollups struct {
...
-func (r *PerfRollups) Setup(env sink.Environment)                            { r.env = env }
-func (r *PerfRollups) Add(name string, version int, value interface{}) error {}
-func (r *PerfRollups) GetInt(name string) (int, error)                       {}
-func (r *PerfRollups) GetInt32(name string) (int32, error)                   {}
-func (r *PerfRollups) GetInt64(name string) (int64, error)                   {}
-func (r *PerfRollups) GetFloat(name string) (float64, error)                 {}
-func (r *PerfRollups) Validate() error                                       {}
-func (r *PerfRollups) Map() map[string]int64                                 {}
-func (r *PerfRollups) MapFloat() map[string]float64                          {}
+type record struct {
```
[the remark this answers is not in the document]
ablack12: Well I need a struct of this layout to unmarshal results from Find. I can define it locally instead if you want but this way I can also use it in the test cases without redefining, but I can do that instead if it's better. I'll change the name regardless.

model/perf_data.go
```diff
@@ -36,18 +38,163 @@ type PerfRollups struct {
...
+		"rollups.name": name,
+	}
+
+	out := record{}
+	rollup := PerfRollupValue{
+		Name:    name,
+		Version: version,
+		Value:   value,
+	}
+	err = c.Find(search).Select(bson.M{"rollups.version": 1, "rollups.name": 1}).One(&out)
```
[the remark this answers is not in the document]
ablack12: Should we be updating r.populated to true somewhere else or?

ablack12:
All the evergreen tests are failing because sthree isn't available. Do we ignore this for now since it's unrelated and this is just a basic model? Or is there a way to skip it?

model/perf_data.go
```diff
@@ -36,18 +50,171 @@ type PerfRollups struct {
...
-func (r *PerfRollups) Setup(env sink.Environment)                            { r.env = env }
-func (r *PerfRollups) Add(name string, version int, value interface{}) error {}
-func (r *PerfRollups) GetInt(name string) (int, error)                       {}
-func (r *PerfRollups) GetInt32(name string) (int32, error)                   {}
-func (r *PerfRollups) GetInt64(name string) (int64, error)                   {}
-func (r *PerfRollups) GetFloat(name string) (float64, error)                 {}
-func (r *PerfRollups) Validate() error                                       {}
-func (r *PerfRollups) Map() map[string]int64                                 {}
-func (r *PerfRollups) MapFloat() map[string]float64                          {}
+type perfRollupEntries struct {
```
tychoish: I think this might be better as a type alias rather than a struct....

model/perf_data.go
```diff
@@ -36,18 +50,171 @@ type PerfRollups struct {
...
+				return errors.New("outdated version")
+			}
+			break
+		}
+	}
+	update := bson.M{
+		newRollupVal: value,
+		newRollupVer: version,
+	}
+	err2 := c.Update(search, bson.M{setExistingRollup: update})
```
tychoish: why not just use err here?

model/perf_data.go
```diff
@@ -6,6 +6,20 @@ import (
...
+
+const (
+	idField           = "_id"
+	rollupsField      = "rollups"
+	rollupNameField   = "rollups.name"
+	rollupVerField    = "rollups.version"
+	newRollupVal      = "rollups.$.val"
+	newRollupVer      = "rollups.$.version"
+	pushNewRollup     = "$push"
+	setExistingRollup = "$set"
```
tychoish: I think we should probably not store constants for the dotted field names and just assemble them at runtime. I think the _id field is probably stored in a var elsewhere, and I'm sort of on the fence about storing $ operators in constants because it probably makes the code harder to read, but I could be convinced.

model/perf_data.go
ablack12: I think this needs to be a struct because One will not marshal into a list.

tychoish: I'd just do something with anon structs for the purpose of unmarshalling. you can also just use the larger model and the decoder will just leave the fields empty.

model/perf_data.go
```diff
@@ -36,18 +50,171 @@ type PerfRollups struct {
...
+		return val, nil
+	}
+	return 0, errors.Errorf("mismatched type for name %s", v.Name)
+}
+
+func (r *PerfRollups) Setup(env sink.Environment) {
+	r.env = env
+}
+
+func (r *PerfRollups) Add(name string, version int, value interface{}) error {
```
tychoish: should we split up this function to make it a bit easier to test?

ablack12:
Remaining tests (besides lint, which is failing from all of the placeholders) are failing because of this error in ftdc:
`time.Since(startAt).Round undefined (type time.Duration has no field or method Round)`
Do we want to address this? Or make a new ticket for it?

tychoish:
I believe this is fixed in the tip of master. rebase/update and see if that helps

ablack12:
This looks good now, sorry for the thousands of commits.

PR #19, opened 2018-11-09
MAKE-508: build restmodels package

julianedwards:
Package build, implementation completed for `PerformanceResult` struct. Skipped implementation for `Export`.

### DIFF
--- a/rest/model/model.go
+++ b/rest/model/model.go
@@ -3,6 +3,6 @@ package model
 // Model defines how to transform to and from an API model for a given
 // interface.
 type Model interface {
-	Export(interface{}) error
 	Import(interface{}) error
+	Export(interface{}) (interface{}, error)
 }
--- a/rest/model/model.go
+++ b/rest/model/model.go
@@ -1,0 +1,8 @@
+package model
+
+// Model defines how to transform to and from an API model for a given
+// interface.
+type Model interface {
+	Export(interface{}) error
+	Import(interface{}) error
+}
--- a/rest/model/model.go
+++ b/rest/model/model.go
@@ -4,5 +4,5 @@ package model
 // interface.
 type Model interface {
 	Import(interface{}) error
-	Export(interface{}) (interface{}, error)
+	Export() (interface{}, error)
 }
--- a/rest/model/perf.go
+++ b/rest/model/perf.go
@@ -24,48 +24,34 @@ func (apiResult *APIPerformanceResult) Import(i interface{}) error {
 		apiResult.CreatedAt = NewTime(r.CreatedAt)
 		apiResult.CompletedAt = NewTime(r.CompletedAt)
 		apiResult.Version = r.Version
+		apiResult.Info = getPerformanceResultID(r.Info)
 
-		apiInfo, err := getPerformanceResultID(r.Info)
-		if err != nil {
-			return errors.Wrap(err, "problem getting info")
-		}
-		apiResult.Info = apiInfo
+		total := getPerformancePoint(r.Total)
+		apiResult.Total = &total
+		rollups := getPerfRollups(r.Rollups)
+		apiResult.Rollups = &rollups
 
 		var apiSource []APIArtifactInfo
 		for _, artifactInfo := range r.Source {
-			apiArtifactInfo, err := getArtifactInfo(artifactInfo)
-			if err != nil {
-				return errors.Wrap(err, "problem getting source data")
-			}
+			apiArtifactInfo := getArtifactInfo(artifactInfo)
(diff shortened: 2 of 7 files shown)


rest/model/perf_test.go
```diff
@@ -0,0 +1,514 @@
...
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	dbmodel "github.com/evergreen-ci/sink/model"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestImportHelperFunctions(t *testing.T) {
```
julianedwards: This might be overkill, but I wrote this before writing `TestImport` and realized that I am testing the same thing there. I do like separating these out into unit tests, but up to you as to whether we should keep these or if the tests in`TestImport` is sufficient.

rest/model/perf.go
```diff
@@ -0,0 +1,200 @@
...
+
+type APIPerformanceResultID struct {
+	Project   APIString        `json:"project"`
+	Version   APIString        `json:"version"`
+	TaskName  APIString        `json:"task_name"`
+	TaskID    APIString        `json:"task_id"`
+	Execution int              `json:"execution"`
+	TestName  APIString        `json:"test_name"`
+	Trial     int              `json:"trial"`
+	Parent    APIString        `json:"parent"`
```
julianedwards: Shouldn't this be a pointer to another struct? 
From the [supplemental design doc|https://docs.google.com/document/d/1Bi-278IaeAwc5RCFal8Ww_lKMWUGVeDRLZhG8oYqJTM/edit#] you wrote:
{
    "project": <evergreen project>,
    "version": <evergreen version id>,
    "task_id": <evergreen task id>,
    "task_name": <task display name>,
    "execution": <int>,     // evergreen execution id
    "test_name": <string>,  // arbitrary, describes a unique experiment/value
    "trial": <int>,         // repeated execution
    "parent": <infoHash>,   // pointer to parent test
    "tags": [...],          // arbitrary test-supplied metadata
    "args" {},              // map of strings to ints as test parameters
    "events": [...],        // 0 or more Events
}

rest/model/model.go
```diff
@@ -0,0 +1,8 @@
+package model
+
+// Model defines how to transform to and from an API model for a given
+// interface.
+type Model interface {
+	Import(interface{}) error
+	Export(interface{}) (interface{}, error)
```
tychoish: I think that export doesn't need to take an argument? or at least there should be some more documentation about how this would get used.

rest/model/perf.go
```diff
@@ -0,0 +1,200 @@
...
+
+type APIPerformanceResult struct {
+	Name         APIString              `json:"name"`
+	Info         APIPerformanceResultID `json:"info"`
+	CreatedAt    APITime                `json:"create_at"`
+	CompletedAt  APITime                `json:"completed_at"`
+	Version      int                    `json:"version"`
+	Source       []APIArtifactInfo      `json:"source_info"`
+	AuxilaryData []APIArtifactInfo      `json:aux_data"`
+	Total        *APIPerformancePoint   `json:"total"`
```
tychoish: Why are these pointers?

rest/model/perf.go
tychoish: I think this is correct as written.

rest/model/perf.go
```diff
@@ -0,0 +1,200 @@
...
+	Project   APIString        `json:"project"`
+	Version   APIString        `json:"version"`
+	TaskName  APIString        `json:"task_name"`
+	TaskID    APIString        `json:"task_id"`
+	Execution int              `json:"execution"`
+	TestName  APIString        `json:"test_name"`
+	Trial     int              `json:"trial"`
+	Parent    APIString        `json:"parent"`
+	Tags      []string         `json:"tags"`
+	Arguments map[string]int32 `json:"args"`
```
tychoish: why int32 and not int?

rest/model/perf.go
julianedwards: Because they are pointers in `PerformanceResult` https://github.com/evergreen-ci/sink/blob/master/model/perf.go#L45-L47

rest/model/perf.go
julianedwards: Thanks

rest/model/perf.go
julianedwards: Copied from original `PerformanceResultID` struct https://github.com/evergreen-ci/sink/blob/master/model/perf.go#L133

rest/model/perf.go
tychoish: should probably make it omit empty then.

PR #20, opened 2018-11-13
MAKE-509: Database method to find all perf test results

julianedwards:


model/perf.go
```diff
@@ -288,11 +299,53 @@ func (r *PerformanceResults) findAllChildren(parent string) error {
...
+		project,
+	}
+	pipe := session.DB(conf.DatabaseName).C(perfResultCollection).Pipe(pipeline)
+	iter := pipe.Iter()
+	defer iter.Close()
+
+	type iterDoc struct {
+		Children []PerformanceResult `bson:"children,omitempty"`
+	}
+	doc := iterDoc{}
```
tychoish: you don't need to have this be a named type, and can just do 
```
doc := struct{
     Children Type `bson:"foo"`
}

PR #21, opened 2018-11-14
 MAKE-510 implement default rollups calculations

ablack12:


ablack12:
I see that you've added some summary information. Should I be doing something with these guages and etc? I'm not sure what information we care about, or what guages means..

model/perf_data.go
```diff
@@ -320,25 +319,27 @@ type PerformanceMetricSummary struct {
 // reporting that this application does across test values.
 type PerformanceTimeSeries []PerformancePoint
 
-// Statistics converts a a series into an intermediate format that we
-// can use to calculate means/totals/etc.
-func (ts PerformanceTimeSeries) Statistics() PerformanceStatistics {
-	out := PerformanceStatistics{
+// statistics converts a a series into an intermediate format that we
+// can use to calculate means/totals/etc to create default rollups
+func (ts PerformanceTimeSeries) statistics() (performanceStatistics, error) {
```
tychoish: probably worth returning a pointer here to avoid needless copying and also being able to return nil.

model/perf_data.go
```diff
@@ -320,25 +319,27 @@ type PerformanceMetricSummary struct {
...
+	out.timers.total = make(stats.Float64Data, len(ts))
 
 	var lastPoint time.Time
 
 	for idx, point := range ts {
 		if idx == 0 {
 			lastPoint = point.Timestamp
 		} else if point.Timestamp.Before(lastPoint) {
-			panic("data is not valid")
+			return out, errors.New("data is not valid")
```
tychoish: return nil? or an empty struct in cases where there's problem.

model/perf_data.go
```diff
@@ -369,102 +369,194 @@ func (ts PerformanceTimeSeries) Statistics() PerformanceStatistics {
...
+func (perf *performanceStatistics) percentile(pval float64) (performanceMetricSummary, error) {
 	var err error
 	catcher := grip.NewBasicCatcher()
-	out := PerformanceMetricSummary{
-		Samples: perf.samples,
-		Span:    perf.span,
+	out := performanceMetricSummary{
+		samples:    perf.samples,
+		span:       perf.span,
+		metricType: fmt.Sprintf("percentile_%d", int(pval)),
```
tychoish: you shouldn't need to type cast to int.... use %f (I think.)

ablack12: if I use %f it comes out as like "90.00000" which I think is aggressive personally.

bsamek: You can specify precision with `%f`. You can e.g. `%.2f` if you want 90.00, or `%.f` if you want 90.

PR #22, opened 2018-11-19
MAKE-511: add grpc method to have user-submitted rollups

julianedwards:
also added testing for other existing grpc methods


perf.proto
```diff
@@ -89,23 +90,42 @@ message MetricsCounters {
...
 message MetricsEvent {
   string id = 1;
   repeated MetricsPoint Event = 3;
 }
 
+message RollupValue {
+  string name = 1;
+  oneof value {
+    int64 int = 2;
+    float fl = 3;
```
tychoish: I think we want to make sure this is a float64 (and its currently a float32

model/perf.go
```diff
@@ -73,6 +73,10 @@ func CreatePerformanceResult(info PerformanceResultInfo, source []ArtifactInfo)
 func (result *PerformanceResult) Setup(e sink.Environment) { result.env = e }
 func (result *PerformanceResult) IsNil() bool              { return !result.populated }
 func (result *PerformanceResult) Find() error {
+	// Need to maintain anything that gets lost after populate result
+	//
+	env := result.env
```
tychoish: why?

model/perf.go
```diff
@@ -86,6 +90,7 @@ func (result *PerformanceResult) Find() error {
 	} else if err != nil {
 		return errors.Wrap(err, "problem finding result config")
 	}
+	result.Setup(env)
```
tychoish: this can't possibly do what we mean? probably omit.

perf.proto
```diff
@@ -89,23 +90,42 @@ message MetricsCounters {
...
+  }
+  int64 version = 4;
+}
+
+message Rollups {
+  repeated RollupValue default_stats = 1;
+  repeated RollupValue user_stats = 2;
```
tychoish: I merged these. rebase and you can make this go away

perf.proto
```diff
@@ -89,23 +90,42 @@ message MetricsCounters {
...
   repeated MetricsPoint Event = 3;
 }
 
+message RollupValue {
+  string name = 1;
+  oneof value {
+    int64 int = 2;
+    float fl = 3;
+  }
+  int64 version = 4;
```
tychoish: I added a user bool to replace the UserStats.

perf.proto
```diff
@@ -89,23 +90,42 @@ message MetricsCounters {
...
+  google.protobuf.Timestamp processed_at = 3;
+  int64 count = 4;
+  bool valid = 5;
+}
+
 service SinkPerformanceMetrics {
   rpc CreateMetricSeries(ResultData) returns (MetricsResponse);
   rpc AttachResultData(ResultData) returns (MetricsResponse);
   rpc AttachAuxilaryData(ResultData) returns (MetricsResponse);
+  rpc AttachRollups(ResultData) returns (MetricsResponse);
```
tychoish: maybe we want a message type that just wraps rollups (so you can associate an ID with it.) ? Otherwise everything takes the same type?

model/perf.go
julianedwards: It's a bug not to restore env, cause then you manipulate `result` and `env` is missing. So if I call `result.Find()`, populate result with some additional data (such as aux data), then call result.Save() it will fail because the `env` field is nil.

model/perf.go
julianedwards: nope, it is intentional

julianedwards: `AttachResultData`, `AttachAuxilaryData`, and `AttachRollups` all fail without restoring `env`.

perf.proto
julianedwards: I still see `DefaultStats` and `UserStats` after rebase

perf.proto
julianedwards: So I was just following the signatures of the rest of these functions, if we do that then shouldn't we also do it for AttachAuxilaryData?

perf.proto
julianedwards: ?

perf.proto
tychoish: yes probably.

julianedwards: If we do this then we expect the user to have access to the ID (which is created by hashing and combining the fields in PerformanceResultInfo). Is that expected behavior?

tychoish: Yeah, because it's included in the CreateMetricSeries response.

rpc/internal/service_test.go
```diff
@@ -0,0 +1,328 @@
...
+	"github.com/stretchr/testify/require"
+	grpc "google.golang.org/grpc"
+	mgo "gopkg.in/mgo.v2"
+)
+
+const (
+	address = "localhost:50051"
+)
+
+type MockEnv struct {
```
tychoish: isn't there a mock environment in the mock package at the top level?

tychoish: this is fine after all, I'm remembering a different package.

rpc/internal/service.go
```diff
@@ -68,38 +68,68 @@ func (srv *perfService) AttachResultData(ctx context.Context, result *ResultData
 		record.Artifacts = append(record.Artifacts, *i.Export())
 	}
 
+	if result.Rollups != nil {
+		rollups, err := result.Rollups.Export()
+		if err != nil {
+			return nil, errors.Wrap(err, "problem getting rollups")
+		}
+		record.Rollups = &rollups
+	}
```
tychoish: is there are rollups in the results, shouldn't we do the Add thing?

rpc/internal/service.go
```diff
@@ -64,41 +64,71 @@ func (srv *perfService) AttachResultData(ctx context.Context, result *ResultData
...
+
+	resp := &MetricsResponse{}
+	resp.Id = record.ID
+
+	rollups, err := rollupData.Rollups.Export()
+	if err != nil {
+		return nil, errors.Wrap(err, "problem getting rollups")
+	}
+	record.Rollups = &rollups
```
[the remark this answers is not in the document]
tychoish: I think this is still relevant.

rpc/internal/service.go
julianedwards: Created `addRollups` helper function

rpc/internal/service.go
```diff
@@ -208,3 +207,16 @@ func (srv *perfService) CloseMetrics(ctx context.Context, end *MetricsSeriesEnd)
 
 	return nil, nil
 }
+
+func addRollups(record *model.PerformanceResult, rollups *model.PerfRollups) {
```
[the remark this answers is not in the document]
julianedwards: https://jira.mongodb.org/browse/MAKE-552

PR #23, opened 2018-11-27
MAKE-536: Implement Connector for Perf Results

julianedwards:


rest/data/interface.go
```diff
@@ -0,0 +1,14 @@
...
+	"github.com/evergreen-ci/sink/model"
+	"github.com/evergreen-ci/sink/util"
+)
+
+type Connector interface {
+	// PerformanceResult
+	FindPerformanceResultById(string) (*model.PerformanceResult, error)
+	FindPerformanceResultsByTaskId(string, util.TimeRange, ...string) ([]model.PerformanceResult, error)
+	FindPerformanceResultsByVersion(string, util.TimeRange, ...string) ([]model.PerformanceResult, error)
+	FindPerformanceResultWithChildren(string, int, ...string) ([]model.PerformanceResult, error)
```
tychoish: can these return the rest model rather than the db model?

rest/data/perf.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/evergreen-ci/sink"
+	"github.com/evergreen-ci/sink/model"
+	"github.com/evergreen-ci/sink/util"
+)
+
+// DBPerformanceResultConnector is a struct that implements the Perf
+// related functions from the Connector through interactions with the backing
+// database.
+type DBPerformanceResultConnector struct {
+	env *sink.Environment
```
tychoish: sink.Environment is an interface so you don't need the direction.

rest/data/impl.go
```diff
@@ -0,0 +1,23 @@
...
+// without forcing changes to the API.
+type DBConnector struct {
+	env *sink.Environment
+
+	DBPerformanceResultConnector
+}
+
+func CreateDBConnector(env sink.Environment) Connector {
+	return &DBConnector{
+		env: &env,
```
tychoish: this shouldn't be a pointer.

rest/data/interface.go
julianedwards: done

model/perf.go
```diff
@@ -275,8 +278,7 @@ func (r *PerformanceResults) createFindQuery(options PerfFindOptions) map[string
 		search[bsonutil.GetDottedKeyName("info", "schema")] = options.Info.Schema
 	}
 	if len(options.Info.Tags) > 0 {
-		search[bsonutil.GetDottedKeyName("info", "tags")] =
-			db.Document{"$in": options.Info.Tags}
+		search[bsonutil.GetDottedKeyName("info", "tags")] = db.Document{"$in": options.Info.Tags}
```
tychoish: it's a little weird that we're using db.Document and bson.M in the same file. they're both map[string]interface{}

rest/data/impl.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+// DBConnector is a struct that implements all of the methods which connect to
+// the service layer of sink. These methods abstract the link between the
+// and the API layers, allowing for changes in the service architecture
+// without forcing changes to the API.
+type DBConnector struct {
+	env sink.Environment
+
+	DBPerformanceResultConnector
```
tychoish: maybe we can just store put all of the methods on DBConnector?

PR #24, opened 2018-11-30
MAKE-534: methods for FTDC chunks to calculate rollups/means/totals from the chunks

julianedwards:


perf/ftdc_rollups.go
```diff
@@ -0,0 +1,207 @@
...
+			case "Timers.Total":
+				perfStats.timers.total = time.Duration(metric.Values[len(metric.Values)-1])
+			case "Gauges.State":
+				perfStats.gauges.stateTotal += sum(metric.Values)
+			case "Gauges.Workers":
+				perfStats.gauges.workersTotal += sum(metric.Values)
+			case "Gauges.Failed":
+				perfStats.gauges.failedTotal += sum(metric.Values)
+			default:
+				return performanceStatistics{}, errors.New(fmt.Sprintf("unknown field name %s", name))
```
tychoish: errors.Errorf

perf/ftdc_rollups.go
```diff
@@ -0,0 +1,207 @@
...
+			Value:         float64(s.gauges.stateTotal) / float64(s.numSamples),
+			Version:       defaultVer,
+			UserSubmitted: false,
+		})
+		rollups = append(rollups, model.PerfRollupValue{
+			Name:          "avgWorkers",
+			Value:         float64(s.gauges.workersTotal) / float64(s.numSamples),
+			Version:       defaultVer,
+			UserSubmitted: false,
+		})
```
tychoish: append is a variadic function so you can do this in one call.

perf/ftdc_rollups.go
```diff
@@ -0,0 +1,207 @@
...
+		Value:         s.counters.size,
+		Version:       defaultVer,
+		UserSubmitted: false,
+	})
+	rollups = append(rollups, model.PerfRollupValue{
+		Name:          "totalSamples",
+		Value:         s.numSamples,
+		Version:       defaultVer,
+		UserSubmitted: false,
+	})
```
tychoish: merge.

perf/ftdc_rollups.go
```diff
@@ -0,0 +1,207 @@
...
+	return rollups
+}
+
+func sum(nums []int64) int64 {
+	sum := int64(0)
+	for _, num := range nums {
+		sum += num
+	}
+	return sum
```
tychoish: can we just move this to util or use something from stats.

perf/ftdc_rollups.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/evergreen-ci/sink/model"
+	"github.com/mongodb/ftdc"
+	"github.com/pkg/errors"
+)
+
+const (
+	defaultVer = 1
+)
+
+type performanceStatistics struct {
```
tychoish: can we merge this with the one in model?

tychoish: or maybe just use events?

julianedwards: Once events.Performance has duration as cumulative we can just use that.

tychoish: let's open a ticket and do this in the next sprint.

PR #43, opened 2019-01-15
MAKE-251: add authentication to cedar application

julianedwards:


config.go
```diff
@@ -15,21 +16,25 @@ type Configuration struct {
 	MongoDBDialTimeout time.Duration
 	UseLocalQueue      bool
 	NumWorkers         int
+	UserManager        gimlet.UserManager
```
tychoish: we can't really put this here... interfaces don't seralize well, and this config object is really about bootstrapping the thing in model.Config

environment.go
```diff
@@ -23,6 +25,8 @@ func resetEnv() { globalEnv = &envState{name: "global", conf: &Configuration{}}
 type Environment interface {
 	Configure(*Configuration) error
 
+	AddAuth(gimlet.UserManager, time.Duration)
```
tychoish: expire after should be configurable in the DB config struct.

tychoish: (also in general, given a config struct or a subset of it, the production of a user manager should be lightweight and not the kind of thing that we need to cache like this.

operations/service.go
```diff
@@ -108,7 +110,33 @@ func Service() cli.Command {
...
+					middlewareConf,
+				)
+				stream := aviation.MakeAuthenticationRequiredStreamingInterceptor(
+					cedarConf.UserManager,
+					middlewareConf,
+				)
+				rpcSrv = grpc.NewServer(
+					grpc.UnaryInterceptor(unary),
+					grpc.StreamInterceptor(stream),
+				)
```
tychoish: I'd just merge all these statements and do it inline,

operations/service.go
```diff
@@ -108,7 +110,33 @@ func Service() cli.Command {
 			//
 			// starting grpc
 			//
-			rpcSrv := grpc.NewServer()
+			middlewareConf := gimlet.UserMiddlewareConfiguration{
+				// TODO: figure out correct input for these
+				CookieName:     "cookieToken",
```
tychoish: mirror what MCI uses

rest/data/perf.go
```diff
@@ -48,8 +48,7 @@ func (dbc *DBConnector) FindPerformanceResultsByTaskId(taskId string, interval u
 			TaskID: taskId,
 			Tags:   tags,
 		},
-		// set MaxDepth to -1 to avoid recursive child search
-		MaxDepth: -1,
```
tychoish: doesn't this revert a thing that you changed last week?

julianedwards: No I forgot to change this with those changes (remember -1 now means to recurse without limit and 0 means do not recurse at all)

operations/service.go
julianedwards: done

PR #45, opened 2019-01-17
MAKE-546: Begin to keep track of what indexes we need to build/maintain

julianedwards:


model/indexes.go
```diff
@@ -0,0 +1,52 @@
+package model
+
+// +build ignore
```
tychoish: I think this needs to be before the package line?

model/indexes.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"github.com/mongodb/anser/bsonutil"
+	"gopkg.in/mgo.v2/bson"
+)
+
+// Indexes for "perf_results" collection.
+const (
+	PerfResultsCreatedCompletedIndex = bson.D{{perfCreatedAtKey, 1}, {perfCompletedAtKey, 1}}
```
tychoish: I was thinking something more like: 

type SystemIndexes struct {
      Spec bson.D
      Options bson.D
      Collection string
}

and then a function that just returns a slice of those objects with the data that you have here.

PR #48, opened 2019-01-23
MAKE-622: Test cedar service send

julianedwards:


rpc/internal/service_test.go
```diff
@@ -316,3 +325,101 @@ func TestAttachResultData(t *testing.T) {
...
+	curatorPath, err := filepath.Abs("curator")
+	require.NoError(t, err)
+	for _, test := range []struct {
+		name    string
+		skip    bool
+		local   bool
+		setUp   func(t *testing.T)
+		cleanUp func(t *testing.T)
+		closer  func(t *testing.T)
+		cmd     *exec.Cmd
```
tychoish: refactor to parameterize this as a function.

rpc/internal/service_test.go
```diff
@@ -316,3 +325,101 @@ func TestAttachResultData(t *testing.T) {
...
+
+	curatorPath, err := filepath.Abs("curator")
+	require.NoError(t, err)
+	for _, test := range []struct {
+		name    string
+		skip    bool
+		local   bool
+		setUp   func(t *testing.T)
+		cleanUp func(t *testing.T)
+		closer  func(t *testing.T)
```
tychoish: I think you can probabaly merge cleanup and closer, but it means that your functions need to user require.

PR #50, opened 2019-01-24
MAKE-622: Test cedar service send (part 2)

julianedwards:


rest/service.go
```diff
@@ -141,6 +141,7 @@ func (s *Service) addRoutes() {
 	s.app.AddRoute("/depgraph/{id}/edges").Version(1).Get().Wrap(checkUser).Handler(s.getDepGraphEdges)
 
 	s.app.AddRoute("/perf/{id}").Version(1).Get().RouteHandler(makeGetPerfById(s.sc))
+	s.app.AddRoute("/perf/remove/{id}").Version(1).Get().Wrap(checkUser).RouteHandler(makeRemovePerfById(s.sc))
```
tychoish: this should be /perf/{id} but the delete method.

rest/perf_routes.go
```diff
@@ -51,6 +51,44 @@ func (h *perfGetByIdHandler) Run(ctx context.Context) gimlet.Responder {
...
+}
+
+// Run calls the data RemovePerformanceResultById function and returns the
+// error.
+func (h *perfRemoveByIdHandler) Run(ctx context.Context) gimlet.Responder {
+	err := h.sc.RemovePerformanceResultById(h.id)
+	if err != nil {
+		return gimlet.MakeJSONErrorResponder(errors.Wrapf(err, "Error removing performance result by id '%s'", h.id))
+	}
+	return gimlet.NewJSONResponse(nil)
```
tychoish: nit: use struct{}{} rather than nil

model/perf.go
```diff
@@ -132,6 +133,27 @@ func (result *PerformanceResult) Save() error {
 	return errors.Wrap(err, "problem saving perf result to collection")
 }
 
+func (result *PerformanceResult) Remove() error {
```
tychoish: I think we probably always want to remove all of the children

model/perf.go
```diff
@@ -133,25 +132,39 @@ func (result *PerformanceResult) Save() error {
...
+	children := PerformanceResults{env: result.env}
+	if err = children.findAllChildrenGraphLookup(result.ID, -1, []string{}); err != nil {
+		return 0, errors.Wrap(err, "problem getting children to remove")
 	}
-	return nil
+
+	ids := []string{result.ID}
+	for _, res := range children.Results {
+		ids = append(ids, res.ID)
+	}
```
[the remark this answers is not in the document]
julianedwards: there is still the root id that needs to be deleted

PR #57, opened 2019-01-30
MAKE-617: log directly to systemd

tychoish:


operations/setup.go
```diff
@@ -16,77 +18,117 @@ const (
...
+	}
+
+	if err = grip.SetSender(sender); err != nil {
+		return errors.WithStack(err)
+	}
+
+	return nil
+}
+
+func newServiceConf(numWorkers int, localQueue bool, mongodbURI, bucket, dbName string) *serviceConf {
```
julianedwards: this function doesn't seem necessary since it is private

tychoish: just to ease conversion from the old function. I agree and think we can refactor it away, but I didn't want to change that just yet (the tests make things a bit complicated.)

julianedwards: sounds good

PR #59, opened 2019-01-30
MAKE-627: move certificates into the database (cert depot impl)

julianedwards:


depot/impl.go
```diff
@@ -0,0 +1,141 @@
...
+	MongoDBURI           string
+	MongoDBDialTimeout   time.Duration
+	MongoDBSocketTimeout time.Duration
+	DatabaseName         string
+	CollectionName       string
+	ExpireAfter          time.Duration
+}
+
+// Create a new cert depot in the specified MongoDB.
+func NewMongoCertDepot(opts MongoCertDepotOptions) (depot.Depot, error) {
```
tychoish: name this new mgo cert depo, and try and structure it some way that we can have a new driver implementation

depot/impl.go
```diff
@@ -0,0 +1,141 @@
...
+		opts.DatabaseName = "certDepot"
+	}
+	if opts.CollectionName == "" {
+		opts.CollectionName = "certs"
+	}
+	if opts.ExpireAfter <= 0 {
+		opts.ExpireAfter = 30 * 24 * time.Hour
+	}
+
+	session, err := mgo.DialWithTimeout("mongodb://localhost:27017", opts.MongoDBDialTimeout)
```
tychoish: write a variant of this function that takes an mgo.Session

depot/impl.go
```diff
@@ -0,0 +1,141 @@
...
+		return errors.New("data is nil")
+	}
+
+	name := depot.GetNameFromCrtTag(tag)
+	u := &DBUser{
+		ID:   name,
+		Cert: string(data),
+		TTL:  time.Now(),
+	}
+	changeInfo, err := m.session.DB(m.databaseName).C(m.collectionName).UpsertId(name, u)
```
tychoish: session clone and defer closes?

depot/impl.go
```diff
@@ -0,0 +1,141 @@
...
+	return errors.Wrap(err, "problem adding data to the database")
+}
+
+// Check returns whether the id/cert pair specified by the tag exists.
+func (m *mongoCertDepot) Check(tag *depot.Tag) bool {
+	name := depot.GetNameFromCrtTag(tag)
+
+	u := &DBUser{}
+	err := m.session.DB(m.databaseName).C(m.collectionName).FindId(name).One(u)
+	grip.WarningWhenf(err != nil && err != mgo.ErrNotFound, "problem looking up '%s' in the database", name)
```
tychoish: assign the conditional to a variable? or have a function? 

also don't we want to log the error value?

tychoish: probably should log more metadata/make this a fields

makefile
```diff
@@ -1,7 +1,7 @@
 # start project configuration
 name := cedar
 buildDir := build
-packages := $(name) rest rest-data rest-model units operations cost model depgraph perf rpc rpc-internal
+packages := $(name) rest rest-data rest-model units operations cost model depgraph perf rpc rpc-internal depot
```
tychoish: maybe we name this certdepot just so it's different than the depot package in certstrap

depot/impl.go
julianedwards: what does it mean exactly to structure for a new driver implementation?

tychoish: I think it's fine as is

certdepot/impl.go
```diff
@@ -0,0 +1,180 @@
...
+
+	"github.com/mongodb/anser/bsonutil"
+	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/message"
+	"github.com/pkg/errors"
+	"github.com/square/certstrap/depot"
+	mgo "gopkg.in/mgo.v2"
+)
+
+type DBUser struct {
```
tychoish: can we call this just User? 

this is just the same model as the other user collection right?

certdepot/impl.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	opts = defaults(opts)
+	return &mongoCertDepot{
+		session:        s,
+		databaseName:   opts.DatabaseName,
+		collectionName: opts.CollectionName,
+		expireAfter:    opts.ExpireAfter,
+	}, nil
+}
```
tychoish: I'd refactor line 52-57 and 67-72 into a private function. 

also see if we want to do any option validation? is there a possibility to error if the options are not valid?

certdepot/impl_test.go
```diff
@@ -0,0 +1,155 @@
...
+	"testing"
+	"time"
+
+	"github.com/square/certstrap/depot"
+	"github.com/stretchr/testify/suite"
+	mgo "gopkg.in/mgo.v2"
+	"gopkg.in/mgo.v2/bson"
+)
+
+type MongoDepotTestSuite struct {
```
tychoish: what kind of coverage does this provide? are there other tests that we'd want? 

would it be possible to have mongoDepot be of the depot interface? I think in response to my earlier question, we'll have to write another implementation of this that uses the new collection, so we'll want to be able to reuse the test suite on an alternate implementation.

certdepot/impl.go
julianedwards: No it is a different kind of user

certdepot/impl.go
julianedwards: error validation done in defaults (I can change the name to validate)

certdepot/impl_test.go
julianedwards: yes it should be of the depot interface, that was my bad. and it covers everything we need right now for this interface I think, the functions are fairly straightforward.

certdepot/impl.go
```diff
@@ -0,0 +1,177 @@
...
+func NewMgoCertDepot(opts MgoCertDepotOptions) (depot.Depot, error) {
+	return newMgoCertDeposit(nil, opts)
+}
+
+// Create a new cert depot in the specified MongoDB, using an existing session.
+func NewMgoCertDepotWithSession(s *mgo.Session, opts MgoCertDepotOptions) (depot.Depot, error) {
+	return newMgoCertDeposit(s, opts)
+}
+
+func validate(opts MgoCertDepotOptions) MgoCertDepotOptions {
```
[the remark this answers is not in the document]
julianedwards: It can never be invalid, we just fill in the defaults. Do you not want as many defaults?

certdepot/impl_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+func TestMongoDepot(t *testing.T) {
+	s := &MongoDepotTestSuite{}
+	s.setup()
+	suite.Run(t, s)
+}
+
+func (s *MongoDepotTestSuite) TearDownTest() {
+	err := s.collection.DropCollection()
+	if err != nil {
```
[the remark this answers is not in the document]
julianedwards: No cause I don't care if it can't find the namespace, just means the collection doesn't exist and we can move on safely.

certdepot/impl_test.go
julianedwards: going to hold off on this, will refactor tests after get this working and in prod

certdepot/impl.go
tychoish: but it's the same model?

certdepot/cert_request.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	if opts.Passphrase != "" {
+		if err = depot.PutEncryptedPrivateKey(opts.Depot, formattedName, key, []byte(opts.Passphrase)); err != nil {
+			return errors.Wrap(err, "problem saving encrypted private key")
+		}
+	} else {
+		if err = depot.PutPrivateKey(opts.Depot, formattedName, key); err != nil {
+			return errors.Wrap(err, "problem saving private key error")
+		}
+	}
```
tychoish: rephrase these with and?

certdepot/cert_request.go
```diff
@@ -0,0 +1,119 @@
...
+	// DNS entries to add as subject alt name.
+	Domain []string
+	// URI values to add as subject alt name.
+	URI []string
+	// Path to private key PEM file (if blank, will generate new keypair).
+	Key string
+}
+
+// NewCertRequest creates a new certificate (CSR).
+func NewCertRequest(opts CertRequestOptions) error {
```
tychoish: split this function up to make it more testable? 

also does it make sense to take the Depot out of the Options struct and then have a method that's like Generate(Deopot) error to do this function?

certdepot/init.go
```diff
@@ -0,0 +1,96 @@
...
+
+	"github.com/mongodb/grip"
+	"github.com/pkg/errors"
+	"github.com/square/certstrap/depot"
+	"github.com/square/certstrap/pkix"
+)
+
+type InitOptions struct {
+	// The depot which stores the id/cert.
+	Depot depot.Depot
```
tychoish: take depot out? 

does it makes sense to collapse these option structures to a "certificate options," and then implement each of the operations as a *function that takes a deopot and returns an error"

certdepot/mongo_depot.go
```diff
@@ -0,0 +1,234 @@
...
+func NewMgoCertDepot(opts MgoCertDepotOptions) (depot.Depot, error) {
+	return newMgoCertDeposit(nil, opts)
+}
+
+// Create a new cert depot in the specified MongoDB, using an existing session.
+func NewMgoCertDepotWithSession(s *mgo.Session, opts MgoCertDepotOptions) (depot.Depot, error) {
+	return newMgoCertDeposit(s, opts)
+}
+
+func validate(opts MgoCertDepotOptions) MgoCertDepotOptions {
```
tychoish: make this a method on opts.

certdepot/mongo_depot.go
```diff
@@ -0,0 +1,234 @@
...
+	opts = validate(opts)
+
+	if s == nil {
+		var err error
+		s, err = mgo.DialWithTimeout(opts.MongoDBURI, opts.MongoDBDialTimeout)
+		if err != nil {
+			return nil, errors.Wrapf(err, "could not connect to db %s", opts.MongoDBURI)
+		}
+		s.SetSocketTimeout(opts.MongoDBSocketTimeout)
+	}
```
tychoish: I'd move the s == nil block to line 51...

certdepot/mongo_depot_test.go
```diff
@@ -0,0 +1,382 @@
...
+import (
+	"testing"
+	"time"
+
+	"github.com/square/certstrap/depot"
+	"github.com/stretchr/testify/suite"
+	mgo "gopkg.in/mgo.v2"
+)
+
+type MongoDepotTestSuite struct {
```
tychoish: I just thought it might be cool to add to your test running against the file depot as a way of validating the behavior of your depot implementation.

certdepot/cert_request.go
julianedwards: but what if it is a database error? I separated them because the error will not always necessarily occur because `opts.Passphrase` is emtpy.

certdepot/mongo_depot.go
julianedwards: it fails with lint if I do that... I get an error that the use of ops is ineffectual:
certdepot/mongo_depot.go:61:8:warning: ineffective assignment to field MongoDBURI (SA4005) (megacheck)
certdepot/mongo_depot.go:64:8:warning: ineffective assignment to field MongoDBDialTimeout (SA4005) (megacheck)
certdepot/mongo_depot.go:67:8:warning: ineffective assignment to field MongoDBSocketTimeout (SA4005) (megacheck)
certdepot/mongo_depot.go:70:8:warning: ineffective assignment to field DatabaseName (SA4005) (megacheck)
certdepot/mongo_depot.go:73:8:warning: ineffective assignment to field CollectionName (SA4005) (megacheck)
certdepot/mongo_depot.go:76:8:warning: ineffective assignment to field ExpireAfter (SA4005) (megacheck)
--- FAIL: lint-certdepot (26.153491873s)

certdepot/mongo_depot.go
julianedwards: Why? It goes nicely here, it makes this callable from both `NewMgoCertDepot` and `NewMgoCertDepotWithSession` and in the event a nil session is passed in to `NewMgoCertDepotWithSession` it will connect to the mongo uri in the options. Unless we shouldn't check for a nil session?

certdepot/mongo_depot_test.go
julianedwards: I agree, I will create a subsequent ticket for that.

certdepot/mongo_depot.go
julianedwards: oh right, duh

certdepot/mongo_depot.go
```diff
@@ -0,0 +1,233 @@
...
+
+	return &mongoCertDepot{
+		session:        s,
+		databaseName:   opts.DatabaseName,
+		collectionName: opts.CollectionName,
+		expireAfter:    opts.ExpireAfter,
+	}, nil
+}
+
+func (opts *MgoCertDepotOptions) validate() {
```
tychoish: error?

julianedwards: it is always nil since we set defaults for everything, should we just return nil?

PR #64, opened 2019-02-11
MAKE-627: move certificates into the database (cert creation)

julianedwards:


evergreen.yaml
```diff
@@ -82,6 +82,7 @@ functions:
     command: shell.exec
     type: test
     params:
+       silent: true
```
tychoish: turn off verbose when you pass silent: true

operations/service.go
```diff
@@ -45,20 +41,9 @@ func Service() cli.Command {
...
-				cli.StringFlag{
-					Name:   rpcCAPathFlag,
-					Usage:  "path to the rpc service ca cert",
-					EnvVar: envVarRPCCAPath,
-				},
-				cli.StringFlag{
-					Name:   rpcCertKeyPathFlag,
-					Usage:  "path to the prc service key",
-					EnvVar: envVarRPCCertKeyPath,
+				cli.BoolFlag{
```
tychoish: should this be boolT? (rpcDisableTLS)

operations/service.go
```diff
@@ -176,3 +176,86 @@ func signalListener(ctx context.Context, trigger context.CancelFunc) {
 
 	trigger()
 }
+
+func setupDepot(conf model.CertDepotConfig) (depot.Depot, error) {
```
tychoish: feels like we should move from here down into certDepot? (and maybe the config type?) and also test it? 

Also I think that model.CertDepotConfig should hold the CA cert/key so that you can bootstrap from a known ca?

rest/service.go
```diff
@@ -71,12 +73,21 @@ func (s *Service) Validate() error {
...
 	}
 
-	s.depot, err = depot.NewFileDepot(s.CertPath)
-	grip.Warning(errors.Wrap(err, "no certificate depot constructed"))
+	if s.Depot == nil {
+		grip.Warning(errors.Wrap(err, "no certificate depot provided"))
+		if err != nil {
+			return errors.Wrap(err, "problem setting up naive user manager")
+		}
+	}
```
tychoish: where does the error on line 86 come from? is this actually the naive user manager? 

it seems like we should also continue to allow this setup to run if there isn't a depot?

julianedwards: sorry that was a copy and paste error, removing it

operations/service.go
```diff
@@ -92,9 +77,7 @@ func Service() cli.Command {
 			dbName := c.String(dbNameFlag)
 			port := c.Int(servicePortFlag)
 
-			rpcCertPath := c.String(rpcCertPathFlag)
-			rpcCertKeyPath := c.String(rpcCertKeyPathFlag)
-			rpcCAPath := c.String(rpcCAPathFlag)
+			rpcTLS := c.Bool(rpcTLSFlag)
```
tychoish: I think this needs to become c.BoolT()

PR #70, opened 2019-02-28
MAKE-465: write job to backfill for independently uploaded ftdc files (on attach artifacts)

julianedwards:


rtimmons:
@bsamek can you step in while @tychoish is out? This is potentially blocking our work to integrate cedar and genny/dsi.

tychoish:
I'll look tonight or tomorrow.

tychoish:
@rtimmons just to be clear, the infrastructure already exists to track and manage artifacts submitted by genny/dsi, and this pull request shouldn't be blocking at all. Let's talk more if you hit a snag.

units/ftdc_rollups.go
```diff
@@ -0,0 +1,129 @@
...
+
+	if err := rollupJob.Validate(); err != nil {
+		return nil, errors.Wrap(err, "failed to setup new ftdc rollups job")
+	}
+	return rollupJob, nil
+}
+
+func (fr *ftdcRollups) Run(ctx context.Context) {
+	defer fr.MarkComplete()
+	defer fr.reset()
```
tychoish: the fact that this might be needed is kind of awkward/worrying. When is the job going to execute more than once?

units/ftdc_rollups.go
```diff
@@ -0,0 +1,129 @@
...
+				Name:    ftdcRollupsJobName,
+				Version: 1,
+			},
+		}
+	}
+}
+
+func (fr *ftdcRollups) Validate() error {
+	fr.setup()
+	fr.SetDependency(dependency.NewAlways())
```
tychoish: I think you could drop this by just having a seperate constructor and follow the pattern that all of the evergreen jobs have... (the duplication and difference between the environment that you test in and would use in actual execution (because of the factory) case coul,d lead to some interesting bugs.

units/ftdc_rollups.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func init() {
+	registry.AddJobType(ftdcRollupsJobName, func() amboy.Job {
+		fr := &ftdcRollups{}
+		fr.setup()
+		fr.SetDependency(dependency.NewAlways())
+
+		return fr
+	})
```
tychoish: I'd have a seperate private and follow the pattern that the evergreen versions have/use.

units/ftdc_rollups.go
```diff
@@ -0,0 +1,129 @@
...
+	fr.ArtifactInfo = nil
+}
+
+func NewFTDCRollupsJob(perfId string, artifactInfo *model.ArtifactInfo) (amboy.Job, error) {
+	rollupJob := &ftdcRollups{
+		PerfID:       perfId,
+		ArtifactInfo: artifactInfo,
+	}
+
+	if err := rollupJob.Validate(); err != nil {
```
tychoish: the validate method is only used once. and need not be public.

units/ftdc_rollups.go
```diff
@@ -0,0 +1,129 @@
...
+		fr.AddError(err)
+		return
+	}
+
+	result := &model.PerformanceResult{ID: fr.PerfID}
+	result.Setup(env)
+	err = result.Find()
+	if err != nil {
+		err = errors.Wrap(err, "problem running query")
+		grip.Warning(err)
```
tychoish: I don't think we need to log here? In general the error is logged at the execution level, and I don't think the extra logging volume actually adds much

units/ftdc_rollups.go
```diff
@@ -0,0 +1,129 @@
...
+		fr.AddError(err)
+		return
+	}
+
+	rollups := result.Rollups
```
tychoish: I wouldn't bother with the extra assignment

units/ftdc_rollups.go
julianedwards: won't, sorry I was copying from the other code in units

units/ftdc_rollups.go
```diff
@@ -0,0 +1,116 @@
...
+	j := &ftdcRollupsJob{
+		Base: job.Base{
+			JobType: amboy.JobType{
+				Name:    ftdcRollupsJobName,
+				Version: 1,
+			},
+		},
+	}
+
+	// TODO: Do I need to set priority ?
```
julianedwards: ?

tychoish: nope.

units/ftdc_rollups.go
```diff
@@ -19,111 +19,97 @@ const (
...
+	return j, nil
 }
 
-func (fr *ftdcRollups) Run(ctx context.Context) {
-	defer fr.MarkComplete()
-	defer fr.reset()
+func (j *ftdcRollupsJob) Run(ctx context.Context) {
+	defer j.MarkComplete()
 
 	env := cedar.GetEnvironment()
```
tychoish: I sometimes stash the env in the job body and call GetEnvironment only when that's nil to theoretically make it mockable.

units/ftdc_rollups.go
```diff
@@ -19,111 +19,97 @@ const (
...
 
-	result := &model.PerformanceResult{ID: fr.PerfID}
+	result := &model.PerformanceResult{ID: j.PerfID}
 	result.Setup(env)
 	err = result.Find()
 	if err != nil {
 		err = errors.Wrap(err, "problem running query")
-		grip.Warning(err)
-		fr.AddError(err)
+		j.AddError(err)
```
tychoish: I'd combine the wrapping and the adding onto one line.

units/ftdc_rollups.go
```diff
@@ -19,111 +19,97 @@ const (
...
-	rollups.Setup(env)
+	result.Rollups.Setup(env)
 	for _, stat := range stats {
-		err = rollups.Add(stat.Name, stat.Version, stat.UserSubmitted, stat.MetricType, stat.Value)
+		err = result.Rollups.Add(stat.Name, stat.Version, stat.UserSubmitted, stat.MetricType, stat.Value)
 		if err != nil {
 			err = errors.Wrapf(err, "problem adding rollup %s", stat.Name)
-			grip.Warning(err)
-			fr.AddError(err)
+			j.AddError(err)
```
tychoish: combine.

rpc/internal/service.go
```diff
@@ -241,17 +241,17 @@ func (srv *perfService) addArtifacts(record *model.PerformanceResult, artifacts
...
 		}
 
-		if ftdc {
+		if hasEventData {
 			return errors.New("cannot have more than one raw events artifact")
 		}
-		ftdc = true
+		hasEventData = true
 
 		job, err := units.NewFTDCRollupsJob(id, &artifact)
```
tychoish: can we not call getRemoteQueue on every loop and put that outside?

julianedwards: we can, but either way it will only get called once since we only ever create one job (it errors if it finds more than one ftdc artifact)

julianedwards: moved it

PR #71, opened 2019-03-05
MAKE-552: Create a merge rollups function on model.PerformanceResult

julianedwards:


model/perf_data_test.go
```diff
@@ -278,3 +274,38 @@ func (s *perfRollupSuite) TestUpdateDefaultRollups() {
...
+			Name:       "latency",
+			Value:      5000,
+			MetricType: MetricTypeLatency,
+		},
+	}
+
+	result := &PerformanceResult{}
+	s.Require().NoError(session.DB(conf.DatabaseName).C(perfResultCollection).FindId(s.r.id).One(result))
+	result.Setup(s.r.env)
+	s.NoError(result.MergeRollups(rollups))
```
tychoish: shouldn't we call this more than once to make sure that it's functionally idempotent?

rpc/internal/service.go
```diff
@@ -231,15 +231,11 @@ func (srv *perfService) CloseMetrics(ctx context.Context, end *MetricsSeriesEnd)
 }
 
 func addRollups(record *model.PerformanceResult, rollups []*RollupValue) error {
```
tychoish: I think we just need to make a function in the .ext.go file that converts a slice of []*RollupValue to []*model.PerfRollupValue and then we can remove this function entirely.

PR #73, opened 2019-03-06
MAKE-642: Write tests for Depot interface

julianedwards:


certdepot/depot_test.go
```diff
@@ -0,0 +1,473 @@
...
+						data, err = d.Get(depot.CrlTag(name))
+						assert.Error(t, err)
+						assert.Nil(t, data)
+					},
+				},
+
+				{
+					name: "DeleteWhenDNE",
+					test: func(t *testing.T, d depot.Depot) {
+						name := "bob"
```
tychoish: super nit: use constants here rather than variables, even at the same scope. It 100% doesn't matter, but seems nice.

certdepot/depot_test.go
```diff
@@ -0,0 +1,473 @@
...
+		t.Run(impl.name, func(t *testing.T) {
+			for _, test := range impl.tests {
+				t.Run(test.name, func(t *testing.T) {
+					d := impl.setup()
+					defer impl.cleanup()
+
+					test.test(t, d)
+				})
+			}
+			t.Run("Put", func(t *testing.T) {
```
tychoish: The main thing I'm interested in changing here is that it might make sense to put more sub-tests in these generic cases. Just to break things up and make it more clear what the expected aspect is that you're trying to test. When things break this makes it easier to figure out what happened.

certdepot/depot_test.go
```diff
@@ -0,0 +1,473 @@
...
+				d := impl.setup()
+				defer impl.cleanup()
+				name := "bob"
+
+				assert.Error(t, d.Put(depot.CrtTag(name), nil))
```
tychoish: should this be a require rather than an assert? 

require means that execution would stop here, and assert means that execution will continue even if the assertion fails. Often we overuse asserts and end up with way too much error output when a test fails. The risk of require is that one failure will mask another orthogonal failure... So it's fine to use asserts (and we certianly do a lot in different places, but...

julianedwards: Yeah I try to put requires only when I know the following code will either panic or result in a ton of failing asserts. Sometimes it is questionable, but I think here nothing will panic and it would be helpful to see if the following code works despite this code not returning an error when it should.

PR #76, opened 2019-03-19
MAKE-583: convert cedar to use new go driver

tychoish:


tychoish:
This is ready for review. I'm not particularly worried by the windows failure, and will continue to address it separately or during review.

PR #77, opened 2019-03-22
MAKE-644: handle certificate revocation and validation

julianedwards:


model/naive_users.go
```diff
@@ -72,6 +73,9 @@ func (*NaiveUserManager) IsRedirect() bool                          { return fal
 func (um *NaiveUserManager) GetUserByID(id string) (gimlet.User, error) {
 	for _, user := range um.users {
 		if user.User == id {
+			if user.Invalid {
```
tychoish: We've added a native (basic users and basic user manager) to the gimlet package. It might be the case that we can get rid of this implementation and just use that.

operations/service.go
```diff
@@ -106,8 +107,33 @@ func Service() cli.Command {
 				return errors.Wrap(err, "problem getting application configuration")
 			}
 
-			var d depot.Depot
+			var um gimlet.UserManager
```
tychoish: I kind of liked keeping more stuff out of operations/service (and I made a pass a while ago to factor it down.) It feels tempting to put a lot of the setup code here, but the more code we push into the packages the better. Because we end up making the user manager field in the service public, as a result of this change, we can just create it in the rest service and then reference it later. 

Another option would be create a function that generates the user manager in the service package (which would make more sense to test.) We could also just have two of them which would probably up our traffic to LDAP, but might be more logical seeming.

julianedwards: Sounds good. It is safe to assume that if gRPC is supporting TLS auth, there is either LDAP or naive auth configured, correct?

model/naive_users.go
julianedwards: Unfortunately, the basic user and user manager's GetUserById function do not support invalid users.

tychoish: can we add that functionality there.

operations/service.go
tychoish: yeah.

model/config.go
```diff
@@ -84,15 +84,34 @@ var (
...
+	AppAuth bool             `bson:"app_auth" json:"app_auth" yaml:"app_auth"`
+	Users   []MockUserConfig `bson:"users" json:"users" yaml:"users"`
 }
 
 var (
 	cedarNaiveAuthConfigAppAuthKey = bsonutil.MustHaveTag(NaiveAuthConfig{}, "AppAuth")
 	cedarNaiveAuthConfigUsersKey   = bsonutil.MustHaveTag(NaiveAuthConfig{}, "Users")
 )
 
+type MockUserConfig struct {
```
tychoish: perhaps call this UserConfig no mock?

PR #79, opened 2019-03-27
MAKE-677: add support for certificate rotation/expiration 

julianedwards:


rest/routes.go
```diff
@@ -830,3 +825,44 @@ func (s *Service) createUserCert(usr string) error {
...
+	cert, err := depot.GetCertificate(s.Depot, usr)
+	if err != nil {
+		return nil, errors.Wrap(err, "problem getting certificate")
+	}
+
+	rawCert, err := cert.GetRawCertificate()
+	if err != nil {
+		return nil, errors.Wrap(err, "problem getting raw certificate")
+	}
+	if rawCert.NotAfter.Before(time.Now().Add(time.Hour)) {
```
tychoish: this feels like it should be a constant, and after thinking let's renew if <4h and make certs last for 48 hours? We might want to push this up to 2weeks too? I'm worried about not having this configurable.

model/config.go
```diff
@@ -18,15 +18,16 @@ import (
...
-	Flags          OperationalFlags               `bson:"flags" json:"flags" yaml:"flags"`
-	Service        ServiceConfig                  `bson:"service" json:"service" yaml:"service"`
+	ID               string                         `bson:"_id" json:"id" yaml:"id"`
+	Splunk           send.SplunkConnectionInfo      `bson:"splunk" json:"splunk" yaml:"splunk"`
+	Slack            SlackConfig                    `bson:"slack" json:"slack" yaml:"slack"`
+	LDAP             LDAPConfig                     `bson:"ldap" json:"ldap" yaml:"ldap"`
+	NaiveAuth        NaiveAuthConfig                `bson:"naive_auth" json:"naive_auth" yaml:"naive_auth"`
+	CertDepot        certdepot.BootstrapDepotConfig `bson:"certdepot" json:"certdepot" yaml:"certdepot"`
+	SSLExpireAfter   time.Duration                  `bson:"ssl_expire" json:"ssl_expire" yaml:"ssl_expire"`
+	SSLRenewalBefore time.Duration                  `bson:"ssl_renewal" json:"ssl_renewal" yaml:"ssl_renewal"`
```
tychoish: can we put these flags in the certdepot config or else in a CAConfig section?

operations/service.go
```diff
@@ -110,7 +110,10 @@ func Service() cli.Command {
 			var err error
 			if rpcTLS {
 				if conf.SSLExpireAfter == 0 {
-					conf.SSLExpireAfter = 365 * 24 * time.Hour
+					conf.SSLExpireAfter = 48 * time.Hour
+				}
+				if conf.SSLRenewalBefore == 0 {
+					conf.SSLRenewalBefore = 4 * time.Hour
```
tychoish: if ssl config ends up in it's own struct then this should go into a validate section.

julianedwards:
Will do

PR #92, opened 2019-04-17
MAKE-729: Write job for backfilling outdated rollups

julianedwards:


model/perf.go
```diff
@@ -436,3 +436,42 @@ func (r *PerformanceResults) findAllChildrenGraphLookup(parent string, maxDepth
...
+		},
+	}
+	err = session.DB(conf.DatabaseName).C(perfResultCollection).Find(search).All(&r.Results)
+	if err != nil {
+		return errors.Wrapf(
+			err,
+			"problem finding perf results with outdated rollup %s (version %d)",
+			name,
+			version,
+		)
```
tychoish: I'd combine this on one line. To be honest you can really combine 466-476 on one or two lines, (but one statement)

units/find_outdated_rollups.go
```diff
@@ -0,0 +1,168 @@
...
+	if j.seenIDs == nil {
+		j.seenIDs = map[string]bool{}
+	}
+
+	factories := []perf.RollupFactory{}
+	for _, t := range j.RollupTypes {
+		factory := perf.RollupFactoryFromType(t)
+		if factory == nil {
+			err := errors.Errorf("problem resolving rollup factory type %s", t)
+			grip.Warning(err)
```
tychoish: I'd drop the warning, the execution of the job will log appropriately.

units/find_outdated_rollups.go
```diff
@@ -0,0 +1,168 @@
...
+	for _, artifact := range artifacts {
+		if artifact.Format == model.FileFTDC {
+			return &artifact
+		}
+	}
+
+	return nil
+}
+
+func findOutdatedFromResult(factories []perf.RollupFactory, result model.PerformanceResult) []perf.RollupFactory {
```
tychoish: does it make sense to move this into the perf package as a method somewhere?

units/ftdc_rollups.go
```diff
@@ -81,40 +97,58 @@ func (j *ftdcRollupsJob) Run(ctx context.Context) {
...
 		j.AddError(err)
 		return
 	}
 
+	rollups := []model.PerfRollupValue{}
+	for _, t := range j.RollupTypes {
+		factory := perf.RollupFactoryFromType(t)
+		if factory == nil {
+			err = errors.Errorf("problem resolving rollup factory type %s", t)
+			grip.Warning(err)
```
tychoish: avoid over logging.

units/ftdc_rollups.go
```diff
@@ -55,19 +58,32 @@ func (j *ftdcRollupsJob) validate() error {
...
 	j := makeFTDCRollupsJob()
 	j.PerfID = perfId
 	j.ArtifactInfo = artifactInfo
+	j.UserSubmitted = user
+
+	j.RollupTypes = []string{}
+	for _, factory := range factories {
+		// TODO: Should we add an assertion here, making sure that the
+		// rollup types are in the "registry"? Can also do this in
+		// validate.
```
tychoish: probably!

units/find_outdated_rollups.go
```diff
@@ -0,0 +1,168 @@
...
+
+	job, err := NewFTDCRollupsJob(result.Info.ID(), getFTDCArtifact(result.Artifacts), outdated, false)
+	if err != nil {
+		err = errors.Wrapf(err, "problem creating FTDC rollups job for %s", result.Info.ID())
+		grip.Warning(err)
+		j.AddError(err)
+		return
+	}
+
+	if err = j.env.GetRemoteQueue().Put(job); err != nil {
```
tychoish: can we pass the queue into this method, or set it as a private field on the job (GetRemoteQueue is RWMutex'd) and setting/passing the queue makes it the job a touch eaiser to test

units/find_outdated_rollups.go
```diff
@@ -0,0 +1,168 @@
...
+			if err := results.FindOutdatedRollups(name, factory.Version(), after); err != nil {
+				err = errors.Wrapf(err, "problem checking for outdated rollups for %s", name)
+				grip.Warning(err)
+				j.AddError(err)
+				continue
+			}
+
+			for _, result := range results.Results {
+				if _, ok := j.seenIDs[result.Info.ID()]; !ok {
+					j.createFTDCRollupsJobs(factories[i:], result)
```
tychoish: why is this factories[i:] and not factories[i]

units/find_outdated_rollups.go
```diff
@@ -0,0 +1,168 @@
...
+	job, err := NewFTDCRollupsJob(result.Info.ID(), getFTDCArtifact(result.Artifacts), outdated, false)
+	if err != nil {
+		err = errors.Wrapf(err, "problem creating FTDC rollups job for %s", result.Info.ID())
+		grip.Warning(err)
+		j.AddError(err)
+		return
+	}
+
+	if err = j.env.GetRemoteQueue().Put(job); err != nil {
+		err = errors.Wrapf(err, "problem putting FTDC rollups job %s on remote queue", j.ID())
```
tychoish: don't we want to put job.ID() in this message?

units/find_outdated_rollups.go
julianedwards: It needs to check the rest of the rollups, this way there is only ever one job created per result (and thus one s3 download). So the first line in `createFTDCRollupsJobs` calls `findOutdatedFromResult` to do the check.

units/find_outdated_rollups.go
julianedwards: It's there.

units/find_outdated_rollups.go
julianedwards: I thought about that, this function is pretty specific to this job. It can always be easily moved in the future if needed, but as of now it seems awkward/overkill to put it in the perf package.

units/find_outdated_rollups.go
julianedwards: should I get rid of all the warnings here?

units/ftdc_rollups.go
julianedwards: So in the Run command I continue on error if I cannot resolve a specific factory. Do you think it is a better idea to fail early and not run anything at all?

model/perf.go
```diff
@@ -436,3 +436,34 @@ func (r *PerformanceResults) findAllChildrenGraphLookup(parent string, maxDepth
...
+
+	search := bson.M{
+		perfCreatedAtKey: bson.M{"$gt": after},
+		bsonutil.GetDottedKeyName(perfArtifactsKey, artifactInfoFormatKey): FileFTDC,
+		"$or": []bson.M{
+			{
+				bsonutil.GetDottedKeyName(perfRollupsKey, perfRollupsStatsKey): bson.M{
+					"$elemMatch": bson.M{
+						perfRollupValueNameKey:    name,
+						perfRollupValueVersionKey: bson.M{"$lt": version},
```
tychoish: nit: lte?

units/ftdc_rollups.go
tychoish: at runtime is probably better, I'd just delete this comment.

units/ftdc_rollups.go
```diff
@@ -55,19 +57,32 @@ func (j *ftdcRollupsJob) validate() error {
...
+		j.RollupTypes = append(j.RollupTypes, factory.Type())
+	}
 
 	if err := j.validate(); err != nil {
 		return nil, errors.Wrap(err, "failed to create new ftdc rollups job")
 	}
 
-	j.SetID(fmt.Sprintf("perf-rollup.%s.%s", perfId, artifactInfo.Path))
+	j.SetID(fmt.Sprintf("perf-rollup.%s.%s.%s", perfId, artifactInfo.Path, time.Now().Truncate(2*time.Hour)))
```
tychoish: should we import the round-part-of hour functions from util in evergreen?

model/perf_data.go
```diff
@@ -103,7 +101,7 @@ func (r *PerfRollups) Setup(env cedar.Environment) {
 	r.env = env
 }
 
-func (r *PerfRollups) Add(ctx context.Context, name string, version int, userSubmitted bool, t MetricType, value interface{}) error {
+func (r *PerfRollups) Add(ctx context.Context, name string, version int, userSubmitted, valid bool, t MetricType, value interface{}) error {
```
tychoish: I feel like two bools in a row is hard to read, the arguments to this function feel like they're getting a little out of control. Should we try and use an arg-struct? or pass a PerfRollupValue directly?

model/perf.go
julianedwards: We don't want to update a rollup if it has the current version, it should be strictly less than.

tychoish:
this looks good let's make the tests to pass

PR #94, opened 2019-04-24
EVG-6087 add limit to perf result by name route

john-m-liu:


model/perf.go
```diff
@@ -272,7 +273,11 @@ func (r *PerformanceResults) Find(options PerfFindOptions) error {
 	}
 
 	r.populated = false
-	err = session.DB(conf.DatabaseName).C(perfResultCollection).Find(search).All(&r.Results)
+	query := session.DB(conf.DatabaseName).C(perfResultCollection).Find(search).Sort("-" + perfCreatedAtKey)
```
tychoish: is there an index for this?

rest/perf_routes.go
```diff
@@ -8,6 +8,8 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/mongodb/grip"
+
```
tychoish: blank line

john-m-liu: fixed

model/perf.go
john-m-liu: There's a created_at index but since this query can be many shapes it may or may not be used

PR #95, opened 2019-04-26
MAKE-711: make grpc client more user friendly

julianedwards:


tychoish:
let's make the tests pass though

PR #98, opened 2019-05-09
MAKE-762: Get last N perf results by task name sorted by version order

julianedwards:


model/perf.go
```diff
@@ -272,10 +275,14 @@ func (r *PerformanceResults) Find(options PerfFindOptions) error {
...
 	r.populated = false
-	query := session.DB(conf.DatabaseName).C(perfResultCollection).Find(search).Sort("-" + perfCreatedAtKey)
+	query := session.DB(conf.DatabaseName).C(perfResultCollection).Find(search)
 	if options.Limit > 0 {
 		query = query.Limit(options.Limit)
 	}
+	if options.Sort != nil {
+		query = query.Sort(options.Sort...)
+	}
+	query = query.Sort("-" + perfCreatedAtKey)
```
julianedwards: maybe make this an else ?

model/perf.go
```diff
@@ -170,6 +170,7 @@ func (result *PerformanceResult) Remove() (int, error) {
 type PerformanceResultInfo struct {
 	Project   string           `bson:"project,omitempty"`
 	Version   string           `bson:"version,omitempty"`
+	Order     int              `bson:"order,omitempty"`
```
tychoish: Does this need to change the hash?

rest/data/perf.go
```diff
@@ -96,7 +97,11 @@ func (dbc *DBConnector) FindPerformanceResultsByTaskId(taskId string, interval u
 
 // FindPerformanceResultsByTaskName queries the database to find all performance
 // results with the given taskName, time inteval, and optional tags.
-func (dbc *DBConnector) FindPerformanceResultsByTaskName(taskName string, interval util.TimeRange, limit int, tags ...string) ([]dataModel.APIPerformanceResult, error) {
+// If limit is greater than 0, the number of results returned will be no
+// greater than limit.
+// If sorted is true, the results will be sorted (descending) by the order
+// field.
+func (dbc *DBConnector) FindPerformanceResultsByTaskName(taskName string, interval util.TimeRange, limit int, sorted bool, tags ...string) ([]dataModel.APIPerformanceResult, error) {
```
tychoish: Let's just always sort this and not have the bool

model/perf.go
julianedwards: added it to the hash, though it probably doesn't need to be there because it isn't really adding to the uniqueness of a result

julianedwards: On second thought, since it is not needed for uniqueness I am not going to add it to the hash. It is not worth the effort of changing the schema if it is not going to make a difference with respect to distinguishing results.

model/perf.go
tychoish: yes

PR #99, opened 2019-05-09
MAKE-762: Update protobuf

julianedwards:


perf.proto
```diff
@@ -10,17 +10,18 @@ import "google/protobuf/timestamp.proto";
...
-  int32 execution = 5;
-  string task_id = 6;
-  string test_name = 7;
-  string parent = 8;
-  int32 trial = 9;
-  repeated string tags = 10;
-  map<string, int32> arguments = 11;
-  bool mainline = 12;
-  google.protobuf.Timestamp created_at = 13;
+  int32 order = 3;
```
tychoish: actually, in order to get backwards compatibility, this has to be field 14 at the end.

PR #103, opened 2019-05-16
MAKE-792 filter perf results by variant

john-m-liu:


rest/data/interface.go
```diff
@@ -10,7 +10,7 @@ type Connector interface {
 	FindPerformanceResultById(string) (*model.APIPerformanceResult, error)
 	RemovePerformanceResultById(string) (int, error)
 	FindPerformanceResultsByTaskId(string, util.TimeRange, ...string) ([]model.APIPerformanceResult, error)
-	FindPerformanceResultsByTaskName(string, util.TimeRange, int, ...string) ([]model.APIPerformanceResult, error)
+	FindPerformanceResultsByTaskName(string, string, util.TimeRange, int, ...string) ([]model.APIPerformanceResult, error)
```
tychoish: Should we change the method name to match?

john-m-liu: I think the method name is still correct since the only required info is the task name

tychoish:
I think the tests should pass?

john-m-liu:
It looks flaky to me. I didn't change anything related to that though it passes on the base commit. I restarted it

PR #104, opened 2019-05-29
MAKE-754: Server cert rotation (and check for expiration)

julianedwards:


units/crons.go
```diff
@@ -78,5 +79,32 @@ func StartCrons(ctx context.Context, env cedar.Environment) error {
 		return queue.Put(job)
 	})
 
+	if rpcTLS {
+		amboy.IntervalQueueOperation(ctx, remote, 24*time.Hour, time.Now(), opts, func(queue amboy.Queue) error {
+			return queue.Put(NewServerCertRotationJob())
+		})
+		amboy.IntervalQueueOperation(ctx, local, time.Hour, time.Now(), opts, func(queue amboy.Queue) error {
```
tychoish: I'd write a job to do this, and call Close on the environment which you could rig to cancel to starting context in a way...

tychoish: also round now to the top of the hour or something... because we want it fuzzed to a minute and not have it happen too often when we're deploying or the app restarts/crashes.)

units/crons.go
```diff
@@ -78,5 +79,32 @@ func StartCrons(ctx context.Context, env cedar.Environment) error {
 		return queue.Put(job)
 	})
 
+	if rpcTLS {
+		amboy.IntervalQueueOperation(ctx, remote, 24*time.Hour, time.Now(), opts, func(queue amboy.Queue) error {
+			return queue.Put(NewServerCertRotationJob())
```
tychoish: we should have the starting time (the thing that's now in the args) be a fixed time (e.g. midnight) to avoid having this happen every time we restart the server(? right?) 

also I think we can use the close method on the environment to cancel the master context here too.

julianedwards: How about doing time.Now() + 12 hours?

units/server_cert_restart.go
```diff
@@ -0,0 +1,86 @@
...
+		j.AddError(err)
+		return
+	}
+
+	localServerCertVersion := j.env.GetServerCertVersion()
+	if localServerCertVersion < 0 {
+		j.env.SetServerCertVersion(conf.CA.ServerCertVersion)
+	} else if localServerCertVersion < conf.CA.ServerCertVersion {
+		grip.Info("restarting application to update server certificate")
+		closeCtx, cancel := context.WithTimeout(context.Background(), time.Minute)
```
tychoish: I'd kill it after 15s

PR #105, opened 2019-05-29
MAKE-774: Create a permissions type for AWS permissions

julianedwards:


model/buildlogger_storage.go
```diff
@@ -0,0 +1,67 @@
...
+type S3Permissions string
+
+const (
+	S3PermissionsPrivate                S3Permissions = s3.ObjectCannedACLPrivate
+	S3PermissionsPublicRead             S3Permissions = s3.ObjectCannedACLPublicRead
+	S3PermissionsPublicReadWrite        S3Permissions = s3.ObjectCannedACLPublicReadWrite
+	S3PermissionsAuthenticatedRead      S3Permissions = s3.ObjectCannedACLAuthenticatedRead
+	S3PermissionsAWSExecRead            S3Permissions = s3.ObjectCannedACLAwsExecRead
+	S3PermissionsBucketOwnerRead        S3Permissions = s3.ObjectCannedACLBucketOwnerRead
+	S3PermissionsBucketOwnerFullControl S3Permissions = s3.ObjectCannedACLBucketOwnerFullControl
```
tychoish: should these be bucket permissions rather than make it s3 specific? and avoid the dependency on the s3 pacakge?

julianedwards: so move these to pail? I was thinking that but we already depend on the aws-sdk-go package in cost package.

tychoish: I think this should be in pail, and refer to a concept/abstraction/etc that isn't s3 specific.

model/buildlogger_storage.go
```diff
@@ -0,0 +1,41 @@
...
+const (
+	LogFormatUnknown LogFormat = "unknown"
+	LogFormatText    LogFormat = "text"
+	LogFormatJSON    LogFormat = "json"
+	LogFormatBSON    LogFormat = "bson"
+)
+
+func (lf LogFormat) Validate() error {
+	switch lf {
+	case LogFormatUnknown, LogFormatText, LogFormatJSON, LogFormatBSON:
```
tychoish: what does it mean to have an unknown log format? 

are there schemas associated with json/bson that we'd want to track which would mean that we'd need more information?

model/buildlogger_storage.go
```diff
@@ -0,0 +1,41 @@
+package model
+
+import (
+	"github.com/evergreen-ci/pail"
+	"github.com/mongodb/anser/bsonutil"
+	"github.com/pkg/errors"
+)
+
+type LogFormat string
```
tychoish: add docs string

model/buildlogger_storage.go
julianedwards: I guess unknown would be the default if a log format is not specified, similar to how we have an unknown format for perf artifacts. Not sure what you mean about the schemas associated with json/bson, could you give an example?

PR #106, opened 2019-06-10
MAKE-772: Create Log and LogInfo struct for database model

julianedwards:


model/buildlogger_test.go
```diff
@@ -0,0 +1,200 @@
...
+	"testing"
+	"time"
+
+	"github.com/evergreen-ci/cedar"
+	"github.com/evergreen-ci/pail"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var (
```
tychoish: can these data be in scope in the test or produced by some kind of constructing function?

model/buildlogger.go
```diff
@@ -0,0 +1,187 @@
...
+	"github.com/mongodb/anser/db"
+	"github.com/mongodb/anser/model"
+	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/message"
+	"github.com/pkg/errors"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+const buildloggerCollection = "buildlogger"
```
tychoish: can we call this "buildlogs"

PR #107, opened 2019-06-10
MAKE-773: Implement CreateLog function

julianedwards:


model/buildlogger.go
```diff
@@ -21,6 +21,16 @@ import (
 
 const buildloggerCollection = "buildlogs"
 
+func CreateLog(info LogInfo, artifact LogArtifactInfo) *Log {
+	return &Log{
+		ID:       info.ID(),
+		Info:     info,
+		Artifact: artifact,
```
tychoish: aren't we going to want to support having more than one artifact?

julianedwards: Not in my design, a log in the database maps to one sub-bucket in aws (where the log is broken up into chunks)

tychoish: how will we know how many subchunks there are supposed to be, and how many lines are stored in each chunk? seems like we'd want to be able to have that info per artifact.

julianedwards: We can add something to store the number of chunks (and maybe the number of lines in each chunk) but that would still be one artifact. Storing the number of chunks would just be for sanity/error checking, i.e. making sure when we download the log file that we have the right number of chunks and each chunk has the right number of lines. Since the timestamp is the key of each chunk file, the entire sub-bucket is downloaded and the log is stitched back together by sorting the filenames.

model/buildlogger_storage.go
```diff
@@ -27,18 +27,35 @@ func (lf LogFormat) Validate() error {
...
 	logArtifactInfoVersionKey     = bsonutil.MustHaveTag(LogArtifactInfo{}, "Version")
+	logArtifactInfoChunksKey      = bsonutil.MustHaveTag(LogArtifactInfo{}, "Chunks")
+)
+
+// LogChunkInfo describes a chunk of log lines stored in s3.
+type LogChunkInfo struct {
+	Key      string `bson:"key"`
+	NumLines int    `bson:"num_lines"`
+	Start    int64  `bson:"start"`
+	End      int64  `bson:"end"`
```
tychoish: shouldn't these be either (preferably) time.Time instances?

model/buildlogger_storage.go
```diff
@@ -27,18 +27,35 @@ func (lf LogFormat) Validate() error {
...
 }
 
 var (
 	logArtifactInfoPrefixKey      = bsonutil.MustHaveTag(LogArtifactInfo{}, "Prefix")
 	logArtifactInfoPermissionsKey = bsonutil.MustHaveTag(LogArtifactInfo{}, "Permissions")
 	logArtifactInfoVersionKey     = bsonutil.MustHaveTag(LogArtifactInfo{}, "Version")
+	logArtifactInfoChunksKey      = bsonutil.MustHaveTag(LogArtifactInfo{}, "Chunks")
+)
+
+// LogChunkInfo describes a chunk of log lines stored in s3.
```
tychoish: in a bucket? could use a different backend?

julianedwards: This artifact is coupled pretty tightly with S3

model/buildlogger_storage.go
julianedwards: No, the timestamps are stored as milliseconds since the unix epoch so these should be in64s.

model/buildlogger.go
```diff
@@ -88,21 +98,52 @@ func (l *Log) Save() error {
...
 		"ns":           model.Namespace{DB: l.env.GetConf().DatabaseName, Collection: buildloggerCollection},
 		"id":           l.ID,
 		"updateResult": updateResult,
-		"artifact":     l.Artifact,
-		"op":           "save buildlogger log",
+		"logChunkInfo": logChunk,
+		"op":           "append log chunk info to buildlogger log",
 	})
-	return errors.Wrap(err, "problem saving log to collection")
+	if err == nil && updateResult.MatchedCount == 0 {
```
julianedwards: I decided to return an error if matched count is 0 (meaning the id DNE in the DB), the alternative is to return matched count and not error. Let me know what you think.

model/buildlogger_storage.go
tychoish: I think we should convert them in the application/database because anything we can do to make handling time math easier would be good.

model/buildlogger_storage.go
tychoish: can we relax it a bit?

model/buildlogger.go
tychoish: sounds good!

model/buildlogger_storage.go
tychoish: Theoretically, given how the application works, (i.e. the protections for content in s3 is all read by the application and then returned to the users from the application, so our security model is really external to the bucket, which I think would  resolve the concerns here. 

Put another way, access to the s3 bucket without the application is not likely to be particularly useful.

PR #109, opened 2019-06-13
MAKE-776: Implement uploader

julianedwards:


model/buildlogger.go
```diff
@@ -109,9 +115,71 @@ func (l *Log) SaveNew() error {
...
+		return errors.New("cannot remove a log with a nil environment")
+	}
+	ctx, cancel := l.env.Context()
+	defer cancel()
+
+	if l.ID == "" {
+		l.ID = l.Info.ID()
+	}
+
+	_, err := l.env.GetDB().Collection(buildloggerCollection).DeleteOne(ctx, bson.M{"_id": l.ID})
```
tychoish: why is the first return value ignorable? I think it can fail to delete in some cases without returning errors.

model/buildlogger.go
```diff
@@ -109,9 +115,71 @@ func (l *Log) SaveNew() error {
...
+	if l.env == nil {
+		return errors.New("cannot not upload log lines with a nil environment")
+	}
+	if len(lines) == 0 {
+		grip.Warning("Log.Upload called with empty LogLine slice")
+		return nil
+	}
+	ctx, cancel := l.env.Context()
+	defer cancel()
+	key := fmt.Sprint(time.Now())
```
tychoish: we might want to have a different key format.

model/buildlogger.go
```diff
@@ -227,3 +279,11 @@ func (id *LogInfo) ID() string {
 
 	return fmt.Sprintf("%x", hash.Sum(nil))
 }
+
+// LogLine describes a buildlogger log line. This is an intermediary type that
+// passes data from RPC calls to the upload phase.
+type LogLine struct {
+	Timestamp time.Time
+	// TODO: string or byte array?
```
tychoish: string, probably. 

does the format of this depend on the log line format?

model/buildlogger.go
```diff
@@ -109,9 +115,77 @@ func (l *Log) SaveNew() error {
...
+// the uploaded lines.
+func (l *Log) Upload(lines []LogLine) error {
+	if !l.populated {
+		return errors.New("cannot upload log lines when log unpopulated")
+	}
+	if l.env == nil {
+		return errors.New("cannot not upload log lines with a nil environment")
+	}
+	if len(lines) == 0 {
+		grip.Warning("Log.Upload called with empty LogLine slice")
```
tychoish: make this a structgured call and put some information about the log id in the message.Fields so we at least have a little bit of a record.

util/time.go
```diff
@@ -2,6 +2,16 @@ package util
...
 
+// UnixMilli returns t as a Unix time, the number of nanoseconds elapsed since
+// January 1, 1970 UTC. The result is undefined if the Unix time in nanoseconds
+// in cannot be represented by an int64 (a date before the year 1678 or after
+// 2262). Note that this means the result of calling UnixMilli on the zero Time
+// on the zero Time is undefined. The result does not depend on the location
+// associated with t.
+func UnixMilli(t time.Time) int64 {
+	return t.UnixNano() / int64(time.Millisecond)
+}
```
tychoish: should we add a converter in the other direction?

julianedwards: I will when I write the downloader

PR #111, opened 2019-06-17
MAKE-783: Create buildlogger proto file

julianedwards:


buildlogger.proto
```diff
@@ -0,0 +1,44 @@
...
+  string version = 2;
+  string variant = 3;
+  string task_name = 4;
+  string task_id = 5;
+  int32 execution = 6;
+  string test_name = 7;
+  int32 trial = 8;
+  string proc_name = 9;
+  Format format = 10;
+  map<string, string> arguments = 11;
```
tychoish: the struct in go is map[string]int32 (which I think is wrong, relative to this, so we should probably change it.

julianedwards: good catch!

julianedwards:
Will attach and implement service in the following tickets in my sprint

PR #114, opened 2019-06-19
MAKE-785: Create and attach gRPC service

julianedwards:


buildlogger.proto
```diff
@@ -42,3 +42,15 @@ message LogEndInfo {
   int32 exit_code = 1;
   google.protobuf.Timestamp completed_at = 2;
 }
+
+message BuildloggerResponse {
+  string id = 1;
```
tychoish: can this be log_id?

buildlogger.proto
```diff
@@ -42,3 +42,15 @@ message LogEndInfo {
...
+  bool success = 2;
+}
+
+
+service Buildlogger {
```
tychoish: should we also have a client-streaming interface?

julianedwards: added

buildlogger.proto
```diff
@@ -42,3 +42,16 @@ message LogEndInfo {
...
+
+message BuildloggerResponse {
+  string log_id = 1;
+  bool success = 2;
+}
+
+
+service Buildlogger {
+  rpc CreateLog(LogInfo) returns (BuildloggerResponse);
+  rpc AppendLogLines(LogLines) returns (BuildloggerResponse);
```
tychoish: this should probably be "repeated" for the log lines. because it's a single line. ( you could change the name to be signular, but I think we should encourage people to batch in some way.

buildlogger.proto
```diff
@@ -42,3 +42,16 @@ message LogEndInfo {
...
+  rpc StreamAppendLogLines(stream LogLines) returns (BuildloggerResponse);
```
tychoish: StreamLog

buildlogger.proto
julianedwards: No, LogLines is multiple lines (there is LogLine and LogLines)

tychoish:
sounds good. fix the ubuntu build before pushing.

PR #118, opened 2019-06-25
MAKE-788: Implement gRPC CloseLog

julianedwards:


model/buildlogger.go
```diff
@@ -217,9 +218,49 @@ func (l *Log) appendLogChunkInfo(logChunk LogChunkInfo) error {
...
+		l.ID = l.Info.ID()
+	}
+
+	update := bson.M{
+		"$set": bson.M{
+			logCompletedAtKey: completedAt,
+			bsonutil.GetDottedKeyName(logInfoKey, logInfoExitCodeKey): exitCode,
+		},
+	}
+	updateResult, err := l.env.GetDB().Collection(buildloggerCollection).UpdateOne(ctx, bson.M{"_id": l.ID}, update)
```
tychoish: nit, I'd combine the update statement into this statement?

model/buildlogger.go
```diff
@@ -217,9 +218,49 @@ func (l *Log) appendLogChunkInfo(logChunk LogChunkInfo) error {
...
+	grip.DebugWhen(err == nil, message.Fields{
+		"ns":           model.Namespace{DB: l.env.GetConf().DatabaseName, Collection: buildloggerCollection},
```
tychoish: I'd just put the collection. I think GetConf is (or could be in the future) more expensive than we'd like, and the db name isn't as confusing.

rpc/internal/buildlogger_service.go
```diff
@@ -66,6 +67,24 @@ func (s *buildloggerService) StreamLog(stream Buildlogger_StreamLogServer) error
...
+	if info.CompletedAt != nil {
+		var err error
+		completedAt, err = ptypes.Timestamp(info.CompletedAt)
+		if err != nil {
+			return nil, errors.Wrap(err, "problem converting completed_at timestamp")
+		}
+	}
+
+	log.Setup(s.env)
+	return &BuildloggerResponse{LogId: log.ID}, errors.Wrapf(log.CloseLog(completedAt, int(info.ExitCode)), "problem closing log with id %s", log.ID)
```
tychoish: It might be the case that you want to avoid returning errors from these functions in all cases (e.g. remember the stuff about error codes and what not? Return errors might not be big enough

rpc/internal/buildlogger_service.go
```diff
@@ -66,6 +67,24 @@ func (s *buildloggerService) StreamLog(stream Buildlogger_StreamLogServer) error
...
+	completedAt := time.Now()
+	if info.CompletedAt != nil {
+		var err error
+		completedAt, err = ptypes.Timestamp(info.CompletedAt)
+		if err != nil {
+			return nil, errors.Wrap(err, "problem converting completed_at timestamp")
+		}
+	}
+
+	log.Setup(s.env)
```
tychoish: This is pretty confusing, and I remember why we need to do it. Could we just make log.Find() or whatever set the env again?

rpc/internal/test_utils.go
```diff
@@ -0,0 +1,8 @@
+package internal
+
+var globalPortCounter = 3999
+
+func getPort() int {
+	globalPortCounter += 1
+	return globalPortCounter
```
tychoish: There's a pattern that we use in a few places where we do this with a channel and a thread that adds ints and returns ints (which avoids using a lock, which you should probably have regardless, I think. 

Also ++ rather than += 1 

also this file should be named utils_test.go so that it only builds in otherwise this ends up being in prod code.

model/buildlogger.go
julianedwards: Really? I feel like it gets too messy when they are combined. If you feel strongly about it I can combine them.

model/buildlogger.go
julianedwards: Set it to just the collection for all of these debug statements. Should I do the same in perf?

rpc/internal/buildlogger_service.go
julianedwards: gRPC will return these errors with an unknown error code: `rpc error: code = Unknown desc = problem finding log record for '3d3dca54db40c6b0ca6a770f3376da833dcdd797': cannot find with a nil environment`. I can use the gRPC status package to set the rpc code, but then I lose the ability to wrap error messages. What do you think?

rpc/internal/buildlogger_service.go
julianedwards: turns out we don't actually need to set the env twice, using the go driver's decode func doesn't wipe it out :)

model/buildlogger.go
tychoish: I feel moderately strongly.

model/buildlogger.go
tychoish: yeah

rpc/internal/buildlogger_service.go
tychoish: Either way is probably fine. I don't think we get a lot out of wrapping in this context.

Alternatively, (and for future work) we could see if there was a way to have the middleware take care of wrapping/unwrapping errors.

PR #121, opened 2019-07-12
MAKE-823: increase allowed msg size for rpc and fix rpc flag

julianedwards:


rpc/service.go
```diff
@@ -113,6 +113,7 @@ func GetServer(env cedar.Environment, conf CertConfig) (*grpc.Server, error) {
 		opts,
 		grpc.UnaryInterceptor(aviation.ChainUnaryServer(unaryInterceptors...)),
 		grpc.StreamInterceptor(aviation.ChainStreamServer(streamInterceptors...)),
+		grpc.MaxRecvMsgSize(5e9),
```
tychoish: should this be configurable? this is maxint right? shouldn't we set this at like some small number of gigs to avoid ooming?

operations/service.go
```diff
@@ -78,7 +78,8 @@ func Service() cli.Command {
 			dbName := c.String(dbNameFlag)
 			port := c.Int(servicePortFlag)
 
-			rpcTLS := c.BoolT(rpcTLSFlag)
+			rpcTLS := !c.Bool(rpcTLSFlag)
+			fmt.Println(rpcTLS)
```
tychoish: don't print?

PR #122, opened 2019-07-22
MAKE-777: Implement downloader

julianedwards:
This is a WIP because I have not written tests. Would like to first go over and get approval for design.

### DIFF
--- a/model/buildlogger.go
+++ b/model/buildlogger.go
@@ -314,17 +314,17 @@ func (l *Log) Download() (*LogIterator, error) {
 	for i := 0; i < runtime.NumCPU(); i++ {
 		wg.Add(1)
 		go func() {
-			defer wg.Done()
-			//defer recovery.LogAndContinue("downloader")
 			defer func() {
 				if r := recover(); r != nil {
 					// TODO: should this be critical?
 					grip.Criticalf("download go routine panicked: %s", r)
 				}
+				wg.Done()
 			}()
 
 			for chunk := range work {
 				if err := ctx.Done(); err != nil {
+					catcher.Add(ctx.Err())
 					return
 				} else {
 					r, err := bucket.Get(ctx, chunk.Key)
--- a/model/buildlogger.go
+++ b/model/buildlogger.go
@@ -1,12 +1,17 @@
 package model
 
 import (
+	"bufio"
+	"context"
 	"crypto/sha1"
 	"fmt"
 	"hash"
 	"io"
+	"runtime"
 	"sort"
+	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/evergreen-ci/cedar"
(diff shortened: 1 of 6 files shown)


model/buildlogger.go
```diff
@@ -266,6 +271,154 @@ func (l *Log) CloseLog(completedAt time.Time, exitCode int) error {
 
 }
 
+// Download returns a LogIterator which iterates lines of the given log. The
+// log should be populated and the environment should not be nil.
+func (l *Log) Download() (*LogIterator, error) {
```
tychoish: we should also benchmark this.

model/buildlogger.go
```diff
@@ -266,6 +271,154 @@ func (l *Log) CloseLog(completedAt time.Time, exitCode int) error {
...
+
+		i.currentReader = nil
+		i.lineCount = 0
+		i.keyIndex++
+
+		return i.Next(ctx)
+	} else if err != nil {
+		i.catcher.Add(errors.Wrap(err, "problem getting line"))
+		return false
+	}
```
tychoish: I don't think you need the else

model/buildlogger.go
tychoish: it might also be good to set up an API so that we can get ranges? and maybe (optionally force serialization? It might be the case that buffering in this way isn't always good, and it'd be nice to be able to select "serial" vs "buffer everything" 

if there end up being a lot of files, this method might also hit open-file descriptor limits.

model/buildlogger.go
```diff
@@ -266,6 +271,154 @@ func (l *Log) CloseLog(completedAt time.Time, exitCode int) error {
...
+	close(work)
+	var wg sync.WaitGroup
+	readers := map[string]io.ReadCloser{}
+	catcher := grip.NewBasicCatcher()
+	for i := 0; i < runtime.NumCPU(); i++ {
+		wg.Add(1)
+		go func() {
+			defer func() {
+				if r := recover(); r != nil {
```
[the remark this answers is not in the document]
julianedwards: Should I then log the error message?

`grip.Criticalf(recovery.HandlePanicWithError(recover(), errors.New("panic")))` ?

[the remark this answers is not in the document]
julianedwards: or declare a default error return value?

model/buildlogger.go
```diff
@@ -266,6 +271,154 @@ func (l *Log) CloseLog(completedAt time.Time, exitCode int) error {
...
+func (i *LogIterator) Err() error {
+	return i.catcher.Resolve()
+}
+
+// Item returns the current LogLine item.
+func (i *LogIterator) Item() LogLine { return i.currentItem }
+
+// Close closes the iterator. This function should always be called once the
+// iterator is done being used.
+func (i *LogIterator) Close() error {
```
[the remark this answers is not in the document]
julianedwards: I think they should be two separate things in order to avoid overwhelming error messages.

model/buildlogger.go
```diff
@@ -266,6 +271,154 @@ func (l *Log) CloseLog(completedAt time.Time, exitCode int) error {
...
+// Download returns a LogIterator which iterates lines of the given log. The
+// log should be populated and the environment should not be nil.
+func (l *Log) Download() (*LogIterator, error) {
+	if !l.populated {
+		return nil, errors.New("cannot downdload log when log unpopulated")
+	}
+	if l.env == nil {
+		return nil, errors.New("cannot download log with a nil environment")
+	}
+	ctx, cancel := l.env.Context()
```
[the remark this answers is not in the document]
julianedwards: isn't this the context we always use for methods on Log?

model/log_iterator.go
```diff
@@ -0,0 +1,289 @@
...
+		return LogLine{}, err
+	}
+
+	return LogLine{
+		Timestamp: time.Unix(0, ts*int64(time.Millisecond)).UTC(),
+		Data:      data[20:],
+	}, nil
+}
+
+func prependTimestamp(data string, t time.Time) string {
```
tychoish: nit: I'd put the timestamp before the data in the arg order...

model/log_iterator.go
```diff
@@ -0,0 +1,289 @@
...
+	ts := fmt.Sprintf("%d", util.UnixMilli(t))
+	pre_zeros := 20 - len(ts)
+	for i := 0; i < pre_zeros; i++ {
+		ts = "0" + ts
+	}
```
tychoish: https://www.dotnetperls.com/padding-go

model/log_iterator.go
```diff
@@ -0,0 +1,289 @@
...
+				catcher.Add(recovery.HandlePanicWithError(recover(), err))
+				wg.Done()
+			}()
+
+			for chunk := range work {
+				if err := ctx.Err(); err != nil {
+					catcher.Add(err)
+					return
+				} else {
+					r, err := i.bucket.Get(ctx, chunk.Key)
```
tychoish: what are the timeouts on these readers?

tychoish:
also I think we need to figure out when to use which log iterator, and we should probably add a ticket to write another iterator that's just buffered... the batched iterator is good, but I'm afraid that the next readers will hang out and go stale because there's nothing to put back pressure on reading the chunks, so that thread pool ends up finishing too soon.

PR #124, opened 2019-08-15
MAKE-778: Implement merger

julianedwards:


model/buildlogger.go
```diff
@@ -299,6 +301,42 @@ func (l *Log) Download(timeRange util.TimeRange) (LogIterator, error) {
...
+		defer close(lines)
+		for h.Len() > 0 {
+			if ctx.Err() != nil {
+				return
+			}
+
+			i := heap.Pop(h)
+			it := i.(LogIterator)
+
+			line := fmt.Sprintf("%s %s", it.Item().Timestamp, it.Item().Data)
```
julianedwards: I spoke to Max, and it looks like we can actually throw away the buildlogger generated timestamp after sorting (resmoke has its own timestamps in their logs). I kept here it for now since it may be useful  in other contexts (and easy enough to strip away or modify).

model/log_iterator.go
```diff
@@ -327,3 +327,31 @@ func filterChunks(timeRange util.TimeRange, chunks []LogChunkInfo) []LogChunkInf
...
+
+func (h LogIteratorHeap) Len() int { return len(h) }
+
+func (h LogIteratorHeap) Less(i, j int) bool {
+	return h[i].Item().Timestamp.Before(h[j].Item().Timestamp)
+}
+
+func (h LogIteratorHeap) Swap(i, j int) { h[i], h[j] = h[j], h[i] }
+
+func (h *LogIteratorHeap) Push(x interface{}) {
```
tychoish: I'd write type safe methods on this type so that you don't have to type coerce these interfaces (e.g. PushIterator and PopIterator or just Add/Next) in the calling code.

model/buildlogger.go
```diff
@@ -299,6 +301,42 @@ func (l *Log) Download(timeRange util.TimeRange) (LogIterator, error) {
...
+	heap.Init(h)
+	lines := make(chan string, len(iterators))
+
+	for _, it := range iterators {
+		if it.Next(ctx) {
+			heap.Push(h, it)
+		}
+	}
+
+	go func() {
```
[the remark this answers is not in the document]
julianedwards: added one below

model/log_iterator.go
julianedwards: added below

PR #125, opened 2019-08-16
MAKE-835: Implement gRPC StreamLog

julianedwards:


rpc/internal/buildlogger_service.go
```diff
@@ -68,8 +69,36 @@ func (s *buildloggerService) AppendLogLines(ctx context.Context, lines *LogLines
...
+
+		if id == "" {
+			id = lines.LogId
+		} else if lines.LogId != id {
+			return newRPCError(codes.Aborted, errors.New("log ID in stream does not match reference, aborting"))
+		}
+
+		_, err = s.AppendLogLines(ctx, lines)
+		if err != nil {
+			return err
```
tychoish: should we put an RPC error code on this?

julianedwards: it calls AppendLogLines which sets the rpc codes, so no need to :)

PR #128, opened 2019-08-22
MAKE-779: Implement API buildlogger model

julianedwards:


rest/model/buildlogger.go
```diff
@@ -0,0 +1,106 @@
...
+	dbmodel "github.com/evergreen-ci/cedar/model"
+	"github.com/pkg/errors"
+)
+
+// APILog describes metadata for a buildlogger log.
+type APILog struct {
+	ID          APIString          `json:"id,omitempty"`
+	Info        APILogInfo         `json:"info,omitempty"`
+	CreatedAt   APITime            `json:"created_at"`
+	CompletedAt APITime            `json:"completed_at"`
```
tychoish: should we compute a duration (float seconds) for people?

rest/model/buildlogger.go
```diff
@@ -0,0 +1,106 @@
...
+// the cedar-based log metadata storage. The prefix field indicates the name of
+// the "sub-bucket".
+type APILogArtifactInfo struct {
+	Type    APIString         `bson:"type"`
+	Prefix  APIString         `bson:"prefix"`
+	Version int               `bson:"version"`
+	Chunks  []APILogChunkInfo `bson:"chunks,omitempty"`
+}
+
+func getLogArtifactInfo(l dbmodel.LogArtifactInfo) APILogArtifactInfo {
```
tychoish: I feel like these import/export operations should probably be methods?

julianedwards: They're just helper functions, I don't think it is really useful to make them methods on APILog... it actually makes the code more awkward. We do the same thing already with APIPerformanceResult.

rest/model/buildlogger.go
julianedwards: is that something useful? also, can't that just be calculated in the UI?

PR #129, opened 2019-08-23
MAKE-847: Move timber benchmarking into cedar

julianedwards:


tychoish:
I think we did all of the review of this elsewhere. Should the benchmarks run in evergreen? or do that as a separate commit?

julianedwards:
I don't think benchmarks should run in evergreen, and if we ever need to turn them on we easily ca.

PR #134, opened 2019-09-03
MAKE-781: Add FindLogsByTaskId to the connector interface

julianedwards:


model/buildlogger.go
```diff
@@ -400,3 +404,148 @@ type LogLine struct {
...
+	}
+
+	if opts.TimeRange.IsZero() || !opts.TimeRange.IsValid() {
+		return errors.New("invalid time range given")
+	}
+
+	l.populated = false
+	findOpts := options.Find()
+	if opts.Limit > 0 {
+		findOpts.SetLimit(int64(opts.Limit))
```
tychoish: isn't it viable/correct to just set it to 0 and remove the if statement, also make opts.Limit int64 rather than require the conversion.

model/buildlogger.go
```diff
@@ -400,3 +404,148 @@ type LogLine struct {
...
+	}
+	findOpts.SetSort(bson.D{{logCreatedAtKey, -1}})
```
tychoish: do we need to add an index?

rest/data/buildlogger.go
```diff
@@ -17,70 +21,260 @@ import (
...
-	log.Setup(dbc.env)
-	it, err := log.Download(ctx, tr)
+	return apiLog, nil
+}
+
+// FindLogsByTaskId queries the database to find the buildlogger logs with the
+// given task id returning the merged logs via a LogIterator with the
+// corresponding time range. The number of logs is limited by limit if and only
+// if limit is greater than 0.
+func (dbc *DBConnector) FindLogsByTaskId(ctx context.Context, taskId string, tr util.TimeRange, limit int) (dbModel.LogIterator, error) {
```
tychoish: Id -> ID

model/buildlogger.go
julianedwards: yeah

tychoish: don't we have a file for that somewhere? just confirm that you build the index before merging this.

PR #138, opened 2019-09-23
MAKE-950: Compress buildlogger logs going to s3

julianedwards:


tychoish:
can you set the timeout for macos to something less than an hour :)

julianedwards:
that is the default timeout for tasks in evergreen, should we change it?

julianedwards:
default is actually 2 hours https://github.com/evergreen-ci/evergreen/blob/1c0ec7fe55d1c385095c750bd05a2a9f282bea33/agent/global.go

PR #139, opened 2019-09-24
MAKE-946: Special API endpoint for resmoke

julianedwards:


rest/buildlogger_routes.go
```diff
@@ -167,16 +172,141 @@ func (h *logMetaGetByTaskIDHandler) Factory() gimlet.RouteHandler {
...
+	it, err := h.sc.FindLogsByTestName(ctx, h.id, h.name, h.tr, h.tags...)
+	if err != nil {
+		return gimlet.MakeJSONErrorResponder(errors.Wrapf(err, "Error getting logs by test name '%s'", h.name))
+	}
+	its = append(its, it)
+
+	it, err = h.sc.FindLogsByTestName(ctx, h.id, "", h.tr, h.tags...)
+	if err == nil {
+		its = append(its, it)
+	} else if !strings.Contains(err.Error(), http.StatusText(http.StatusNotFound)) {
```
tychoish: would errors.Cause()? 

in evergreen we check for gimlet.ErrorResponse typed errors. the http status text and strings.Contains seems a bit weird. What about other errors?

rest/buildlogger_routes.go
```diff
@@ -167,16 +172,141 @@ func (h *logMetaGetByTaskIDHandler) Factory() gimlet.RouteHandler {
...
+}
+
+// Run calls FindLogMetadataByTestName and returns the merged logs.
+func (h *logMetaGetByTestNameHandler) Run(ctx context.Context) gimlet.Responder {
+	testLogs, err := h.sc.FindLogMetadataByTestName(ctx, h.id, h.name, h.tags...)
+	if err != nil {
+		return gimlet.MakeJSONErrorResponder(errors.Wrapf(err, "Error getting log metadata by test name '%s'", h.name))
+	}
+	globalLogs, err := h.sc.FindLogMetadataByTestName(ctx, h.id, "", h.tags...)
+	if err != nil && !strings.Contains(err.Error(), http.StatusText(http.StatusNotFound)) {
```
tychoish: see above about errors

rest/data/buildlogger.go
```diff
@@ -254,15 +351,117 @@ func (mc *MockConnector) FindLogMetadataByTaskID(ctx context.Context, taskID str
...
 				Message:    "corrupt data",
 			}
 		}
+
 	}
 
 	return apiLogs, ctx.Err()
 }
+
+func containsTags(subset, tags []string) bool {
```
tychoish: move to util?

rest/buildlogger_routes.go
julianedwards: This is a weird case because I want to check for a not found error (meaning the database query returned nothing) and not do anything if that happens, otherwise return an error. Since it is looking for logs with an an empty `TestName` field, I do not necessarily want to error if there are none found (because there are already test level logs found). Unfortunately the `gimlet.ErrorResponse` does implement the `Cause` interface so I can't use `errors.Cause()`.

tychoish: I think casting is more clear in this case.

PR #141, opened 2019-09-27
MAKE-954: tail function for buildlogger

julianedwards:


model/log_iterator.go
```diff
@@ -349,9 +349,10 @@ type mergingIterator struct {
 
 // NewMergeIterator returns a LogIterator that merges N buildlogger logs,
 // passed in as LogIterators, respecting the order of each line's timestamp.
-func NewMergingIterator(ctx context.Context, iterators ...LogIterator) LogIterator {
+func NewMergingIterator(ctx context.Context, reverse bool, iterators ...LogIterator) LogIterator {
```
tychoish: could we just add a .Reverse() method to the log iterator? 

bool arguments to functions always feel a bit messy and tend to proliferate.

model/log_iterator_test.go
```diff
@@ -323,6 +363,81 @@ func TestLogIteratorReader(t *testing.T) {
 	})
 }
 
+func TestLogIteratorTailReader(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	tmpDir, err := ioutil.TempDir(".", "merge-logs-test")
+	require.NoError(t, err)
+	defer func() {
+		_ = os.RemoveAll(tmpDir)
```
tychoish: should probably be a checked error

model/log_iterator_test.go
```diff
@@ -323,6 +363,81 @@ func TestLogIteratorReader(t *testing.T) {
 	})
 }
 
+func TestLogIteratorTailReader(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	tmpDir, err := ioutil.TempDir(".", "merge-logs-test")
```
tychoish: I think I'd not do . and just rely on TMPDIR being set correctly.

model/log_iterator.go
julianedwards: what if instead I create a separate constructor functions, for example, `NewReverseMergingIterator`?

rest/data/buildlogger.go
```diff
@@ -182,24 +199,43 @@ func (dbc *DBConnector) FindLogsByTestName(ctx context.Context, taskID, testName
...
 }
 
-// FindLogMetadataByTestName queries the database to find the buildlogger logs
-// the given task id, test name, and optional tags, returning only the metadata
-// for those logs.
-func (dbc *DBConnector) FindLogMetadataByTestName(ctx context.Context, taskID, testName string, tags ...string) ([]model.APILog, error) {
+// FindResmokeLogs is a special function for resmoke test logs that queries the
+// database to find logs based on the given task id, test name, time range,
+// "group id", and optional tags.
+func (dbc *DBConnector) FindResmokeLogs(ctx context.Context, taskID, testName, groupID string, tr util.TimeRange, tags ...string) (io.Reader, error) {
```
tychoish: I think we should have a better name for this.

rest/data/interface.go
```diff
@@ -43,21 +43,25 @@ type Connector interface {
...
 	// id, test name, and optional tags.
-	FindLogsByTestName(context.Context, string, string, util.TimeRange, ...string) (dbModel.LogIterator, error)
+	FindLogsByTestName(context.Context, string, string, util.TimeRange, ...string) (io.Reader, error)
 	// FindLogsByTestName returns the metadata for the buildlogger logs
 	// with the given task id, test name, and optional tags.
 	FindLogMetadataByTestName(context.Context, string, string, ...string) ([]model.APILog, error)
+	// FindResmokeLogs is a special function for resmoke test logs that
+	// returns logs based on the given task id, test name, time range,
+	// "group id", and optional tags.
+	FindResmokeLogs(context.Context, string, string, string, util.TimeRange, ...string) (io.Reader, error)
```
tychoish: I think we shouldn't call this remoke.

rest/service.go
```diff
@@ -237,4 +237,5 @@ func (s *Service) addRoutes() {
 	s.app.AddRoute("/buildlogger/meta/task_id/{task_id}").Version(1).Get().RouteHandler(makeGetLogMetaByTaskID(s.sc))
 	s.app.AddRoute("/buildlogger/task_name/{task_id}/{test_name}").Version(1).Get().RouteHandler(makeGetLogByTestName(s.sc))
 	s.app.AddRoute("/buildlogger/meta/test_name/{task_id}/{test_name}").Version(1).Get().RouteHandler(makeGetLogMetaByTestName(s.sc))
+	s.app.AddRoute("/buildlogger/resmoke/{task_id}/{test_name}/{group_id}").Version(1).Get().RouteHandler(makeGetLogResmoke(s.sc))
```
tychoish: namechange.

model/log_iterator.go
```diff
@@ -29,6 +29,8 @@ type LogIterator interface {
 	Err() error
 	// Item returns the current LogLine item held by the iterator.
 	Item() LogLine
+	// Reverse reverses the iterator.
+	Reverse() error
```
tychoish: I feel like this should return a LogIterator or you should be able to pass an option which is either "FORWARD" or "REVERSE" and have it error if you've started iterating?

rest/service.go
```diff
@@ -237,5 +237,5 @@ func (s *Service) addRoutes() {
 	s.app.AddRoute("/buildlogger/meta/task_id/{task_id}").Version(1).Get().RouteHandler(makeGetLogMetaByTaskID(s.sc))
 	s.app.AddRoute("/buildlogger/task_name/{task_id}/{test_name}").Version(1).Get().RouteHandler(makeGetLogByTestName(s.sc))
 	s.app.AddRoute("/buildlogger/meta/test_name/{task_id}/{test_name}").Version(1).Get().RouteHandler(makeGetLogMetaByTestName(s.sc))
-	s.app.AddRoute("/buildlogger/resmoke/{task_id}/{test_name}/{group_id}").Version(1).Get().RouteHandler(makeGetLogResmoke(s.sc))
+	s.app.AddRoute("/buildlogger/group/{task_id}/{test_name}/{group_id}").Version(1).Get().RouteHandler(makeGetLogGroup(s.sc))
```
tychoish: does it make sense to put /meta and /group at the end rather than the beginning?

PR #144, opened 2019-10-07
MAKE-763: Use new-style s3 URLs

julianedwards:


tychoish:
what are these test failures? seems fine otherwise.

PR #146, opened 2019-11-13
MAKE-1020: Add change point detector to cedar

ancostas:
I've probably done everything wrong since this is my first golang experience. Feel free to laugh 😬 .


tychoish:
can you move your ticket to the MAKE project and change the title? cedar is public and TIG isn't.

perf/change_point_detector.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+	"github.com/evergreen-ci/cedar"
+	"io"
+	"net/http"
+)
+
+type Detector interface {
```
tychoish: ChangeDetector

perf/change_point_detector.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+
+func NewDefaultDetector() Detector {
+	return NewDetector(http.DefaultClient.Do, http.NewRequest, cedar.GetEnvironment().GetConf().SignalProcessingURI, cedar.GetEnvironment().GetConf().SignalProcessingToken)
+}
+
+func NewDetector(requestExecutor func(req *http.Request) (*http.Response, error), requestMaker func(method, url string, body io.Reader) (*http.Request, error), baseURL string, token string) Detector {
+	executor := func(method, url string, body io.Reader) (*http.Response, error) {
+		req, err := requestMaker("POST", baseURL+"/change_points/detect", body)
```
tychoish: this function isn't defined?

config.go
```diff
@@ -15,6 +15,8 @@ type Configuration struct {
 	DatabaseName       string
 	QueueName          string
 	MongoDBURI         string
+	SignalProcessingURI string
```
tychoish: as of our last discussion, you were going to use a golang implementation that didn't require accessing an external service. 

Additionally, and this is a confusing quirk, but there are two levels of configuration... this one which controls lower level service behavior and is configured per-process, and a configuration in the database (model.Configuration) that has all of the shared configuration. These values should go there.

perf/change_point_detector.go
```diff
@@ -0,0 +1,74 @@
...
+	return signalProcessingClient{
+		execute: executor,
+	}
+}
+
+type detectChangesResponse struct {
+	ChangePoints []ChangePoint
+}
+
+func (client signalProcessingClient) DetectChanges(series []float64) ([]ChangePoint, error) {
```
tychoish: you almost certainly want this to be a pointer receiver.

perf/change_point_detector.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+type Algorithm struct {
+	Name          string
+	Version       string
+	Configuration map[string]interface{}
+}
+
+type signalProcessingClient struct {
+	execute    func(method, url string, body io.Reader) (*http.Response, error)
```
tychoish: this feels like too much indirection, this probably doesn't need to be a type, you can just have different detector implementations and do your dependency injection at a higher level.

tychoish: additionally would it be possible, as this gets refactored, to change the way we pass data into this operation from a reader into something easier to introspect?

perf/change_point_detector.go
```diff
@@ -0,0 +1,74 @@
...
+
+type ChangePoint struct {
+	Index     int
+	Algorithm Algorithm
+}
+
+type Algorithm struct {
+	Name          string
+	Version       string
+	Configuration map[string]interface{}
```
tychoish: would it be possible to have Configuration use a non-map type?

perf/change_point_detector.go
ancostas: `requestMaker`? I think it's passed in to `NewDetector` as an arg. There is probably a better way to set this up in a way that can be mocked.

perf/change_point_detector.go
ancostas: Originally I didn't have this on there at all, but it seemed Like I needed this available in order to pass a mock wrapping `http.NewRequest`/setting up base url + header/`http.DefaultClient.Do`. Is there a better way to mock the client?

perf/change_point_detector.go
ancostas: The configuration is basically dynamic (varies depending on the algorithm being used). Not sure what would be better in this case.

config.go
ancostas: I think we closed out the thread with "we'll go with whatever is best" 😅 . Can we go with a python service? There is active development on the python implementation and it is not out of the question that it may even be completely replaced by some other python algo -- ideally this would be irrelevant to cedar.

Also w.r.t. the configuration --> should these not be passed as some kind of secret to the environment for cedar's processes? The token is sensitive and probably shouldn't live in a database.

perf/change_point_detector.go
tychoish: I would just implement types and factor the methods on that so that you don't really need to mock things? Having indirection through an interface and a function type is overkill and just makes it really hard to figure out what's going on, and you can get almost all of the coverage you want by factoring you code into clearer units.

config.go
tychoish: We put all kinds of tokens in the database and there's no better way to do that at the moment. 

Also my memory was "we'll support multiple implementations, but starting with the golang one," and looking back at the design document this seems to be the state of things.

perf/change_point_detector.go
tychoish: http clients are even worse than this. I wouldn't really mock things at the level of the client.

perf/change_point_detector.go
tychoish: shouldn't this match something that we pass into the ChangeDetector?

perf/change_point_detector.go
```diff
@@ -0,0 +1,74 @@
...
+	resp, err := client.execute("POST", "/change_points/detect", bytes.NewReader(body))
+	if err != nil {
+		return nil, err
+	}
+	var decoded detectChangesResponse
+	err = json.NewDecoder(resp.Body).Decode(&decoded)
+	if err != nil {
+		return nil, err
+	}
+	err = resp.Body.Close()
```
bsamek: This might not run if you return on 67. It might be better to `defer resp.Body.Close()` immediately after error checking `client.execute`. (I suspect the error returned by `Close()` is uninteresting, so a [defer](https://gobyexample.com/defer) might make more sense.

tychoish:
There've been a bunch of comments on this, and I just wanted to summarize: 

# HTTP Clients
I've pushed a thing in order to produce HTTP clients and also pool them (because they're a resource leak waiting to happen.) basically do the following: 

```
func () {
client := util.GetHTTPClient()
defer util.PutHTTPClient(client)
client.Do(<magic>)
}()
```

# Golang Implementation of Change Points
We can put it in a different pull request, if it makes the most sense, but the pure golang change points detection needs to be done at some point. 

# Robustness and Operationalization 
The current implementation should be able to retry HTTP requests either using rehttp or a similar library, or by writing some kind of jittery/timeout friendly request retryer (i'd use the library. 

I'd also want you to, in some way, pass a context to anything making a network request.

tychoish:
I've pushed a commit on this ticket to include the go implementation of the change point detector. From my perspective this PR can be deferred behind the work to wire the change point detection to the actual data. The approach in this code could be taken at a later point to further validate the go algorithm against the python algorithm, but that doesn't seem to be a pressing concern at the moment.

ancostas:
Quick note to reviewers: for a random metric selected from our legacy perf cluster, calculating change points over the entire history of that metrics takes ~400ms in staging (including network latency.) Funnily enough, calculating the same change points in my Jupyter notebook takes ~1 sec for the same time series. 

The metric in question:
```
{
	"project" : "sys-perf-4.0",
	"variant" : "linux-3-node-replSet-initialsync",
	"task" : "initialsync",
	"test" : "canary_server-cpuloop-10x",
        "thread_level": "1"
}
```

This should hopefully be good enough, when scaled horizontally (e.g. using auto-scaling). Should the processing time begin to grow too long due to sheer length of the series being processed (somewhat unlikely since we are processing smaller windows after the initial setup), we may like to investigate other algorithms for making initial passes over the data -- but that is a separate story altogether.

perf/change_point_detector.go
ancostas: Sure, but each algorithm will have its own unique configuration --> I'm not sure how best to represent that in golang on an abstract level; for a given specific algorithm we can obviously specify exactly how to configure it, but it would only be conforming to the structure/interface provided by this abstraction?

ancostas:
@tychoish I've moved configuration into the database. What kind of service/feature flags did you have in mind for this thing --> algorithm configuration? Retrying policy? The former is currently an addition to be made downstream, but I could add some configurata for the latter.

ancostas:
@tychoish with regards to retrying http requests, is this something the default client you've added should do on its own? Also, what do you mean when you request the ability to pass context to the default client?

perf/change_point_detector.go
```diff
@@ -0,0 +1,79 @@
...
+
+type ChangePoint struct {
+	Index     int
+	Algorithm Algorithm
+}
+
+type Algorithm struct {
+	Name          string
+	Version       string
+	Configuration map[string]float64
```
tychoish: Having thought about this more, I'm pretty convinced that this ought to be a slice of key value pairs.

perf/change_point_detector.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+type ChangePoint struct {
+	Index     int
+	Algorithm Algorithm
+}
+
+type Algorithm struct {
+	Name          string
+	Version       string
```
tychoish: version should be an integer.

perf/change_point_detector.go
```diff
@@ -0,0 +1,79 @@
...
+		return errors.WithStack(err)
+	}
+
+	client := util.GetHTTPClient()
+	defer util.PutHTTPClient(client)
+
+	req, err := http.NewRequest(method, route, bytes.NewBuffer(body))
+	if err != nil {
+		return errors.WithStack(err)
+	}
```
tychoish: add: 

`req = req.WithContext(ctx)`

and add contexts to the signatures up through and including the interface.

perf/change_point_detector.go
```diff
@@ -0,0 +1,79 @@
...
+	req.Header.Add("Content-Type", "application/json")
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
+		return errors.New("Failed to detect changes in metric data, status: "+string(resp.StatusCode))
```
tychoish: use errors.Errorf() rather than string concatenation to format the error string. 

there's also http.StatusText() to get the thing you want. I'm not sure that casting an int to a string does exactly what you want. 

are there status codes in the 200 or 300 range that we should handle? are there codes in the 500 range that mean that we should retry with backoff? this question also applies to the error condition on line 66. 

There is some existing infrastructure in evergreen that we use to make each implementation of retries with backoff better, but you should decide what the conditions that are retriable and then use those tools, which I'm happy to help with.

perf/change_point_detector.go
```diff
@@ -0,0 +1,79 @@
+package perf
+
+import (
+	"bytes"
+	"encoding/json"
+	"github.com/evergreen-ci/cedar/util"
+	"github.com/evergreen-ci/gimlet"
+	"github.com/pkg/errors"
+	"net/http"
```
tychoish: run goimports.

model/config.go
```diff
@@ -44,13 +45,24 @@ func NewCedarConfig(env cedar.Environment) *CedarConfig {
...
+	cedarConfigurationSplunkKey         = bsonutil.MustHaveTag(CedarConfig{}, "Splunk")
+	cedarConfigurationSlackKey          = bsonutil.MustHaveTag(CedarConfig{}, "Slack")
+	cedarConfigurationLDAPKey           = bsonutil.MustHaveTag(CedarConfig{}, "LDAP")
+	cedarConfigurationCAKey             = bsonutil.MustHaveTag(CedarConfig{}, "CA")
+	cedarConfigurationFlagsKey          = bsonutil.MustHaveTag(CedarConfig{}, "Flags")
+	cedarConfigurationServiceKey        = bsonutil.MustHaveTag(CedarConfig{}, "Service")
+	cedarConfigurationChangeDetectorKey = bsonutil.MustHaveTag(CedarConfig{}, "ChangeDetector")
+)
+
+type ChangeDetectorConfig struct {
```
tychoish: should we add an implementation name/string?

perf/change_point_detector.go
```diff
@@ -0,0 +1,79 @@
...
+	Version       string
+	Configuration map[string]float64
+}
+
+type signalProcessingClient struct {
+	token   string
+	baseURL string
+}
+
+func NewChangeDetector(baseURL, token string) ChangeDetector {
```
tychoish: can this have a slightly less generic name, given that it can only construct a single type of detector?

perf/change_point_detector.go
ancostas: Is that not just a map without the ability to look up by key in O(1) time? I may be missing something w.r.t. golang types.

perf/change_point_detector.go
ancostas: I've added this one for the retrying stuff and error handling story: https://jira.mongodb.org/browse/MAKE-1050

We will obviously do this before this goes operational, I'd just like to divorce it from this ticket since it can be framed as an improvement to the naive client (which believes nothing will go wrong, but shares the same interface with a client that gracefully handles recoverable errors).

model/config.go
ancostas: I think I did what you were asking for, let me know if I misinterpreted what you meant.

model/config.go
```diff
@@ -56,8 +56,10 @@ var (
 )
 
 type ChangeDetectorConfig struct {
-	URI   string `bson:"uri" json:"uri" yaml:"uri"`
-	Token string `bson:"token" json:"token" yaml:"token"`
+	ImplementationName string `bson:"implementation_name" json:"implementation_name" yaml:"implementation_name"`
```
tychoish: rename to "implementation" in both the field name and the tag.

model/config.go
```diff
@@ -56,8 +56,10 @@ var (
...
+	URI                string `bson:"uri" json:"uri" yaml:"uri"`
+	Token              string `bson:"token" json:"token" yaml:"token"`
+	IsEnabled          bool   `bson:"is_enabled" json:"is_enabled" yaml:"is_enabled"`
```
tychoish: I would put the is_enabled flag in the service flag area of the code (see earlier comment) rather than here.

perf/change_point_detector.go
tychoish: I think this is the correct structure: 

```
type ChangePoint struct {
	Index int           `bson:"index" json:"index" yaml:"index"`
	Info  AlgorithmInfo `bson:"info" json:"info" yaml:"info"`
}

type AlgorithmInfo struct {
	Name    string            `bson:"name" json:"name" yaml:"name"`
	Version int               `bson:"version" json:"version" yaml:"version"`
	Options []AlgorithmOption `bson:"options" json:"options" yaml:"options"`
}

type AlgorithmOption struct {
	Name  string      `bson:"name" json:"name" yaml:"name"`
	Value interface{} `bson:"value" json:"value" yaml:"value"`
}
```

perf/change_point_detector.go
```diff
@@ -36,15 +52,37 @@ func NewMicroServiceChangeDetector(baseURL, token string) ChangeDetector {
...
 
-	return changePoints.ChangePoints, nil
+	var result []ChangePoint
+	for _, point := range jsonChangePoints.ChangePoints {
+		mapped := ChangePoint{
+			Index: point.Index,
+			Info: AlgorithmInfo{
+				Name:    point.Algorithm.Name,
+				Version: point.Algorithm.Version,
+				Options: nil,
```
tychoish: you can ommit this.

perf/change_point_detector.go
```diff
@@ -36,15 +52,37 @@ func NewMicroServiceChangeDetector(baseURL, token string) ChangeDetector {
...
+
+		for k, v := range point.Algorithm.Configuration {
+			additionalOption := AlgorithmOption{
+				Name:  k,
+				Value: v,
+			}
+			mapped.Info.Options = append(mapped.Info.Options, additionalOption)
+		}
+
+		result = append(result, mapped)
```
tychoish: do you plan to change the results from the service to use the new format?

config.go
```diff
@@ -11,15 +11,16 @@ import (
...
+	BucketName                string
+	DatabaseName              string
+	QueueName                 string
+	MongoDBURI                string
+	MongoDBDialTimeout        time.Duration
+	SocketTimeout             time.Duration
+	DisableLocalQueue         bool
+	DisableRemoteQueue        bool
+	NumWorkers                int
+	SignalProcessingIsEnabled bool
```
tychoish: sorry, put this in config_flags.go

ancostas:
@tychoish Retrying is lumped in with error handling in https://jira.mongodb.org/browse/MAKE-1050 (see previous comment). Will add some logging I suppose.

perf/change_point_detector.go
ancostas: Not at the moment; is there some compelling reason to do this everywhere vs what the service currently exposes?

perf/change_point_detector.go
```diff
@@ -0,0 +1,136 @@
...
+	token   string
+	baseURL string
+}
+
+func NewMicroServiceChangeDetector(baseURL, user string, token string) ChangeDetector {
+	return &signalProcessingClient{user: user, token: token, baseURL: baseURL}
+}
+
+func (spc *signalProcessingClient) DetectChanges(series []float64, ctx context.Context) ([]ChangePoint, error) {
+	grip.Debug(`Detecting change points.`)
```
tychoish: omit this, probably? or add context as below?

perf/change_point_detector.go
tychoish: I think this is probably the best way to store the data, and given that you control the service on the other end, I think it probably makes sense to avoid reparsing all of your data, and these kinds of API changes are likely to be easier to make now rather than later.

perf/change_point_detector.go
```diff
@@ -0,0 +1,136 @@
...
+				Name:  k,
+				Value: v,
+			}
+			mapped.Info.Options = append(mapped.Info.Options, additionalOption)
+		}
+
+		result = append(result, mapped)
+	}
+
+	grip.Debug(`Change point detection completed.`)
```
tychoish: can you do 

```grip.Debug(message.Fields{
   "message": "change point detection completed",
   "num_points": len(points),
   "cp_detected": len(cp),
   "duration_secs": time.Since(startAt).Seconds(),
   "implementation": <>,
})
```
Or something to the effect.

perf/change_point_detector.go
```diff
@@ -0,0 +1,136 @@
...
+
+	req, err := http.NewRequest(method, route, bytes.NewBuffer(body))
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	req.WithContext(ctx)
+	req.Header.Add("Cookie", fmt.Sprintf("auth_user=%v;auth_token=%v", spc.user, spc.token))
+	req.Header.Add("Content-Type", "application/json")
+
+	grip.Debugf(`Requesting change points from %q.`, route)
```
tychoish: I'd prefer some more structured logging rather than formatted strings. also the %q is a bit of an anomaly as are the backticks

tychoish: on reflection I might only log in the outer function, and not here.

ancostas:
@tychoish Updated the logging as requested and fmted the config files that needed fmting.

W.r.t. the other points:
* https://jira.mongodb.org/browse/MAKE-1050 includes adding retrying logic as I mentioned before; we won't be handling non-naive cases in this tix
* At this point the service has been consumed by other works; we can just consume the contract it provides without requiring a lockstep release for all the other works --> we can discuss changing the contract as a separate ticket but that should probably flow through the planning process of DAG team

ancostas:
evergreen merge

PR #149, opened 2019-11-20
MAKE-1024: new evergreen configuration

tychoish:


bsamek:
There's an error that the config file is invalid. Maybe `evergreen validate`?

PR #150, opened 2019-12-05
MAKE-1051: API route that prints logs with timestamp

julianedwards:


model/log_iterator.go
```diff
@@ -633,23 +640,25 @@ func (r *logIteratorReader) writeToBuffer(data, buffer []byte, n int) int {
...
+	it        LogIterator
+	n         int
+	printTime bool
+	r         io.Reader
 }
 
 // NewLogIteratorTailReader returns an io.Reader that reads up to n log lines
 // from the *reversed* log iterator.
-func NewLogIteratorTailReader(ctx context.Context, it LogIterator, n int) io.Reader {
+func NewLogIteratorTailReader(ctx context.Context, it LogIterator, n int, printTime bool) io.Reader {
```
tychoish: I'd leave the implementation as it is, but avoid having the bool arguments and just make different constructors. 

Alternatly we could have one constructor for both implementations and have a log-iterator-args struct.

rest/buildlogger_routes.go
```diff
@@ -133,6 +136,7 @@ func (h *logGetByTaskIDHandler) Parse(_ context.Context, r *http.Request) error
 	h.id = gimlet.GetVars(r)["task_id"]
 	vals := r.URL.Query()
 	h.tags = vals["tags"]
+	h.printTime = vals.Get("printTime") == "true"
```
tychoish: should printTime be a constant?

rest/data/interface.go
```diff
@@ -42,27 +42,32 @@ type Connector interface {
 	// Buildlogger Log
 	//////////////////
 	// FindLogByID returns the buildlogger log with the given id as a
-	// LogIterator with the corresponding time range.
-	FindLogByID(context.Context, string, util.TimeRange) (io.Reader, error)
+	// LogIterator with the corresponding time range. If the bool is set to
+	// true, the log lines should be concatenated with their timestamps.
+	FindLogByID(context.Context, string, util.TimeRange, bool) (io.Reader, error)
```
tychoish: I'm not a huge fan of printTime as a boolean argument, but aside from putting these all in as arg structs I don't really have a better idea?

PR #151, opened 2019-12-09
MAKE-815: Distinguish public read and authorized read for logs

julianedwards:


model/config.go
```diff
@@ -18,15 +18,16 @@ import (
...
-	Splunk    send.SplunkConnectionInfo `bson:"splunk" json:"splunk" yaml:"splunk"`
-	Slack     SlackConfig               `bson:"slack" json:"slack" yaml:"slack"`
-	LDAP      LDAPConfig                `bson:"ldap" json:"ldap" yaml:"ldap"`
-	NaiveAuth NaiveAuthConfig           `bson:"naive_auth" json:"naive_auth" yaml:"naive_auth"`
-	CA        CAConfig                  `bson:"ca" json:"ca" yaml:"ca"`
-	Bucket    BucketConfig              `bson:"bucket" json:"bucket" yaml:"bucket"`
-	Flags     OperationalFlags          `bson:"flags" json:"flags" yaml:"flags"`
-	Service   ServiceConfig             `bson:"service" json:"service" yaml:"service"`
+	ID           string                    `bson:"_id" json:"id" yaml:"id"`
+	EvergreenURL string                    `bson:"evergreen_url" json:"evergreen_url" yaml:"evergreen_url"`
```
tychoish: I kind of want to avoid putting things like this in the top level. Perhaps we should put it in its own sub-struct so we can put auth tokens or whatever there.

rest/buildlogger_routes.go
```diff
@@ -151,6 +181,15 @@ func (h *logGetByTaskIDHandler) Parse(_ context.Context, r *http.Request) error
 
 // Run calls FindLogsByTaskID and returns the merged logs.
 func (h *logGetByTaskIDHandler) Run(ctx context.Context) gimlet.Responder {
+	apiLogs, err := h.sc.FindLogMetadataByTaskID(ctx, h.id)
+	if err != nil {
+		return gimlet.MakeJSONErrorResponder(errors.Wrapf(err, "Error getting log metadata by task id '%s'", h.id))
+	}
+	resp := h.sc.EvergreenProxyAuthLogRead(ctx, h.userToken, h.evgURL, *apiLogs[0].Info.Project)
```
tychoish: the *apiLogs[0] looks like it could panic really easily.

rest/data/auth.go
```diff
@@ -0,0 +1,91 @@
...
+// the user is authorized, nil is returned, otherwise a gimlet.Responder with
+// the appropriate http error code is returned.
+func (dbc *DBConnector) EvergreenProxyAuthLogRead(ctx context.Context, userToken, baseURL, resourceId string) gimlet.Responder {
+	urlString := fmt.Sprintf("%s?resource=%s&resource_type=project&permission=project_logs&required_level=10", baseURL, resourceId)
+	req, err := http.NewRequest(http.MethodGet, urlString, nil)
+	if err != nil {
+		return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "error creating http request"))
+	}
+	req = req.WithContext(ctx)
+	req.Header = map[string][]string{"mci-token": {userToken}}
```
tychoish: I think there's a method to add a key to the header that's better to use than the map itself. 

also put "mci-token" in a constant somewwhere.

rest/data/auth.go
```diff
@@ -0,0 +1,91 @@
...
+	defer util.PutHTTPClient(client)
+
+	b := &backoff.Backoff{
+		Min:    100 * time.Millisecond,
+		Max:    10 * time.Second,
+		Factor: 2,
+	}
+	var responder gimlet.Responder
+	for i := 0; i < 10; i++ {
+		resp, err := client.Do(req)
```
tychoish: do you have to reset the request if it fails? I'm not sure. We have a function object wrapper that does the retry logic in evergreen so I'm not sure.

rest/buildlogger_routes.go
```diff
@@ -29,24 +30,29 @@ type logGetByIDHandler struct {
...
-		sc: h.sc,
+		sc:     h.sc,
+		evgURL: h.evgURL,
 	}
 }
 
 // Parse fetches the id and time range from the http request.
 func (h *logGetByIDHandler) Parse(_ context.Context, r *http.Request) error {
 	h.id = gimlet.GetVars(r)["id"]
+	h.userToken = r.Header.Get(cedar.AuthTokenCookie)
```
tychoish: I feel like this token isn't going to be the right token for evergreen? don't we need to see if we can get the mci auth token, not the cedar auth token?

julianedwards: Maybe not... it was to my understanding that since we use LDAP & the evergreen group for authentication the auth cookie token would be the same.

rest/data/auth.go
julianedwards: should it be a global constant?

rest/buildlogger_routes.go
julianedwards: yeah the connector functions will return a not found error if there are no logs found (so this is guaranteed to always have at least one log).

rest/data/auth.go
julianedwards: Requests are safe to use: "Or we just say it's only safe to reuse a Request if Body is nil or the the Request.Body has been read to EOF. (the caller could wrap the Body in their own wrapper which notes EOF if they wanted)." from Brad Fitz here https://github.com/golang/go/issues/19653.

rest/buildlogger_routes.go
tychoish: pretty sure that the token/cookie content is a random string generated in gimlet code during authentication... it wouldn't be the same between different applications.

PR #152, opened 2019-12-11
MAKE-1061: add retryable client

tychoish:


util/http.go
```diff
@@ -48,12 +50,84 @@ func GetHTTPClient() *http.Client { return httpClientPool.Get().(*http.Client) }
...
+	}
+
+	client.Transport = rehttp.NewTransport(client.Transport,
+		rehttp.RetryAll(retryFns...),
+		rehttp.ExpJitterDelay(conf.BaseDelay, conf.MaxDelay))
+
+	return client
+}
+
+func GetDefultHTTPRetryableClient() *http.Client {
```
ancostas: typo

tychoish:
evergreen uses rehttp for interacting with the github api, and this is just a configuration layer for that, so I feel ok about it.

PR #153, opened 2019-12-12
MAKE-1060: Create backstop change point bootstrapper

ancostas:
@tychoish I will add some more tests to this since there is a lot going on, but please give me the initial feedback in the meantime.


rest/data/perf.go
```diff
@@ -233,6 +236,65 @@ func (dbc *DBConnector) FindPerformanceResultWithChildren(ctx context.Context, i
...
+				"message": "Unable to to decode aggregation result",
+				"error":   err,
+			})
+		}
+
+		job := units.NewRecalculateChangePointsJob(result)
+		queue := dbc.env.GetRemoteQueue()
+		err = queue.Put(ctx, job)
+		if err != nil {
+			grip.Error(map[string]interface{}{
```
tychoish: message.WrapError

rest/data/perf.go
```diff
@@ -233,6 +236,65 @@ func (dbc *DBConnector) FindPerformanceResultWithChildren(ctx context.Context, i
...
+	defer cur.Close(ctx)
+	if err != nil {
+		return gimlet.ErrorResponse{StatusCode: http.StatusInternalServerError, Message: fmt.Sprint("Failed to aggregate recalculation metrics")}
+	}
+	for cur.Next(ctx) {
+		var result units.MetricGrouping
+		err := cur.Decode(&result)
+		if err != nil {
+			grip.Error(map[string]interface{}{
+				"message": "Unable to to decode aggregation result",
```
tychoish: I'd use message.WrapError, rather than a map

rest/data/perf.go
```diff
@@ -233,6 +236,65 @@ func (dbc *DBConnector) FindPerformanceResultWithChildren(ctx context.Context, i
...
+				"error":   err,
+			})
+		}
+
+		job := units.NewRecalculateChangePointsJob(result)
+		queue := dbc.env.GetRemoteQueue()
```
tychoish: I wouldn't call GetRemoteQueue() in the loop.

rest/data/perf.go
```diff
@@ -233,6 +236,65 @@ func (dbc *DBConnector) FindPerformanceResultWithChildren(ctx context.Context, i
...
+		return gimlet.ErrorResponse{StatusCode: http.StatusInternalServerError, Message: fmt.Sprint("Failed to aggregate recalculation metrics")}
+	}
+	for cur.Next(ctx) {
+		var result units.MetricGrouping
+		err := cur.Decode(&result)
+		if err != nil {
+			grip.Error(map[string]interface{}{
+				"message": "Unable to to decode aggregation result",
+				"error":   err,
+			})
```
tychoish: we should probably continue/break if this is the case

rest/data/perf.go
```diff
@@ -233,6 +236,65 @@ func (dbc *DBConnector) FindPerformanceResultWithChildren(ctx context.Context, i
...
+			},
+		},
+	}
+	cur, err := db.Collection("perf_results").Aggregate(ctx, pipeline)
+	defer cur.Close(ctx)
+	if err != nil {
+		return gimlet.ErrorResponse{StatusCode: http.StatusInternalServerError, Message: fmt.Sprint("Failed to aggregate recalculation metrics")}
+	}
+	for cur.Next(ctx) {
+		var result units.MetricGrouping
```
tychoish: put the declaration outside of the loop? maybe make it a pointer? in the declaration?

rest/data/perf.go
```diff
@@ -281,7 +343,7 @@ func (mc *MockConnector) RemovePerformanceResultById(_ context.Context, id strin
 // performance results with the given taskId and that fall within interval,
 // filtered by tags.
 func (mc *MockConnector) FindPerformanceResultsByTaskId(_ context.Context, taskId string, interval util.TimeRange, tags ...string) ([]dataModel.APIPerformanceResult, error) {
-	results := []dataModel.APIPerformanceResult{}
+	var results []dataModel.APIPerformanceResult
```
tychoish: there's no reason to make this change, or ones like them below...

rest/data/perf_test.go
```diff
@@ -2,6 +2,7 @@ package data
 
 import (
 	"context"
+	"github.com/mongodb/amboy"
 	"testing"
 	"time"
```
tychoish: there's a format-on-save program that sorts imports that we typically use (goimport) that you should be using

rest/data/perf_test.go
```diff
@@ -15,6 +16,63 @@ import (
 
 const testDBName = "cedar_rest_data_test"
 
+type MockQueue struct {
```
tychoish: I think you can refactor your code such that the components produce the input (the output of the query), produce jobs, and enqueue them, and avoid the need for this mock. 

You can use the local queue, (and maybe with a noop-flag) on the job to ensure that the jobs enqueue properly, though I wouldn't. The goal of your testing should be with the behavior of the code, and I think it's safe to trust that your dependencies (e.g. the queue) behaves as you'd expect.

units/recalculate_change_points.go
```diff
@@ -0,0 +1,243 @@
...
+	"github.com/mongodb/amboy"
+	"github.com/mongodb/amboy/dependency"
+	"github.com/mongodb/amboy/job"
+	"github.com/mongodb/amboy/registry"
+)
+
+type recalculateChangePointsJob struct {
+	*job.Base      `bson:"metadata" json:"metadata" yaml:"metadata"`
+	env            cedar.Environment
+	metricGrouping MetricGrouping
```
tychoish: I think this has to be exported, othewise it won't work in the remote

units/recalculate_change_points.go
```diff
@@ -0,0 +1,243 @@
...
+	"fmt"
+	"github.com/evergreen-ci/cedar/model"
+	"github.com/evergreen-ci/cedar/perf"
+	"github.com/mongodb/grip"
+	"sort"
+	"time"
+
+	"github.com/evergreen-ci/cedar"
+	"github.com/evergreen-ci/cedar/util"
+	"gopkg.in/mgo.v2/bson"
```
tychoish: don't use the legacy bson library, also omit the blank line below here.

units/recalculate_change_points.go
```diff
@@ -0,0 +1,243 @@
+package units
+
+import (
+	"context"
+	"fmt"
+	"github.com/evergreen-ci/cedar/model"
+	"github.com/evergreen-ci/cedar/perf"
+	"github.com/mongodb/grip"
```
tychoish: these should be sorted correctly.

units/recalculate_change_points.go
```diff
@@ -0,0 +1,243 @@
...
+	return bson.M{
+		"info.project":   grouping.Id.Project,
+		"info.variant":   grouping.Id.Variant,
+		"info.task_name": grouping.Id.Task,
+		"info.test_name": grouping.Id.Test,
+	}
+
+}
+
+func makeTimeSeriesPipeline(grouping MetricGrouping) []bson.M {
```
tychoish: these queries should be in the model package. the units package should only have the business logic.

units/recalculate_change_points.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+func (j *recalculateChangePointsJob) Run(ctx context.Context) {
+	defer j.MarkComplete()
+	if j.env == nil {
+		j.env = cedar.GetEnvironment()
+	}
+	conf := model.NewCedarConfig(j.env)
+	err := conf.Find()
+	if err != nil {
```
tychoish: should we return?

units/recalculate_change_points.go
```diff
@@ -0,0 +1,243 @@
...
+					"order": "$order",
+				},
+			},
+		},
+	})
+
+	return pipeline
+}
+
+type errorMsg = map[string]interface{}
```
tychoish: just use message.Wrap(error, message.Fields{})

units/recalculate_change_points.go
```diff
@@ -0,0 +1,243 @@
...
+	conf := model.NewCedarConfig(j.env)
+	err := conf.Find()
+	if err != nil {
+		grip.Error(errorMsg{
+			"message": "Unable to get cedar configuration",
+			"error":   err,
+		})
+	}
+	detector := perf.NewMicroServiceChangeDetector(conf.ChangeDetector.URI, conf.ChangeDetector.User, conf.ChangeDetector.Token)
+	collection := j.env.GetDB().Collection("perf_results")
```
tychoish: put this query in the models package. also use constants in the model package for results.

units/recalculate_change_points.go
```diff
@@ -0,0 +1,243 @@
...
+		if err != nil {
+			grip.Error(errorMsg{
+				"message": "Unable to decode aggregated time series",
+				"project": j.metricGrouping.Id.Project,
+				"variant": j.metricGrouping.Id.Variant,
+				"task":    j.metricGrouping.Id.Task,
+				"test":    j.metricGrouping.Id.Test,
+				"error":   err,
+			})
+			continue
```
tychoish: why not break? is it going to  get better

rest/service.go
```diff
@@ -230,6 +230,7 @@ func (s *Service) addRoutes() {
 	s.app.AddRoute("/perf/task_name/{task_name}").Version(1).Get().RouteHandler(makeGetPerfByTaskName(s.sc))
 	s.app.AddRoute("/perf/version/{version}").Version(1).Get().RouteHandler(makeGetPerfByVersion(s.sc))
 	s.app.AddRoute("/perf/children/{id}").Version(1).Get().RouteHandler(makeGetPerfChildren(s.sc))
+	s.app.AddRoute("/perf/signal_processing/recalculate").Version(1).Get().RouteHandler(makePerfSignalProcessingRecalculate(s.sc))
```
tychoish: let's call it "change_points"

rest/perf_routes.go
```diff
@@ -283,3 +283,41 @@ func (h *perfGetChildrenHandler) Run(ctx context.Context) gimlet.Responder {
...
+	var err error
+	return errors.Wrap(err, "failed to parse request")
+}
+
+func (h *perfSignalProcessingRecalculateHandler) Run(ctx context.Context) gimlet.Responder {
+	err := h.sc.ScheduleSignalProcessingRecalculateJobs(ctx)
+	if err != nil {
+		return gimlet.MakeJSONErrorResponder(errors.Wrapf(err, "Error scheduling signal processing recalculation jobs"))
+	}
+	response := &struct {
```
tychoish: I would just return an empty document

rest/perf_routes.go
```diff
@@ -283,3 +283,41 @@ func (h *perfGetChildrenHandler) Run(ctx context.Context) gimlet.Responder {
...
+
+func (h *perfSignalProcessingRecalculateHandler) Factory() gimlet.RouteHandler {
+	return &perfSignalProcessingRecalculateHandler{
+		sc: h.sc,
+	}
+}
+
+func (h *perfSignalProcessingRecalculateHandler) Parse(_ context.Context, r *http.Request) error {
+	var err error
+	return errors.Wrap(err, "failed to parse request")
```
tychoish: this always returns nil,.

rest/perf_routes.go
```diff
@@ -283,3 +283,41 @@ func (h *perfGetChildrenHandler) Run(ctx context.Context) gimlet.Responder {
 	}
 	return gimlet.NewJSONResponse(perfResults)
 }
+
+///////////////////////////////////////////////////////////////////////////////
+//
+// POST /perf/signal_processing/recalculate
```
tychoish: While I think having a manual switch to recalculate everything is good, there should be a background process to add the backfil on some interval. The application has all the information it needs to be figure out what and when to recalculate without requiring an external process to pull this lever.

rest/data/perf.go
ancostas: Doesn't that also also accept `interface{}`? How should I be instantiating this `interface{}` if not with a map? I'm guessing there is a much simpler way that has eluded me 😅

rest/data/perf.go
ancostas: I see, thought it was just getting a lazily loaded instance; moved it up regardless.

rest/data/perf.go
ancostas: moved it out

rest/data/perf.go
ancostas: IDE was complaining about it --> I think this marginally saves some memory that would otherwise be allocated: https://stackoverflow.com/questions/29164375/correct-way-to-initialize-empty-slice.

rest/data/perf_test.go
ancostas: added goimports and go fmt to apply on save

rest/perf_routes.go
ancostas: Why?

rest/perf_routes.go
ancostas: What am I supposed to return? I don't really need to parse anything, was going by whats on the other routes.

rest/perf_routes.go
ancostas: Added a tix for a cron based backstop https://jira.mongodb.org/browse/MAKE-1069

rest/data/perf_test.go
ancostas: don't I need some kind of mocked queue in order to assert that my code did in fact enqueue the job I expected it to?

model/perf.go
```diff
@@ -637,3 +638,151 @@ func (r *PerformanceResults) FindOutdatedRollups(ctx context.Context, name strin
 
 	return errors.WithStack(it.Close(ctx))
 }
+
+func GetTimeSeriesIds(ctx context.Context, db *mongo.Database) (*mongo.Cursor, error) {
```
tychoish: I don't think that the mongo.DB or mong.Cursor should be part of the public interface of the model package. Make these functions internal.

model/perf.go
```diff
@@ -637,3 +638,151 @@ func (r *PerformanceResults) FindOutdatedRollups(ctx context.Context, name strin
...
+	pipeline := bson.A{
```
tychoish: I've never used a bson.A, I think a []bson.M{} does the same thing (which you use below,) I'd say stay consistent here.

model/perf.go
```diff
@@ -637,3 +638,151 @@ func (r *PerformanceResults) FindOutdatedRollups(ctx context.Context, name strin
...
+		bson.M{
+			"$match": bson.M{
+				"info.order": bson.M{
```
tychoish: are these field names in pacakge level constants (bsonutil.MustHaveKey values)?

model/perf.go
```diff
@@ -637,3 +638,151 @@ func (r *PerformanceResults) FindOutdatedRollups(ctx context.Context, name strin
...
+			"_id": "$measurement",
+			"time_series": bson.M{
+				"$push": bson.M{
+					"id":    "$_id",
+					"value": "$value",
+					"order": "$order",
+				},
+			},
+		},
+	})
```
tychoish: why not just do: 

`return []bson.M{{<...>}, {<...>}}`?

The appends make it harder to read and don't serve a purpose.

model/perf.go
```diff
@@ -637,3 +638,151 @@ func (r *PerformanceResults) FindOutdatedRollups(ctx context.Context, name strin
...
+	}
+	pullUpdate := bson.M{
+		"$pull": bson.M{
+			"change_points": bson.M{
+				"measurement": timeSeriesId.Id.Measurement,
+			},
+		},
+	}
+	_, err := db.Collection(perfResultCollection).UpdateMany(ctx, seriesFilter, pullUpdate)
+	return err
```
tychoish: errors.Wrap() or errors.WithStack()...

model/perf.go
```diff
@@ -637,3 +638,151 @@ func (r *PerformanceResults) FindOutdatedRollups(ctx context.Context, name strin
...
+	return pipeline
+}
+func GetTimeSeries(ctx context.Context, db *mongo.Database, metricGrouping TimeSeriesId) (*mongo.Cursor, error) {
+	timeSeriesAggregator := makeTimeSeriesPipeline(metricGrouping)
+	return db.Collection(perfResultCollection).Aggregate(ctx, timeSeriesAggregator)
+}
+
+func ClearChangePoints(ctx context.Context, db *mongo.Database, timeSeriesId TimeSeriesId) error {
+	seriesFilter := bson.M{
+		"$match": bson.M{
```
tychoish: I don't know that $match works in update statements?

model/perf.go
```diff
@@ -637,3 +638,151 @@ func (r *PerformanceResults) FindOutdatedRollups(ctx context.Context, name strin
...
+				},
+			},
+		},
+	})
+
+	return pipeline
+}
+func GetTimeSeries(ctx context.Context, db *mongo.Database, metricGrouping TimeSeriesId) (*mongo.Cursor, error) {
+	timeSeriesAggregator := makeTimeSeriesPipeline(metricGrouping)
+	return db.Collection(perfResultCollection).Aggregate(ctx, timeSeriesAggregator)
```
tychoish: combine these lines?

model/perf.go
```diff
@@ -637,3 +638,151 @@ func (r *PerformanceResults) FindOutdatedRollups(ctx context.Context, name strin
...
+}
+
+func ClearChangePoints(ctx context.Context, db *mongo.Database, timeSeriesId TimeSeriesId) error {
+	seriesFilter := bson.M{
+		"$match": bson.M{
+			"info.project":   timeSeriesId.Id.Project,
+			"info.variant":   timeSeriesId.Id.Variant,
+			"info.task_name": timeSeriesId.Id.Task,
+			"info.test_name": timeSeriesId.Id.Test,
```
tychoish: wouldn't it make sense to add to this query, some filtering based on weather the changepoint/measurement exists. 

Currently the pull (which isn't the cheapest update operation,) has to iterate over a lot of perf_results that could be filtered out by the query (and I don't know if our query parser will rewite the query to do this.)

model/perf.go
```diff
@@ -637,3 +638,151 @@ func (r *PerformanceResults) FindOutdatedRollups(ctx context.Context, name strin
...
+	filter := bson.M{"_id": resultToUpdate}
+	newChangePoint := bson.M{
+		"measurement":   measurement,
+		"algorithm":     algorithm,
+		"calculated_at": time.Now(),
+	}
+	update := bson.M{
+		"$push": bson.M{
+			"change_points": newChangePoint,
+		},
```
tychoish: I'd combine the newChangePoint into the update statement.

rest/data/perf.go
```diff
@@ -3,12 +3,12 @@ package data
...
 	"fmt"
-	"github.com/evergreen-ci/cedar/units"
-	"github.com/mongodb/grip"
-	"go.mongodb.org/mongo-driver/bson"
 	"net/http"
 	"sort"
 
+	"github.com/evergreen-ci/cedar/units"
+	"github.com/mongodb/grip/message"
+
```
tychoish: this blank line should be omitted.

units/recalculate_change_points.go
```diff
@@ -42,200 +42,88 @@ func makeChangePointsJob() *recalculateChangePointsJob {
...
-	} `bson:"_id"`
-}
-
-func makeSeriesFilter(grouping MetricGrouping) bson.M {
-	return bson.M{
-		"info.project":   grouping.Id.Project,
-		"info.variant":   grouping.Id.Variant,
-		"info.task_name": grouping.Id.Task,
-		"info.test_name": grouping.Id.Test,
+func makeMessage(msg string, id model.TimeSeriesId) message.Fields {
```
tychoish: make this a method on your job?

units/recalculate_change_points.go
```diff
@@ -42,200 +42,88 @@ func makeChangePointsJob() *recalculateChangePointsJob {
...
 	if err != nil {
-		grip.Error(errorMsg{
-			"message": "Unable to aggregate time series",
-			"project": j.metricGrouping.Id.Project,
-			"variant": j.metricGrouping.Id.Variant,
-			"task":    j.metricGrouping.Id.Task,
-			"test":    j.metricGrouping.Id.Test,
-			"error":   err,
-		})
+		message.WrapError(err, makeMessage("Unable to aggregate time series", j.TimeSeriesId))
```
tychoish: this isn't logged, you still have to pass the message to a log statment (e.g. grip.Error()

units/recalculate_change_points.go
```diff
@@ -42,200 +42,88 @@ func makeChangePointsJob() *recalculateChangePointsJob {
...
-			grip.Error(errorMsg{
-				"message": "Unable to decode aggregated time series",
-				"project": j.metricGrouping.Id.Project,
-				"variant": j.metricGrouping.Id.Variant,
-				"task":    j.metricGrouping.Id.Task,
-				"test":    j.metricGrouping.Id.Test,
-				"error":   err,
-			})
-			continue
+			message.WrapError(err, makeMessage("Unable to decode aggregated time series", j.TimeSeriesId))
```
tychoish: actually log this.

units/recalculate_change_points.go
```diff
@@ -42,200 +42,88 @@ func makeChangePointsJob() *recalculateChangePointsJob {
...
-			"error":   err,
-		})
+		message.WrapError(err, makeMessage("Unable to aggregate time series", j.TimeSeriesId))
 		return
 	}
 	for cur.Next(ctx) {
-		var result measurementTimeSeries
-
+		var result model.TimeSeries
 		err = cur.Decode(result)
```
tychoish: I'd combine this into the if statement. Also I *think* it needs to be a pointer? Your tests should catch this...

units/recalculate_change_points.go
```diff
@@ -42,200 +42,88 @@ func makeChangePointsJob() *recalculateChangePointsJob {
...
 	}
+	if j.env == nil {
+		j.env = cedar.GetEnvironment()
+	}
+	db := j.env.GetDB()
 	detector := perf.NewMicroServiceChangeDetector(conf.ChangeDetector.URI, conf.ChangeDetector.User, conf.ChangeDetector.Token)
-	collection := j.env.GetDB().Collection("perf_results")
-	timeSeriesAggregator := makeTimeSeriesPipeline(j.metricGrouping)
-	cur, err := collection.Aggregate(ctx, timeSeriesAggregator)
+	cur, err := model.GetTimeSeries(ctx, db, j.TimeSeriesId)
```
tychoish: It's a bit out of our normal pattern to actually deal with mongo/bson types outside of the model package at all, so I might push more of this logic into the models package.

tychoish:
also my last patches to cedar passed all tests, so I think if you merge master things should pass the all tests here.

model/perf.go
ancostas: I can onlt find `bsonutil.MustHaveTag` --> where is `MustHaveKey`?

model/perf.go
ancostas: refactor type

model/perf.go
ancostas: This would make debugging much more annoying.

model/perf.go
ancostas: Added some filtering.

rest/data/perf.go
ancostas: Not sure why goimports keeps adding blank lines.

units/recalculate_change_points.go
ancostas: Why? Won't we just be adding characters for the same functionality?

units/recalculate_change_points.go
ancostas: Will add test. Combining statements gets undone by go fmt

ancostas:
@julianedwards @tychoish Updated based on the comments.

With regards to not returning a cursor from model package --> should I be pushing results from the cursor into a buffered channel and consuming it or such? How do you usually return the results from a cursor without exhausting it? Make a cursor wrapper or such?

model/perf.go
tychoish: sorry, I remembered incorrectly, MustHaveTag is the one we want.

model/perf.go
tychoish: `return db.Collection(perfResultCollection).Aggregate(ctx, makeTimeSeriesPipeline(metricGrouping))` 

shouldn't affect debug-ability?

rest/data/perf.go
tychoish: I can't imagine that matters, particularly in the mock, and I would (with the possible exception of gofmt changes) avoid making changes outside of the scope of your ticket unless needed, just to make things easier to review, and also to make looking over diffs easier later.

rest/data/perf_test.go
tychoish: there's something weird... the extra line on 9 shouldn't be here.

rest/data/perf_test.go
tychoish: you can use a local queue, and ensure that it's added. (either by checking the number of jobs before and after you call the function,) or by checking the results of the queue. 

In general, though, you can trust that the Put method works.

rest/perf_routes.go
tychoish: `return nil` is acceptable, the code above is confusing because it looks like it's returning an error, rather than nil, but you have to know about the behavior of errors.Wrap() to know what's happening here.

I think the goal should be "be as clear as possible,"

rest/perf_routes.go
tychoish: because http return codes have semantic meaning, so I'd just use one of them.

rest/service.go
```diff
@@ -232,6 +232,7 @@ func (s *Service) addRoutes() {
 	s.app.AddRoute("/perf/task_name/{task_name}").Version(1).Get().RouteHandler(makeGetPerfByTaskName(s.sc))
 	s.app.AddRoute("/perf/version/{version}").Version(1).Get().RouteHandler(makeGetPerfByVersion(s.sc))
 	s.app.AddRoute("/perf/children/{id}").Version(1).Get().RouteHandler(makeGetPerfChildren(s.sc))
+	s.app.AddRoute("/perf/change_points").Version(1).Post().RouteHandler(makePerfSignalProcessingRecalculate(s.sc))
```
tychoish: this needs authentication. at the very least add `.Wrap(checkUser)`

model/perf.go
ancostas: `makeTimeSeriesPipeline` returns something without writing to a variable; this consumes the return value without writing it to a variable. Hard to just inspect the value being passed into this if creating the value and passing it in occur on the same line.

units/recalculate_change_points.go
tychoish: It's clearer, and takes advantage of an existing namespace (in a package which may eventually have a lot of "messages" to make". 

I also think it's good ot do things like put the j.ID() in to the messages, so you can check which execution writes a given message.

units/recalculate_change_points.go
tychoish: we have lots of statments in the following form:

```
if err := cur.Decode(&result); err != nil {
    return err 
}
```

and it's fine. what are your arguments to gofmt?

model/perf.go
tychoish: I mean you can log in the first function or not? in order to debug this you'd have to insert a print statement? I'm not sure anyone has ever used the debugger with cedar.

rest/data/perf_test.go
ancostas: I mean, I do trust that it works, that's why I've mocked the thing! Is there some reason not to mock things in golang?

ancostas:
@julianedwards I think the only failing test can be ignored. Is that right? Appears to be flaky in that it only fails on windows.

model/perf.go
```diff
@@ -718,31 +737,32 @@ func makeTimeSeriesPipeline(timeSeriesId TimeSeriesId) []bson.M {
...
-				"_id": "$measurement",
-				"time_series": bson.M{
-					"$push": bson.M{
-						"id":    "$_id",
-						"value": "$value",
-						"order": "$order",
-					},
-				},
+				"value": "$rollups.stats.val",
+				"order": "$info.order",
```
tychoish: does anything here populate the _id field? 

are any of the key names stored in variables?

units/recalculate_change_points.go
```diff
@@ -63,69 +66,60 @@ func makeMessage(msg string, id model.TimeSeriesId) message.Fields {
...
 	}
 	if j.env == nil {
 		j.env = cedar.GetEnvironment()
 	}
-	db := j.env.GetDB()
-	detector := perf.NewMicroServiceChangeDetector(conf.ChangeDetector.URI, conf.ChangeDetector.User, conf.ChangeDetector.Token)
-	cur, err := model.GetTimeSeries(ctx, db, j.TimeSeriesId)
+	if j.ChangePointDetector == nil {
+		err := j.conf.Find()
+		if err != nil {
```
tychoish: don't log, just `j.AddError(errors.Wrap(err, "unable to get cedar configuration"))`

units/recalculate_change_points_test.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	return changePoints, nil
+}
+
+func TestRecalculateChangePointsJob(t *testing.T) {
+
+	if runtime.GOOS == "darwin" && os.Getenv("EVR_TASK_ID") != "" {
+		t.Skip("avoid less relevant failing test in evergreen")
+	}
```
tychoish: I mean this makes the test basically useless, and means that I'll never run your tests?

units/recalculate_change_points_test.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func tearDown(env cedar.Environment) error {
+	conf, session, err := cedar.GetSessionWithConfig(env)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	defer session.Close()
+	return errors.WithStack(session.DB(conf.DatabaseName).DropDatabase())
+}
```
tychoish: can you use the Client interface rather than the session one? 

additionally we should perhaps call close on the env if we're hoping to clear the entire state of the connection (to be called at the end of the test rather than inbetween tests.)

units/recalculate_change_points_test.go
```diff
@@ -0,0 +1,213 @@
...
+			newRollup.rollups[1].Value = float64(1000)
+		}
+		rollups = append(rollups, newRollup)
+		i++
+	}
+	return rollups
+}
+
+func init() {
+	dbName := "cedar_signal_processing"
```
tychoish: perhaps put the word test in this?

units/recalculate_change_points.go
```diff
@@ -63,69 +66,60 @@ func makeMessage(msg string, id model.TimeSeriesId) message.Fields {
...
-	if err != nil {
-		grip.Error(message.WrapError(err, message.Fields{
-			"message": "Unable to get cedar configuration",
-		}))
-		return
+	if j.conf == nil {
+		j.conf = model.NewCedarConfig(j.env)
 	}
-	if conf.Flags.DisableSignalProcessing == true {
+	if j.conf.Flags.DisableSignalProcessing == true {
```
tychoish: you don't need the == true

units/recalculate_change_points.go
```diff
@@ -63,69 +66,60 @@ func makeMessage(msg string, id model.TimeSeriesId) message.Fields {
...
 		grip.Info(makeMessage("signal processing is disabled, skipping processing", j.TimeSeriesId))
```
tychoish: in general we shouldn't log *every* time, just sometimes. 

`grip.InfoWhen(sometimes.Percent(10), makeMessage...)`

model/perf.go
ancostas: I think `_id` carries over automatically in mongodb.

Perhaps, do you mean in like `PerformanceResult` struct?

units/recalculate_change_points_test.go
ancostas: Copied from `TestFTDCRollupsJob`. Will remove here, but I guess that test is useless as well then?

units/recalculate_change_points_test.go
ancostas: Could you direct me to a test that does this the way you are looking for? I just copied this from `TestFTDCRollupsJob`.

units/recalculate_change_points.go
```diff
@@ -5,6 +5,10 @@ import (
 	"fmt"
 	"sort"
 
+	"github.com/mongodb/grip/sometimes"
+
+	"github.com/pkg/errors"
+
```
tychoish: whitespace?

model/perf.go
tychoish: bump

model/perf.go
tychoish: bump

model/perf.go
tychoish: exactly using bsonutil.MustHaveTag()

model/perf.go
```diff
@@ -637,3 +657,142 @@ func (r *PerformanceResults) FindOutdatedRollups(ctx context.Context, name strin
...
+		},
+	}
+	_, err := env.GetDB().Collection(perfResultCollection).UpdateMany(ctx, seriesFilter, pullUpdate)
+	return errors.Wrap(err, "Unable to clear change points")
+}
+
+func CreateChangePoint(ctx context.Context, env cedar.Environment, resultToUpdate string, measurement string, algorithm AlgorithmInfo) error {
+	filter := bson.M{"_id": resultToUpdate}
+	update := bson.M{
+		"$push": bson.M{
```
tychoish: does this make idempotency, upon re-run, more complicated?

rest/data/perf.go
tychoish: bump

rest/data/perf_test.go
tychoish: I don't think it provides benefit and I'd prefer to not mock things out except in some lower level cases. 

Please remove it.

units/recalculate_change_points_test.go
tychoish: I guess it's alright for now.

rest/data/perf.go
ancostas: I think this has been done.

tychoish: that doesn't seem to be reflected in the latest patch.

model/perf.go
ancostas: This seems much more complicated than what I need here.

rest/data/perf_test.go
ancostas: What can I mock at a lower level such that I don't need a queue locally and can detect that this job gets scheduled as I expect it to?

model/perf.go
ancostas: I think any change points that match from previous jobs get cleared out immediately prior to this. So re-running would be clear-->add one-->clear-->add one-->clear-->add one. Seems idempotent to me.

ancostas:
@tychoish @julianedwards bump...

model/perf.go
tychoish: per our in person conversation, we think it makes sense to combine these operations into a single operation,.

rest/data/perf_test.go
tychoish: I'd just use https://godoc.org/github.com/mongodb/amboy/queue#NewLocalLimitedSize

and just check the size of the queue before and after.

ancostas: I've done this, but I think it would be really nice to have some way of checking the TimeSeriesIds of the jobs that have been enqueued.

model/perf.go
ancostas: I've used a debugger, that's why these are separate. Much easier than adding print statements IMO.

ancostas:
@tychoish I'm using https://godoc.org/github.com/mongodb/amboy/queue#NewLocalLimitedSize now and getting `queue not open. could not add recalculate-change-points.rollup2.rollup2.rollup2.rollup2.OperationsTotal.2020-01-09 17:20:00 +0000 UTC`.

Is there something I need to do to open the queue?

tychoish:
why is "rollup2" in the job name so many times? 

you can try starting the queue and seeing what happens... NewOrderedLocalQueue may also be a good one to use, it definitely allows adding without starting the job but it's a weird implementation in every other respect.

ancostas:
I think the "rollup2" is in there since I put that as the project/variant/task/test/measurement in the test data pumped into the db. Going to update them to rollup2project/..etc.

model/perf.go
```diff
@@ -637,3 +655,174 @@ func (r *PerformanceResults) FindOutdatedRollups(ctx context.Context, name strin
...
+	if err != nil {
+		return nil, errors.Wrap(err, "Could not decode time series")
+	}
+	return &TimeSeries{
+		TimeSeriesId: timeSeriesId,
+		Data:         res,
+	}, nil
+}
+
+func ReplaceChangePoints(ctx context.Context, env cedar.Environment, timeSeries *TimeSeries, changePoints []ChangePoint) []error {
```
tychoish: please don't return a slice of errors, use `grip.NewBasicCatcher` and then the methods on that interface to collect errors.

model/perf.go
```diff
@@ -637,3 +655,174 @@ func (r *PerformanceResults) FindOutdatedRollups(ctx context.Context, name strin
...
+			"change_points": bson.M{
+				"measurement": timeSeriesId.Measurement,
+			},
+		},
+	}
+	_, err := env.GetDB().Collection(perfResultCollection).UpdateMany(ctx, seriesFilter, pullUpdate)
+	return errors.Wrap(err, "Unable to clear change points")
+}
+
+func CreateChangePoint(ctx context.Context, env cedar.Environment, resultToUpdate string, measurement string, algorithm AlgorithmInfo) error {
```
tychoish: didn't we talk about about merging this into the previous function?

rest/data/perf.go
```diff
@@ -233,6 +237,31 @@ func (dbc *DBConnector) FindPerformanceResultWithChildren(ctx context.Context, i
...
+	queue := dbc.env.GetRemoteQueue()
+	ids, err := model.GetTimeSeriesIds(ctx, dbc.env)
+	if err != nil {
+		return gimlet.ErrorResponse{StatusCode: http.StatusInternalServerError, Message: fmt.Sprint("Failed to get time series ids")}
+	}
+	for _, id := range ids {
+		job := units.NewRecalculateChangePointsJob(id)
+		err = queue.Put(ctx, job)
+		if err != nil {
+			return errors.Wrap(err, message.WrapError(err, message.Fields{
```
tychoish: I would errors.New() this. 

also I would consider doing a continue on error kind of thing here, to avoid cases where a collision early on blocks later jobs. 

we can talk about changing the logging/error reporting strategy.

model/perf.go
ancostas: Ok

model/perf.go
ancostas: Clear change points clears them all for a metric, create adds one to a specific point in its history. I added `ReplaceChangePoints` to do both in an idempotent fashion. I'm going to make these private so nobody calls them accidentally from outside of the package.

rest/data/perf.go
ancostas: Changed to errors.New(). Used `grip.NewBasicCatcher` here as well to continue on error.

PR #155, opened 2019-12-17
MAKE-1071, MAKE-1072: Handle new and revised data signal processing.

ancostas:
This is waiting on https://github.com/evergreen-ci/cedar/pull/153 before it really needs to be reviewed.


units/ftdc_rollups.go
```diff
@@ -149,5 +152,29 @@ func (j *ftdcRollupsJob) Run(ctx context.Context) {
 		if err != nil {
 			j.AddError(errors.Wrapf(err, "problem adding rollup %s for perf result %s", r.Name, j.PerfID))
 		}
+		j.createSignalProcessingJob(ctx, result, r)
```
ancostas: This would handle both cases (new results and revised old results), right?

PR #158, opened 2020-01-14
MAKE-1069: Cron backfiller for signal processing recalculation

ancostas:
So, this is meant to do the following:

* Provision the db with `metric_metadata`:
```
{
	"_id" : {
		"task" : "service_architecture_workloads",
		"variant" : "M60-atlas-repl",
		"project" : "sys-perf-atlas",
		"test" : "PrimaryRequiredReads-ActorFinished",
		"measurement" : "Latency95thPercentile"
	},
	"lastChangePointDetection" : null,
	"lastTimeSeriesUpdate" : ISODate("2020-01-14T18:09:16.481Z")
}
```
* Periodically update the above metadata with novel metrics that are added down the line
* Automatically update `lastTimeSeriesUpdate` for the metric whenever rollups are added or revised
* Automatically update `lastChangePointDetection` whenever the metric has change points detected and added
* Periodically check `metric_metadata` for any metrics where `lastChangePointDetection` < `lastTimeSeriesUpdate`, and schedule a change point detection job for them.

I've set this thing to run hourly.


model/perf.go
```diff
@@ -21,6 +21,7 @@ import (
 )
 
 const perfResultCollection = "perf_results"
+const metricMetadataCollection = "metric_metadata"
```
tychoish: can this be `perf_change_point_metadata`

model/perf.go
```diff
@@ -656,13 +657,90 @@ func (r *PerformanceResults) FindOutdatedRollups(ctx context.Context, name strin
...
+		LastTimeSeriesUpdate     time.Time    `bson:"lastTimeSeriesUpdate"`
+	}
+	err = cur.All(ctx, &res)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Unable to decode metrics needing change results")
+	}
+	result := []TimeSeriesId{}
+	for _, val := range res {
+		result = append(result, val.Id)
+	}
```
tychoish: I'd just do an aggregation/projection to reshape the documents on the db rather than copying them out.

model/perf.go
```diff
@@ -656,13 +657,90 @@ func (r *PerformanceResults) FindOutdatedRollups(ctx context.Context, name strin
...
+func MarkMetricAsUpdated(ctx context.Context, env cedar.Environment, id TimeSeriesId) error {
+	err := createTimeStamp(ctx, env, id, "lastTimeSeriesUpdate")
+	if err != nil {
+		return errors.Wrapf(err, "Unable to mark metric as updated")
+	}
+	return nil
+}
+
+func MarkMetricAsAnalyzed(ctx context.Context, env cedar.Environment, id TimeSeriesId) error {
+	err := createTimeStamp(ctx, env, id, "lastChangePointDetection")
```
tychoish: please use bsonutil.MustHaveTags() values for these field names

model/perf.go
```diff
@@ -656,13 +657,90 @@ func (r *PerformanceResults) FindOutdatedRollups(ctx context.Context, name strin
...
+				"$lastChangePointDetection",
+				"$lastTimeSeriesUpdate",
+			},
+		},
+	})
+	if err != nil {
+		return nil, errors.Wrapf(err, "Unable to get metrics needing change point detection")
+	}
+	defer cur.Close(ctx)
+	var res []struct {
```
tychoish: can you declare a type in the package that has these values, with the MustHaveTag values?

model/perf.go
```diff
@@ -684,12 +762,20 @@ func GetTimeSeriesIds(ctx context.Context, env cedar.Environment) ([]TimeSeriesI
...
+		bson.M{
 			"$replaceRoot": bson.M{
 				"newRoot": "$_id",
 			},
-		},
-	}
+		})
+}
+
+func GetTimeSeriesIds(ctx context.Context, env cedar.Environment) ([]TimeSeriesId, error) {
```
tychoish: can we use ID rather than Id at least on any new code?

model/perf.go
```diff
@@ -684,12 +762,20 @@ func GetTimeSeriesIds(ctx context.Context, env cedar.Environment) ([]TimeSeriesI
...
+	pipeline := timeSeriesIdPipeline()
```
tychoish: I'd just call the function inline on the next line rather than declare the variables.

units/add_new_metric_metadata.go
```diff
@@ -0,0 +1,73 @@
...
+
+func init() {
+	registry.AddJobType("periodic-change-point-detection", func() amboy.Job { return makePeriodicChangePointJob() })
+}
+
+func makePeriodicChangePointJob() *PeriodicChangePointJob {
+	j := &PeriodicChangePointJob{
+		Base: &job.Base{
+			JobType: amboy.JobType{
+				Name:    "periodic-change-point-detection",
```
tychoish: the job name should be a constant, both here and on line 25

units/add_new_metric_metadata.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/evergreen-ci/cedar/model"
+	"github.com/evergreen-ci/cedar/util"
+	"github.com/mongodb/amboy"
+	"github.com/mongodb/amboy/dependency"
+	"github.com/mongodb/amboy/job"
+	"github.com/mongodb/amboy/registry"
+	"github.com/pkg/errors"
+)
+
+type PeriodicChangePointJob struct {
```
tychoish: I would not export this type.

units/add_new_metric_metadata.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	j.SetDependency(dependency.NewAlways())
+	return j
+}
+
+func NewPeriodicChangePointJob() amboy.Job {
+	j := makePeriodicChangePointJob()
+	// Every ten minutes at most
+	timestamp := util.RoundPartOfHour(10)
+	j.SetID(fmt.Sprintf("%s.%s", j.JobType.Name, timestamp))
```
tychoish: I would prefer if you passed an ID string into this function, rather than constructing the timestamp rounding in the constructor here.

units/ftdc_rollups.go
```diff
@@ -149,5 +149,17 @@ func (j *ftdcRollupsJob) Run(ctx context.Context) {
 		if err != nil {
 			j.AddError(errors.Wrapf(err, "problem adding rollup %s for perf result %s", r.Name, j.PerfID))
 		}
+		j.markTimeSeriesAsUpdated(ctx, result, r)
 	}
 }
+
+func (j *ftdcRollupsJob) markTimeSeriesAsUpdated(ctx context.Context, result *model.PerformanceResult, rollup model.PerfRollupValue) {
+	id := model.TimeSeriesId{
```
tychoish: I've seen constructing an ID in a few functions now, is it possible to create a constrcutor for TimeSeriesID that takes the other model types?

units/crons.go
```diff
@@ -79,6 +79,10 @@ func StartCrons(ctx context.Context, env cedar.Environment, rpcTLS bool) error {
 
 		return queue.Put(ctx, job)
 	})
+	amboy.IntervalQueueOperation(ctx, remote, time.Hour, time.Now(), opts, func(ctx context.Context, queue amboy.Queue) error {
+		job := NewPeriodicChangePointJob()
+		return queue.Put(ctx, job)
+	})
```
tychoish: I would avoid creating seperate interval queue operations when the same queue and interval is used, so I'd put this in the above operation. 

I'd also pass the timestamp in using the `tsFormat` as in line 72

units/crons.go
```diff
@@ -79,6 +79,10 @@ func StartCrons(ctx context.Context, env cedar.Environment, rpcTLS bool) error {
 
 		return queue.Put(ctx, job)
 	})
+	amboy.IntervalQueueOperation(ctx, remote, time.Hour, time.Now(), opts, func(ctx context.Context, queue amboy.Queue) error {
```
tychoish: did we add a service flag to be able to disable the job, as we see on lines 63-70? Should we check that here?

units/add_new_metric_metadata.go
```diff
@@ -0,0 +1,73 @@
...
+	if j.queue == nil {
+		j.queue = j.env.GetRemoteQueue()
+	}
+	defer j.MarkComplete()
+	if j.conf == nil {
+		j.conf = model.NewCedarConfig(j.env)
+	}
+	if j.env == nil {
+		j.env = cedar.GetEnvironment()
+	}
```
tychoish: the order of operations here is somewhat suspect: you nil check the env after you've (potentially) used it a few times.

units/add_new_metric_metadata.go
```diff
@@ -0,0 +1,73 @@
...
+	err := model.AddNewMetricMetadata(ctx, j.env)
+	if err != nil {
+		j.AddError(errors.Wrapf(err, "Job to add new metric metadata has failed"))
+		return
+	}
+	needUpdates, err := model.GetMetricsNeedingChangePointDetection(ctx, j.env)
+	if err != nil {
+		j.AddError(errors.Wrapf(err, "Unable to get metrics needing change point detection"))
```
tychoish: these two errors.Wrapf calls should either be errors.Wrap calls or include some kind of ID

units/add_new_metric_metadata.go
```diff
@@ -0,0 +1,73 @@
...
+
+func NewPeriodicChangePointJob() amboy.Job {
+	j := makePeriodicChangePointJob()
+	// Every ten minutes at most
+	timestamp := util.RoundPartOfHour(10)
+	j.SetID(fmt.Sprintf("%s.%s", j.JobType.Name, timestamp))
+	return j
+}
+
+func (j *PeriodicChangePointJob) Run(ctx context.Context) {
```
tychoish: is there any logging that makes sense here?

units/add_new_metric_metadata.go
```diff
@@ -0,0 +1,73 @@
...
+
+func makePeriodicChangePointJob() *PeriodicChangePointJob {
+	j := &PeriodicChangePointJob{
+		Base: &job.Base{
+			JobType: amboy.JobType{
+				Name:    "periodic-change-point-detection",
+				Version: 1,
+			},
+		},
+		env: cedar.GetEnvironment(),
```
tychoish: I think if the nil check of the environment at job-runtime populates the env, then it's safe to skip this.

model/perf_data.go
```diff
@@ -65,8 +65,9 @@ var (
 // PerfRollups describes the "rolled up", or calculated metrics from time
 // series data collected in a given performance test, of a performance result.
 type PerfRollups struct {
-	Stats       []PerfRollupValue `bson:"stats"`
-	ProcessedAt time.Time         `bson:"processed_at"`
+	Stats                  []PerfRollupValue `bson:"stats"`
+	ProcessedAt            time.Time         `bson:"processed_at"`
+	ChangePointsDetectedAt time.Time         `bson:"change_points_detected_at"`
```
tychoish:
```
// PerformanceResult describes a single result of a performance test from
// Evergreen.
type PerformanceResult struct {
	ID          string                `bson:"_id,omitempty"`
	Info        PerformanceResultInfo `bson:"info,omitempty"`
	CreatedAt   time.Time             `bson:"created_at"`
	CompletedAt time.Time             `bson:"completed_at"`
	Version     int                   `bson:"version,omitempty"`

	// The source timeseries data is stored in a remote location,
	// we'll probably need to store an identifier so we know which
	// service to use to access that data. We'd then summarize
	// that data and store it in the document.
	//
	// The structure has a both a schema to describe the layout
	// the data (e.g. raw, results,) format (e.g. bson/ftdc/json),
	// and tags to describe the source (e.g. user submitted,
	// generated.)
	Artifacts            []ArtifactInfo `bson:"artifacts"`
	FailedRollupAttempts int            `bson:"failed_rollup_attempts"`

	Rollups  PerfRollups  `bson:"rollups"`
	Analysis PerfAnalysis `bson:"analysis"`

	env       cedar.Environment
	populated bool
}

type PerfAnalysis struct {
	ChangePoints []ChangePoint `bson:"change_points"`
	ProcessedAt  time.Time     `bson:"processed_at"`
}
```

units/add_new_metric_metadata.go
ancostas: ok.

units/add_new_metric_metadata.go
ancostas: What do you mean?

units/add_new_metric_metadata.go
ancostas: Added a log statement so we can see when it's running.

units/crons.go
ancostas: I think I've done what you mean? Had a bit of trouble understand though.

model/perf.go
ancostas: Think this has been handled.

model/perf.go
ancostas: Think that's refactored out now.

units/ftdc_rollups.go
ancostas: Refactored out I think.

model/perf.go
```diff
@@ -80,8 +80,11 @@ func CreatePerformanceResult(info PerformanceResultInfo, source []ArtifactInfo,
 			Stats:       append([]PerfRollupValue{}, rollups...),
 			ProcessedAt: createdAt,
 		},
-		populated:    true,
-		ChangePoints: []ChangePoint{},
+		populated: true,
+		Analysis: PerfAnalysis{
+			ChangePoints: []ChangePoint{},
+			ProcessedAt:  time.Time{},
```
tychoish: I believe that the zero/default initlizations of these values are fine, and that you don't need to specify any of these values at all.

ancostas:
evergreen merge

ancostas:
evergreen merge

PR #159, opened 2020-01-14
MAKE-1050: Add error handling to signal processing detector client

ancostas:
It looks like the retrying logic should pretty much handle everything. I've set max retries to 10 since I don't think things will get better after 10 failures and we should examine that case more closely, and removed 500 from the recoverable errors since this corresponds to an unhandled exception from our API.


ancostas:
evergreen merge

PR #160, opened 2020-01-16
MAKE-1105: Add proc_name to buildlogger task_id API endpoint

julianedwards:


rest/data/interface.go
```diff
@@ -51,11 +51,11 @@ type Connector interface {
...
-	// FindLogsByTaskID returns the buildlogger logs with the given task id
-	// and optional tags merged via a LogIterator with the corresponding
-	// time range. If the bool is set to true, the log lines should be
-	// concatenated with their timestamps.
-	FindLogsByTaskID(context.Context, string, util.TimeRange, int, bool, ...string) (io.Reader, error)
+	// FindLogsByTaskID returns the buildlogger logs with the given task
+	// id process name (can be left empty), and optional tags merged via a
+	// LogIterator with the corresponding time range. If the bool is set to
+	// true, the log lines should be concatenated with their timestamps.
+	FindLogsByTaskID(context.Context, string, util.TimeRange, int, bool, string, ...string) (io.Reader, error)
```
tychoish: Does it make sense to pass this in as an arg struct

julianedwards: yeah, especially now that I also need to pass in execution number. I am going to change it

PR #161, opened 2020-01-21
MAKE-1111: Trigger signal processing job in cedar when YCSB added manually

ancostas:


rpc/internal/perf_service.go
```diff
@@ -106,6 +106,19 @@ func (srv *perfService) AttachRollups(ctx context.Context, rollupData *RollupDat
...
 
+	id := model.PerformanceResultSeriesId{
+		Project: record.Info.Project,
+		Variant: record.Info.Variant,
+		Task:    record.Info.TaskName,
+		Test:    record.Info.TestName,
+	}
+
+	processingJob := units.NewRecalculateChangePointsJob(id)
+	err := srv.env.GetRemoteQueue().Put(ctx, processingJob)
```
tychoish: combine 116 and 117

rpc/internal/perf_service.go
```diff
@@ -106,6 +106,19 @@ func (srv *perfService) AttachRollups(ctx context.Context, rollupData *RollupDat
 		return nil, newRPCError(codes.InvalidArgument, errors.Wrapf(err, "problem attaching rollup data for perf result '%s'", record.ID))
 	}
 
+	id := model.PerformanceResultSeriesId{
```
tychoish: two things: 

1. A nit. I'd prefer to have the d in ID capitalized. 
2. is there a way to have a method to produce this kind of object from the record/info type? I've seen this pattern used a lot, and it seems easy to get it wrong....

ancostas: Capitalized ID and added a `ToPerformanceResultSeriesId` to `PerformanceResultInfo`.

rpc/internal/perf_service.go
ancostas: Go fmt file is pushing them back apart for me (line length?). Not for you?

PR #169, opened 2020-01-28
MAKE-1132: Handle case of no measurements

ancostas:


model/change_point_detection.go
```diff
@@ -257,6 +257,12 @@ func GetPerformanceData(ctx context.Context, env cedar.Environment, performanceR
 	if err != nil {
 		return nil, errors.Wrap(err, "Could not decode time series")
 	}
+	if len(res) < 1 {
+		return &PerformanceData{
+			PerformanceResultId: performanceResultId,
+			Data:                []MeasurementData{},
+		}, nil
+	}
```
ancostas: This handles the case of empty rollups.stats gracefully.

PR #172, opened 2020-01-30
MAKE-1137: Retry 499 status codes in change detector

ancostas:


tychoish:
sorry you actually want to change this in `perf/chainge_point_detector.go` because you don't use the defaults in your client.

PR #173, opened 2020-01-31
MAKE-1115: Add priority to buildlogger logs

julianedwards:


model/buildlogger.go
```diff
@@ -155,7 +156,14 @@ func (l *Log) Append(ctx context.Context, lines []LogLine) error {
 
 	linesCombined := ""
 	for _, line := range lines {
-		linesCombined += prependTimestamp(line.Timestamp, line.Data)
+		// unlikely scenario, but just in case priority is out of range.
+		if line.Priority > level.Emergency {
+			line.Priority = level.Emergency
+		} else if line.Priority < level.Invalid {
+			line.Priority = level.Invalid
```
tychoish: shouldn't we make this trace?

model/log_iterator.go
```diff
@@ -472,22 +473,27 @@ func (i *mergingIterator) Close() error {
...
 	}, nil
 }
 
-func prependTimestamp(t time.Time, data string) string {
+func prependPriorityAndTimestamp(p level.Priority, t time.Time, data string) string {
+	priority := fmt.Sprintf("%3d", p)
 	ts := fmt.Sprintf("%20d", util.UnixMilli(t))
 
-	return fmt.Sprintf("%s%s\n", ts, data)
+	return fmt.Sprintf("%s%s%s\n", priority, ts, data)
```
tychoish: can we combine these fmt.Sprintf calls into one fmt.Sprintf?

PR #174, opened 2020-01-31
MAKE-1140: Reconfigure change point detector retry policy to be lenient

ancostas:


perf/change_point_detector.go
```diff
@@ -96,9 +96,9 @@ func (spc *signalProcessingClient) doRequest(method, route string, ctx context.C
 	}
 
 	conf := util.HTTPRetryConfiguration{
-		MaxRetries:      10,
+		MaxRetries:      50,
 		TemporaryErrors: true,
-		MaxDelay:        5 * time.Second,
+		MaxDelay:        5 * time.Minute,
```
tychoish: compromise: 30 seconds

PR #180, opened 2020-02-25
MAKE-1172: add POST routes for login

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-1172

* Add POST routes for login. The GET routes will be removed after people update their CLIs.
* Dont close request bodies. The docs say that the underlying http.Server is responsible for closing them so the `http.Handler` doesn't have to.


Kimchelly:
The curator test is failing only because it's posting to the actual cedar server.

PR #182, opened 2020-02-27
MAKE-1168: Use API headers with evg proxy auth

julianedwards:


Kimchelly:
Logic-wise this seems fine, but should we try to split this function into smaller ones to make it more testable and to pare down its responsibilities?

PR #186, opened 2020-03-10
MAKE-1151: Initialize float series as empty array to avoid null JSON …

zamj:
…value

When marshaled, nil slices get encoded as `null` while empty arrays get encoded as `[]`.

https://golang.org/pkg/encoding/json/#Marshal

https://play.golang.org/p/jvZ7Sx9hoz2

### DIFF
--- a/units/recalculate_change_points.go
+++ b/units/recalculate_change_points.go
@@ -99,7 +99,7 @@ func (j *recalculateChangePointsJob) Run(ctx context.Context) {
 		sort.Slice(series.TimeSeries, func(i, j int) bool {
 			return series.TimeSeries[i].Order < series.TimeSeries[j].Order
 		})
-		var float_series []float64
+		float_series := make([]float64, 0)
 		for _, item := range series.TimeSeries {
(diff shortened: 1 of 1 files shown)


units/recalculate_change_points.go
```diff
@@ -99,7 +99,7 @@ func (j *recalculateChangePointsJob) Run(ctx context.Context) {
 		sort.Slice(series.TimeSeries, func(i, j int) bool {
 			return series.TimeSeries[i].Order < series.TimeSeries[j].Order
 		})
-		var float_series []float64
+		float_series := make([]float64, 0)
```
julianedwards: do `float_series := []float64{}`

tychoish: why not make([]float64, len(series.TimeSeries), and then avoid append and just set indexes....

julianedwards: oh yeah i missed the for loop on line 103, this makes more sense

zamj: Sure, done

PR #189, opened 2020-03-17
MAKE-789: string join to avoid concat

tychoish:


model/buildlogger.go
```diff
@@ -154,16 +154,17 @@ func (l *Log) Append(ctx context.Context, lines []LogLine) error {
 	}
 	key := fmt.Sprint(util.UnixMilli(time.Now()))
 
+	lines := make([]string, len(lines))
 	linesCombined := ""
```
julianedwards: remove this

model/buildlogger.go
```diff
@@ -154,16 +154,17 @@ func (l *Log) Append(ctx context.Context, lines []LogLine) error {
 	}
 	key := fmt.Sprint(util.UnixMilli(time.Now()))
 
+	lines := make([]string, len(lines))
```
julianedwards: need to give this a different name

PR #193, opened 2020-03-26
MAKE-1202: Better REST logging

julianedwards:


rest/buildlogger_routes.go
```diff
@@ -89,6 +90,13 @@ func (h *logGetByIDHandler) Run(ctx context.Context) gimlet.Responder {
 	}
 	data, next, paginated, err := h.sc.FindLogByID(ctx, opts)
 	if err != nil {
+		grip.Error(message.Fields{
```
tychoish: I'd do message.WrapErr

rest/buildlogger_routes.go
```diff
@@ -212,6 +227,13 @@ func (h *logGetByTaskIDHandler) Run(ctx context.Context) gimlet.Responder {
 	}
 	data, next, paginated, err := h.sc.FindLogsByTaskID(ctx, opts)
 	if err != nil {
+		grip.Error(message.Fields{
+			"method":  "GET",
+			"route":   "/buildlogger/task_id/{task_id}",
```
tychoish: shouldn't we put the task id here?

util/http.go
```diff
@@ -146,3 +149,13 @@ func GetHTTPRetryableClient(conf HTTPRetryConfiguration) *http.Client {
 func GetDefaultHTTPRetryableClient() *http.Client {
 	return GetHTTPRetryableClient(NewDefaultHTTPRetryConf())
 }
+
+func LogRequestError(r *http.Request, err error) {
```
tychoish: I was thinking that we should keep this in the rest package.

util/http.go
```diff
@@ -146,3 +149,13 @@ func GetHTTPRetryableClient(conf HTTPRetryConfiguration) *http.Client {
...
+	grip.Error(message.Fields{
```
tychoish: errors wrap this so it noops if there's no error?

PR #194, opened 2020-03-30
MAKE-1193: fix string concat by using bytes.Buffer for log lines append

julianedwards:


model/buildlogger.go
```diff
@@ -154,16 +154,19 @@ func (l *Log) Append(ctx context.Context, lines []LogLine) error {
 	}
 	key := fmt.Sprint(util.UnixMilli(time.Now()))
 
-	linesCombined := make([]string, len(lines))
-	for idx, line := range lines {
+	lineBuffer := bytes.NewBuffer([]byte{})
```
tychoish: &bytes.Buffer{} rather than the NewBuffer call

model/buildlogger.go
```diff
@@ -154,16 +154,19 @@ func (l *Log) Append(ctx context.Context, lines []LogLine) error {
...
+	for _, line := range lines {
 		// unlikely scenario, but just in case priority is out of range.
 		if line.Priority > level.Emergency {
 			line.Priority = level.Emergency
 		} else if line.Priority < level.Invalid {
 			line.Priority = level.Trace
 		}
 
-		linesCombined[idx] = prependPriorityAndTimestamp(line.Priority, line.Timestamp, line.Data)
+		_, err := lineBuffer.WriteString(prependPriorityAndTimestamp(line.Priority, line.Timestamp, line.Data))
```
tychoish: should we ensure that we don't miss a newline?

julianedwards: what do you mean?

PR #198, opened 2020-04-07
MAKE-1211: Add triage section to change points in cedar

ancostas:


perf/change_point_detector.go
```diff
@@ -66,6 +66,12 @@ func (spc *signalProcessingClient) DetectChanges(ctx context.Context, series []f
 				Name:    point.Algorithm.Name,
 				Version: point.Algorithm.Version,
 			},
+			CalculatedOn: time.Now(),
+			Measurement: measurement,
+			Triage: model.TriageInfo{
+				TriagedOn: time.Time{},
+				Status:    model.TriageStatusUntriaged,
+			},
```
ancostas, line 74: These concerns seemed to fit here much better.

model/change_point_detection.go
```diff
@@ -57,6 +59,34 @@ var (
...
+
+var (
+	perfTriageInfoTriagedOnKey = bsonutil.MustHaveTag(TriageInfo{}, "TriagedOn")
+	perfTriageInfoStatusKey    = bsonutil.MustHaveTag(TriageInfo{}, "Status")
+)
+
+type TriageStatus string
+
+const (
+	TriageStatusUntriaged          TriageStatus = "untriaged"
```
tychoish, line 75: I think it makes sense to use the empty string "" as the untriaged value.

perf/change_point_detector.go
```diff
@@ -66,6 +66,12 @@ func (spc *signalProcessingClient) DetectChanges(ctx context.Context, series []f
 				Name:    point.Algorithm.Name,
 				Version: point.Algorithm.Version,
 			},
+			CalculatedOn: time.Now(),
+			Measurement: measurement,
+			Triage: model.TriageInfo{
+				TriagedOn: time.Time{},
```
tychoish, line 72: this isn't needed, I would rely on default initialization.

perf/change_point_detector.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 type ChangeDetector interface {
-	DetectChanges(context.Context, []float64) ([]model.ChangePoint, error)
+	DetectChanges(context.Context, []float64, string) ([]model.ChangePoint, error)
```
tychoish, line 21: can't there be documentation here. I wouldn't change the interface just to initialize the data.

model/change_point_detection.go
```diff
@@ -298,15 +328,11 @@ func clearChangePoints(ctx context.Context, env cedar.Environment, performanceRe
 	return errors.Wrap(err, "Unable to clear change points")
 }
 
-func createChangePoint(ctx context.Context, env cedar.Environment, resultToUpdate string, measurement string, algorithm AlgorithmInfo) error {
+func createChangePoint(ctx context.Context, env cedar.Environment, resultToUpdate string, cp ChangePoint) error {
```
tychoish, line 331: I'd leave this the way it was.

model/change_point_detection.go
ancostas: Design doc dictates "untriaged", and I'd really prefer this explicit label vs an empty one.

model/change_point_detection.go
ancostas: We need to persist the triage section as well now; rather than keeping adding arguments to this, it seems simpler to just persist the change point itself IMO.

perf/change_point_detector.go
ancostas: I can document it; seems simpler to specify the measurement you are detecting a change for rather than fixing up the result elsewhere in the codebase (which can be forgotten).

perf/change_point_detector.go
ancostas: Default initialization would wipe the "untriaged" status, no?

model/change_point_detection.go
tychoish: right, but doing this forces us to leak our data layout into the implementation details of the change detector, which feels more complicated to maintain long term.

perf/change_point_detector.go
tychoish: right but, this forces us to leak the data limitation into the implementation of this interface. 

In retrospect, it's probably a problem that we return a type from the model package at all.

model/change_point_detection.go
ancostas: Cleaned up the use of model package in the change detector as discussed.

ancostas:
@tychoish Failing test is just some stuff in wrong order (groan). Will fix that up, but feel free to review.

model/change_point_detection.go
```diff
@@ -30,6 +30,32 @@ type ChangePoint struct {
...
+	cp := ChangePoint{
+		Index: index,
+		Algorithm: AlgorithmInfo{
+			Name:    algorithmName,
+			Version: algorithmVersion,
+		},
+		CalculatedOn: time.Now(),
+		Measurement:  measurement,
+		Triage: TriageInfo{
+			TriagedOn: time.Time{},
```
tychoish, line 43: Omit this line.

model/change_point_detection.go
```diff
@@ -30,6 +30,32 @@ type ChangePoint struct {
 	Triage       TriageInfo    `bson:"triage" json:"triage" yaml:"triage"`
 }
 
+func CreateChangePoint(index int, measurement string, algorithmName string, algorithmVersion int, algorithmConfiguration map[string]interface{}) ChangePoint {
```
tychoish, line 33: I'd pass AlgorithmInfo in as an argument.

perf/change_point_detector.go
```diff
@@ -9,80 +9,100 @@ import (
...
 	"github.com/pkg/errors"
 )
 
+type Algorithm interface {
+	Name() string
+	Version() int
+	Configuration() map[string]interface{}
+}
+
+type EDivisiveMeans struct {
```
tychoish, line 25: I'd make this private?

perf/change_point_detector.go
```diff
@@ -9,80 +9,100 @@ import (
...
 	"github.com/evergreen-ci/gimlet"
 	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/message"
 	"github.com/pkg/errors"
 )
 
+type Algorithm interface {
+	Name() string
+	Version() int
+	Configuration() map[string]interface{}
```
tychoish, line 22: would it be reasonable to have this not be a map?

units/recalculate_change_points_test.go
```diff
@@ -6,6 +6,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/evergreen-ci/cedar/perf"
+
```
tychoish, line 10: omit blank line.

units/recalculate_change_points.go
```diff
@@ -104,7 +104,14 @@ func (j *recalculateChangePointsJob) Run(ctx context.Context) {
 			floatSeries[i] = item.Value
 		}
 
-		changePoints, err := j.changePointDetector.DetectChanges(ctx, floatSeries)
+		result, err := j.changePointDetector.DetectChanges(ctx, floatSeries)
+
+		var changePoints []model.ChangePoint
+		for _, pointIndex := range result {
+			mapped := model.CreateChangePoint(pointIndex, series.Measurement, j.changePointDetector.Algorithm().Name(), j.changePointDetector.Algorithm().Version(), AlgorithmConfigurationToOptions(j.changePointDetector.Algorithm().Configuration()))
```
ancostas, line 111: Looks like `units` is the glue package between `model` and `perf`!

units/recalculate_change_points.go
```diff
@@ -122,3 +122,14 @@ func (j *recalculateChangePointsJob) Run(ctx context.Context) {
 	j.AddError(model.ReplaceChangePoints(ctx, j.env, performanceData, mappedChangePoints))
 	j.AddError(model.MarkPerformanceResultsAsAnalyzed(ctx, j.env, performanceData.PerformanceResultId))
 }
+
+func AlgorithmConfigurationToOptions(configurationValues []perf.AlgorithmConfigurationValue) []model.AlgorithmOption {
```
tychoish, line 126: nit: make this private

PR #200, opened 2020-04-13
MAKE-1216: Add change point list grouped by version endpoint to cedar

zamj:


model/change_point_detection.go
```diff
@@ -22,7 +23,7 @@ var (
 )
 
 type ChangePoint struct {
-	Index        int
+	Index        int           `bson:"index" json:"index"`
```
tychoish, line 26: add yaml just for consistency.

model/change_point_detection.go
```diff
@@ -104,10 +105,11 @@ func (ts TriageStatus) Validate() error {
...
 type PerformanceResultSeriesID struct {
-	Project string `bson:"project"`
-	Variant string `bson:"variant"`
-	Task    string `bson:"task"`
-	Test    string `bson:"test"`
+	Project     string `bson:"project" json:"project"`
+	Variant     string `bson:"variant" json:"variant"`
+	Task        string `bson:"task" json:"task"`
+	Test        string `bson:"test" json:"test"`
+	ThreadLevel int32  `bson:"thread_level" json:"thread_level"`
```
tychoish, line 112: this seems like we're collapsing information across some kind of boundary

also thread_level isn't a top level concept in cedar's data model, and I think it might make sense to rethink this a bit.

model/change_point_detection.go
```diff
@@ -126,6 +128,17 @@ type PerformanceData struct {
...
 }
 
+type ChangePointWithPerformanceData struct {
+	PerformanceResultSeriesID `bson:",inline"`
+	ChangePoint               ChangePoint `bson:"change_point"`
+	PerfResultId              string      `bson:"perf_result_id" json:"perf_result_id"`
+}
+
+type GetChangePointsGroupedByVersionResult struct {
+	VersionId    string                           `bson:"version_id" json:"version_id"`
```
tychoish, line 138: VersionID

model/change_point_detection.go
```diff
@@ -126,6 +128,17 @@ type PerformanceData struct {
 	Data                []MeasurementData         `bson:"data"`
 }
 
+type ChangePointWithPerformanceData struct {
+	PerformanceResultSeriesID `bson:",inline"`
```
tychoish, line 132: Can we avoid inlining? I think that locks us into the legacy bson library in a way that I'm not comfortable with.

model/change_point_detection.go
```diff
@@ -126,6 +128,17 @@ type PerformanceData struct {
...
+	ChangePoint               ChangePoint `bson:"change_point"`
+	PerfResultId              string      `bson:"perf_result_id" json:"perf_result_id"`
```
tychoish, line 134: ID

model/change_point_detection.go
```diff
@@ -328,6 +341,94 @@ func ReplaceChangePoints(ctx context.Context, env cedar.Environment, performance
 	return catcher.Resolve()
 }
 
+func GetTotalPagesForChangePointsGroupedByVersion(ctx context.Context, env cedar.Environment, projectId string, pageSize int) (int, error) {
+	type CountResult struct {
```
tychoish, line 345: use an anon struct

model/change_point_detection.go
```diff
@@ -328,6 +341,94 @@ func ReplaceChangePoints(ctx context.Context, env cedar.Environment, performance
...
+		},
+		{
+			"$group": bson.M{
+				"_id":           "$version",
+				"change_points": bson.M{"$push": "$$ROOT"},
+				"version_id":    bson.M{"$first": "$version"},
+			},
+		},
+	}
+	pipe = append(getMatchAndUnwindForChangePointsGroupedByVersion(projectId), pipe...)
```
tychoish, line 418: can you have the builder functions be variadic bson.M functions that return []bson.M so the ordering is a bit more clear?

model/change_point_detection.go
ancostas: Why ID? This is for the performance result, not the change point itself.

model/change_point_detection.go
ancostas: @tychoish Q: do perf results have different thread levels for the measurements being made under them?

In any case, thread level should be applied to the individual measurement time series/change points.

model/change_point_detection.go
```diff
@@ -312,7 +325,7 @@ func GetPerformanceData(ctx context.Context, env cedar.Environment, performanceR
 func ReplaceChangePoints(ctx context.Context, env cedar.Environment, performanceData *PerformanceData, mappedChangePoints map[string][]ChangePoint) error {
 	err := clearChangePoints(ctx, env, performanceData.PerformanceResultId)
 	if err != nil {
-		return errors.Wrapf(err, "Unable to clear change points for measurement %s", performanceData.PerformanceResultId)
+		return errors.Wrapf(err, "Unable to clear change points for measurement %s", performanceData.PerformanceResultId.Project)
```
ancostas, line 328: Why?

zamj: I was getting complaints from the linter. Was a warning rather than an error. I can change them back if this was intended

model/change_point_detection.go
```diff
@@ -328,6 +341,94 @@ func ReplaceChangePoints(ctx context.Context, env cedar.Environment, performance
...
+	}
+	return 0, errors.New("Not able to get count of total changepoints matching query")
+}
+
+func getMatchAndUnwindForChangePointsGroupedByVersion(projectId string) []bson.M {
+	return []bson.M{
+		{
+			"$match": bson.M{
+				bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoProjectKey):             projectId,
+				bsonutil.GetDottedKeyName(perfAnalysisKey, perfAnalysisChangePointsKey, "0"): bson.M{"$exists": true},
```
ancostas, line 376: I guess this is more efficient than `{$ne:[]}`?

model/change_point_detection.go
zamj: I changed it back to how it was

model/change_point_detection.go
```diff
@@ -328,6 +341,94 @@ func ReplaceChangePoints(ctx context.Context, env cedar.Environment, performance
...
+			},
+		},
+		{
+			"$unwind": "$" + bsonutil.GetDottedKeyName(perfAnalysisKey, perfAnalysisChangePointsKey),
+		},
```
ancostas, line 377: So we are unwinding the change points, and then counting them, right? Shouldn't we be grouping them by version and counting the versions?

model/change_point_detection.go
zamj: Looks like it's about the same - when I do an explain plan, the one in the code is less than 100 ms faster

model/change_point_detection.go
```diff
@@ -328,6 +341,94 @@ func ReplaceChangePoints(ctx context.Context, env cedar.Environment, performance
...
+	}
+}
+
+func GetChangePointsGroupedByVersion(ctx context.Context, env cedar.Environment, projectId string, page, pageSize int) ([]GetChangePointsGroupedByVersionResult, error) {
+	pipe := []bson.M{
+		{
+			"$sort": bson.M{
+				bsonutil.GetDottedKeyName(perfAnalysisKey, perfAnalysisChangePointsKey, perfChangePointCalculatedOnKey): -1,
```
ancostas, line 386: I think we want crono sort on the version by default (i.e. by order).

model/change_point_detection.go
zamj: Misunderstood - thought we were paging by changepoint rather than by version. I'll change it

zamj: Changed

model/change_point_detection.go
tychoish: What do you mean "applied" we don't treat thread level different from any other parameter passed to the test. It is captured but not specially.

model/change_point_detection.go
tychoish: sorry PerfResultID rather just doing s/Id/ID/

rest/service.go
```diff
@@ -233,6 +233,7 @@ func (s *Service) addRoutes() {
 	s.app.AddRoute("/perf/version/{version}").Version(1).Get().RouteHandler(makeGetPerfByVersion(s.sc))
 	s.app.AddRoute("/perf/children/{id}").Version(1).Get().RouteHandler(makeGetPerfChildren(s.sc))
 	s.app.AddRoute("/perf/change_points").Version(1).Post().Wrap(checkUser).RouteHandler(makePerfSignalProcessingRecalculate(s.sc))
+	s.app.AddRoute("/perf/project/{projectId}/version/change_points").Version(1).Get().RouteHandler(makeGetChangePointsByVersion(s.sc))
```
tychoish, line 236: version seems weird in this path, but I'm not sure what I'd put?

zamj: Yeah, I'm not sure of a better way to describe it. It's currently listed in the design as this so that's what I went with

tychoish: can we change it to /perf/project/{projectID}/change_points_by_version

zamj: @ancostas any thoughts?

ancostas: For a single version we'd be doing `/perf/project/{p_id}/version/{v_id}/change_points`, right?

Wouldn't the bulk route then ideally be `/perf/project/{projectId}/version/change_points` for multiple versions w/ change points under them?

ancostas: Separately, FWIW, I think it would be nice if we could just go with the design doc's prescription. There exists a case where the UI consuming this route had already been developed, and we'd basically be breaking it and requiring an update in a separate codebase. That's not happenening in this case, but something I think is worth keeping in mind -- this is why I try to include exact prescriptions for the APIs being developed in my designs, without saying much about how exactly to implement them.

tychoish: we haven't used that paradigm for aggregate routes anywhere in cedar or evergreen. 

what about /perf/project/{projectID}/version/{versionID}/change_points and /perf/project/{projectID}/change_points

model/change_point_detection.go
ancostas: ms or ns? If ms that seems pretty long from a user perspective, and we might want to look into the indexing?

In any case, I'll leave it up to you; I just prefer `{$exists: true, $ne:[]}` since aesthetically it translates better to "is set, and isn't empty", but may just be me.

model/change_point_detection.go
```diff
@@ -328,6 +334,75 @@ func ReplaceChangePoints(ctx context.Context, env cedar.Environment, performance
 	return catcher.Resolve()
 }
 
+func GetTotalPagesForChangePointsGroupedByVersion(ctx context.Context, env cedar.Environment, projectId string, pageSize int) (int, error) {
+	countKey := "count"
+	pipe := appendAfterBaseGetChangePointsByVersionAgg(projectId, bson.M{
```
ancostas, line 339: Why do we need a separate function to append? Seems like we could just concatenate the stages it produces with other stages in any order we wish?

model/change_point_detection.go
```diff
@@ -328,6 +334,75 @@ func ReplaceChangePoints(ctx context.Context, env cedar.Environment, performance
...
+			"$match": bson.M{
+				bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoProjectKey):             projectId,
+				bsonutil.GetDottedKeyName(perfAnalysisKey, perfAnalysisChangePointsKey, "0"): bson.M{"$exists": true},
+			},
+		},
+		{
+			"$group": bson.M{
+				"_id":          "$info.version",
+				"perf_results": bson.M{"$push": "$$ROOT"},
+				"version_id":   bson.M{"$first": "$info.version"},
```
ancostas, line 372: Is this necessary, since it's already in `_id`?

model/change_point_detection.go
```diff
@@ -328,6 +334,75 @@ func ReplaceChangePoints(ctx context.Context, env cedar.Environment, performance
...
+				"order":        bson.M{"$first": "$info.order"},
```
ancostas, line 373: @tychoish question for you: is there any case where these orders are wildly different for a single version? w.r.t. chrono sort.

model/change_point_detection.go
zamj: These are the starting stages for both the totalPages aggregation and the getting the changepoints aggregation, I don't think there's a case, at least right now, where it makes sense to put these stages in any other order. As well, this change was originally requested by Sam [here](https://github.com/evergreen-ci/cedar/pull/200#discussion_r407820378).
 
If this is something you feel strongly about, please discuss with him so that you're not at cross purposes.

model/change_point_detection.go
zamj: I thought it seemed a bit clearer to have it be an explicit field but its not needed. I removed it

model/change_point_detection.go
ancostas: I mean, I don't think that appendAfterSpecificSituation function is very useful anywhere else, so it seems like cleaning it up to be something like createStandardPipelineStage would make it more reusable since you can do whatever you want with that stage. I don't want to hold up this PR too much though, so go ahead like this.

rest/service.go
ancostas: I mean, to me the route is "Get versions with change points under a specific project", so it seems like the resource would be /projects/p_id/versions/change_points. Or /projects/p_id/versions?include=change_points?

I don't really care though, other than I find diverging from the agreed design a bit hard to scale if we had more people working on this asynchronously. Since we don't, go ahead and make the route whatever makes sense to you two, but please update the technical design such that whomever catches the UI ticket for this  knows what the correct API endpoint is.

zamj: I think Alex's take on how to think about this route is the correct one. This does seem to be very contentious though and I think is one of the last pieces to resolve on this PR. If you want to discuss further Sam, we can set up a quick call to talk through and hopefully come to an agreement.

PR #201, opened 2020-04-13
MAKE-1239: Implement the TestResults model 

julianedwards:


model/test_results.go
```diff
@@ -0,0 +1,309 @@
...
+
+	return errors.Wrapf(err, "problem setting test recorded for %s", t.ID)
+}
+
+// TestResultsInfo describes information unique to a single task execution.
+type TestResultsInfo struct {
+	Project   string `bson:"project,omitempty"`
+	Version   string `bson:"version,omitempty"`
+	Variant   string `bson:"variant,omitempty"`
+	Distro    string `bson:"distro,omitempty"`
```
tychoish, line 256: seems odd to have this here

model/test_results.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	return fmt.Sprintf("%x", hash.Sum(nil))
+}
+
+// TestResult describes a single test result to be stored as a BSON object in
+// some type of pail bucket storage.
+type TestResult struct {
+	TestName       string    `bson:"test_name"`
+	TestFile       string    `bson:"test_file"`
```
tychoish, line 301: what's the difference between test name and test file

my gut says it'd be good to also have the taskname here just for sanity later, but I'm negotiable.

model/test_results_storage.go
```diff
@@ -0,0 +1,22 @@
...
+// TestResultsArtifactInfo describes a bucket of test results for a given task
+// execution stored in some kind of offline storage. It is the bridge between
+// pail-backed offline log storage and the cedar-based log metadata storage.
+// The prefix field indicates the name of the "sub-bucket". The top level
+// bucket is accesible via the cedar.Environment interface.
+type TestResultsArtifactInfo struct {
+	Type    PailType        `bson:"type"`
+	Prefix  string          `bson:"prefix"`
+	Version int             `bson:"version"`
+	Tests   map[string]bool `bson:"tests,omitempty"`
```
tychoish, line 14: what's the bool? I think a []string{} is better, but I still don't know that this makes sense in the DB

julianedwards: The bool is to track if it has been accounted for in the historical stats data, in case we have to do a backfill

model/test_results.go
julianedwards: it is used for the historical stats data

model/test_results.go
julianedwards: The test file is the name where the actual test data is written to on the host. Not sure if it is absolutely necessary but it is part of the _id in the current historical test stats.

model/test_results.go
```diff
@@ -181,79 +183,112 @@ func (t *TestResults) Append(ctx context.Context, results []TestResult) error {
...
+// Download returns a TestResult slice with the corresponding results stored in
+// the offline blob storage. The TestResults should be populated and the
+// environment should not be nil.
+func (t *TestResults) Download(ctx context.Context) ([]TestResult, error) {
+	if !t.populated {
+		return []TestResult{}, errors.New("cannot download with populated test results")
+	}
 	if t.env == nil {
-		return errors.New("cannot append with a nil environment")
+		return []TestResult{}, errors.New("cannot download test results with a nil environment")
```
tychoish, line 197: nil, rather than empty?

model/test_results_test.go
```diff
@@ -0,0 +1,468 @@
...
+		Trial:          seedRand.Intn(10),
+		Status:         "Pass",
+		LineNum:        seedRand.Intn(1000),
+		TaskCreateTime: time.Now().Add(-time.Hour).UTC().Round(time.Millisecond),
+		TestStartTime:  time.Now().Add(-30 * time.Hour).UTC().Round(time.Millisecond),
+		TestEndTime:    time.Now().UTC().Round(time.Millisecond),
+	}
+}
+
+var seedRand *rand.Rand = rand.New(rand.NewSource(time.Now().UnixNano()))
```
tychoish, line 468: I'd just call rand.Seed() in init....  and call the rand package's init functions.

model/test_results_storage.go
```diff
@@ -8,15 +8,13 @@ import "github.com/mongodb/anser/bsonutil"
 // The prefix field indicates the name of the "sub-bucket". The top level
 // bucket is accesible via the cedar.Environment interface.
 type TestResultsArtifactInfo struct {
-	Type    PailType        `bson:"type"`
-	Prefix  string          `bson:"prefix"`
-	Version int             `bson:"version"`
-	Tests   map[string]bool `bson:"tests,omitempty"`
+	Type    PailType `bson:"type"`
+	Prefix  string   `bson:"prefix"`
+	Version int      `bson:"version"`
```
tychoish, line 13: don't we need to save timestamps for when analysis runs ran?

model/test_results.go
```diff
@@ -181,79 +183,112 @@ func (t *TestResults) Append(ctx context.Context, results []TestResult) error {
...
-		"collection":   testResultsCollection,
-		"id":           t.ID,
-		"updateResult": updateResult,
-		"test_names":   testNames,
-		"op":           "append test names to a test results record",
-	})
-	if err == nil && updateResult.MatchedCount == 0 {
-		err = errors.Errorf("could not find test results record with id %s in the database", t.ID)
+	if err != nil {
+		return []TestResult{}, errors.Wrap(err, "problem creating bucket")
```
tychoish, line 219: return nil instead of the empty slice.

model/test_results.go
```diff
@@ -181,79 +183,112 @@ func (t *TestResults) Append(ctx context.Context, results []TestResult) error {
...
+	}
+
+	results := []TestResult{}
+	for iter.Next(ctx) {
+		r, err := iter.Item().Get(ctx)
+		if err != nil {
+			return []TestResult{}, errors.Wrap(err, "problem getting test result")
+		}
+		defer func() {
+			_ = r.Close()
```
tychoish, line 234: is this error not interesting?

model/test_results.go
```diff
@@ -181,79 +183,112 @@ func (t *TestResults) Append(ctx context.Context, results []TestResult) error {
...
+		return []TestResult{}, errors.Wrap(err, "problem creating bucket")
+	}
+
+	iter, err := bucket.List(ctx, "")
+	if err != nil {
+		return []TestResult{}, errors.Wrap(err, "problem listing bucket items")
+	}
+
+	results := []TestResult{}
+	for iter.Next(ctx) {
```
tychoish, line 228: I suspect that we may want to return some kind of iterator, rather than reading all of the results into memory at once. I'm not sure that we need to do this right now, and this logic does work. 

An iterator around a buffered channel might be good, or at least having a basic iterator type that we could try (and race?) different implementations, even if we start with this one.

model/test_results.go
```diff
@@ -181,79 +183,112 @@ func (t *TestResults) Append(ctx context.Context, results []TestResult) error {
...
+		defer func() {
+			_ = r.Close()
+		}()
+
+		data, err := ioutil.ReadAll(r)
+		if err != nil {
+			return []TestResult{}, errors.Wrap(err, "problem reading test result data")
+		}
+		result := TestResult{}
+		if err = bson.Unmarshal(data, &result); err != nil {
```
tychoish, line 242: I think it might be a fun exercise to see what the bson unmarshal speed would be like vs writing DocumentMarshal and DocumentUnmarshaler functions in birch?

model/test_results_storage.go
julianedwards: no I don't think its necessary per our discussion on Friday. We will just have a job to recalculate historical stats for a given range of dates.

model/test_results.go
julianedwards: Yeah given that there will may be 10s of thousands of results per task that is probably a good idea. Definitely something we can push off until I get to the rest code.

model/test_results.go
julianedwards: I guess I can log it, but its in a defer statement so can't do much about it

model/test_results.go
tychoish: open a ticket to track

model/test_results.go
```diff
@@ -216,31 +216,34 @@ func (t *TestResults) Download(ctx context.Context) ([]TestResult, error) {
...
 		if err != nil {
-			return []TestResult{}, errors.Wrap(err, "problem getting test result")
+			return nil, errors.Wrap(err, "problem getting test result")
 		}
 		defer func() {
-			_ = r.Close()
+			err = r.Close()
+			grip.Error(message.WrapError(err, message.Fields{
+				"message": "problem closing test results bucket iterator",
+			}))
```
tychoish, line 237: there's no need to change because I think this is all going to change but, this is really not all that different from: 
```
grip.Error(message.WrapError(r.Close(), "problem closing test results bucket iterator"))
```
Thought there's other things you can put in the message, for more context, including what test it's coming from/bucket name, etc.

PR #204, opened 2020-04-16
MAKE-1213: Restrict calculation of change points to the portion of the time series after the latest triaged change point

ancostas:
Sorry this got a bit involved on the testing side; I felt our current test was a bit too rigid (is this passing because it's a super simple case?) and needed to be a bit more dynamic (mimicking what cedar actually will encounter).

Now it:
* Generates a random set of time series, of random lengths
* Generates a random number of change points in these time series
* Feeds them in randomly grouped under perf results (rollups can be added/removed over time)
* Ensures that they all get detected and persisted as you would expect them to be

Also, I added a test the restricted analysis this ticket is actually about :smile:

### DIFF
--- a/model/change_point_detection.go
+++ b/model/change_point_detection.go
@@ -262,7 +262,7 @@ func GetPerformanceData(ctx context.Context, env cedar.Environment, performanceR
 		},
 		//Filter out any change points unrelated to this rollup
 		{
-			"$set": bson.M{
+			"$addFields": bson.M{
 				bsonutil.GetDottedKeyName(perfAnalysisKey, perfAnalysisChangePointsKey): bson.M{
 					"$filter": bson.M{
 						"input": "$" + bsonutil.GetDottedKeyName(perfAnalysisKey, perfAnalysisChangePointsKey),
@@ -302,7 +302,7 @@ func GetPerformanceData(ctx context.Context, env cedar.Environment, performanceR
 		},
 		// Flatter the change points into one array
 		{
-			"$set": bson.M{
+			"$addFields": bson.M{
 				"change_points": bson.M{
 					"$reduce": bson.M{
 						"input":        "$change_points",
--- a/model/change_point_detection.go
+++ b/model/change_point_detection.go
@@ -260,10 +260,14 @@ func GetPerformanceData(ctx context.Context, env cedar.Environment, performanceR
 		{
 			"$unwind": "$" + bsonutil.GetDottedKeyName(perfRollupsKey, perfRollupsStatsKey),
 		},
+		//Filter out any change points unrelated to this rollup
 		{
-			"$unwind": bson.M{
-				"path": "$" + bsonutil.GetDottedKeyName(perfAnalysisKey, perfAnalysisChangePointsKey),
-				"preserveNullAndEmptyArrays": true,
+			"$filter": bson.M{
+				"input": "$" + bsonutil.GetDottedKeyName(perfRollupsKey, perfRollupsStatsKey),
+				"as": "cp",
+				"cond": bson.M{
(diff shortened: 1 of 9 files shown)


ancostas:
Not sure why test-units is failing... looks like an off by one error I handled early in setting this up, and this passes for me locally :raised_eyebrow:  -- will look into it tomorrow; feel free to make the initial pass in the meantime.

ancostas:
@tychoish Idk what's wrong here; this works for me locally. It's like it's not able to get results using the same aggregation when run through evergreen?

model/change_point_detection.go
```diff
@@ -428,3 +461,18 @@ func createChangePoint(ctx context.Context, env cedar.Environment, resultToUpdat
...
+		perfIDKey: perfResultId,
+		bsonutil.GetDottedKeyName(perfAnalysisKey, perfAnalysisChangePointsKey, perfChangePointMeasurementKey): measurement,
+	}
+	update := bson.M{
+		"$set": bson.M{
+			bsonutil.GetDottedKeyName(perfAnalysisKey, perfAnalysisChangePointsKey, "$", perfChangePointTriageKey, perfTriageInfoStatusKey):    status,
+			bsonutil.GetDottedKeyName(perfAnalysisKey, perfAnalysisChangePointsKey, "$", perfChangePointTriageKey, perfTriageInfoTriagedOnKey): time.Now(),
+		},
+	}
+	_, err := env.GetDB().Collection(perfResultCollection).UpdateOne(ctx, filter, update)
```
tychoish, line 476: you have to check the result (first returned value). the error value is only non-nil when there's a connection, parsing, or other related error, if you failed to find or update a document, that's in the result.

model/change_point_detection.go
```diff
@@ -428,3 +461,18 @@ func createChangePoint(ctx context.Context, env cedar.Environment, resultToUpdat
 	_, err := env.GetDB().Collection(perfResultCollection).UpdateOne(ctx, filter, update)
 	return errors.Wrap(err, "Unable to create change point")
 }
+
+func TriageChangePoint(ctx context.Context, env cedar.Environment, perfResultId string, measurement string, status TriageStatus) error {
```
tychoish, line 465: I'd be consistent about using ID rather than Id

tychoish:
I'm also worried about the failing test, or if we need to be careful or should expect anything unusual in the deploy.

model/change_point_detection.go
ancostas: Sorry, force of habit!

ancostas:
@tychoish Tests were failing since we use 4.0.13 when testing through evergreen, and `$set` operator in aggregations isn't supported until 4.2. This was... pretty troublesome to debug.

units/recalculate_change_points_test.go
```diff
@@ -10,21 +12,95 @@ import (
...
+				break
+			}
+		}
+	}
+	return newVals
+}
+
+func makePerfResultsWithChangePoints(unique string, seed int64) ([]testResultsAndRollups, map[string][]int) {
+	// deterministic testing on failure
+	fmt.Println("Seed for recalculate test: " + strconv.FormatInt(seed, 10))
```
tychoish, line 46: can you omit printing in tests (use the logger, if you must.)

units/recalculate_change_points_test.go
```diff
@@ -34,33 +110,35 @@ func makePerfResults(length int, breakpoint int, unique string) []TestResultsAnd
...
+	for measurement, cpIndices := range timeSeriesChangePoints {
+		measurementName := "measurement_" + strconv.Itoa(measurement)
+		changePoints[measurementName] = append(changePoints[measurementName], cpIndices...)
+	}
+
+	return rollups, changePoints
 }
 
-func init() {
+func setup() {
```
tychoish, line 141: nit: give this a less generic name

PR #205, opened 2020-04-17
MAKE-1214: Add triage endpoint to cedar

ancostas:
@tychoish We're transacting!!!


model/change_point_detection.go
```diff
@@ -428,3 +434,42 @@ func createChangePoint(ctx context.Context, env cedar.Environment, resultToUpdat
...
+	session, err := env.GetDB().Client().StartSession()
+	if err != nil {
+		return errors.Wrap(err, "Unable to start session for triage update")
+	}
+	defer session.EndSession(ctx)
+	if err := session.StartTransaction(); err != nil {
+		return errors.Wrap(err, "Unable to start transaction for triage update")
+	}
+	return mongo.WithSession(ctx, session, func(sc mongo.SessionContext) error {
+		for perfResultId, measurement := range changePoints {
```
tychoish, line 448: how many change points do we expect there to be typically and maximally?

model/change_point_detection.go
```diff
@@ -428,3 +434,42 @@ func createChangePoint(ctx context.Context, env cedar.Environment, resultToUpdat
...
+			fmt.Print(res)
+			if err != nil || res.ModifiedCount != 1 {
+				err2 := session.AbortTransaction(ctx)
+				if err2 != nil {
+					return errors.Wrap(err, "Failed to abort transaction during failed change point triage")
+				}
+				return errors.New(fmt.Sprintf("Unable to triage change point <%s> for performance result %s", measurement, perfResultId))
+			}
+		}
+		err := session.CommitTransaction(sc)
```
tychoish, line 469: do we need to retry this? is there special retry logic?

model/change_point_detection.go
```diff
@@ -428,3 +434,42 @@ func createChangePoint(ctx context.Context, env cedar.Environment, resultToUpdat
...
+				},
+			}
+			res, err := env.GetDB().Collection(perfResultCollection).UpdateOne(sc, filter, update)
+			fmt.Print(res)
+			if err != nil || res.ModifiedCount != 1 {
+				err2 := session.AbortTransaction(ctx)
+				if err2 != nil {
+					return errors.Wrap(err, "Failed to abort transaction during failed change point triage")
+				}
+				return errors.New(fmt.Sprintf("Unable to triage change point <%s> for performance result %s", measurement, perfResultId))
```
tychoish, line 466: errors.Errorf

model/change_point_detection.go
```diff
@@ -428,3 +434,42 @@ func createChangePoint(ctx context.Context, env cedar.Environment, resultToUpdat
...
+				bsonutil.GetDottedKeyName(perfAnalysisKey, perfAnalysisChangePointsKey, perfChangePointMeasurementKey): measurement,
+			}
+			update := bson.M{
+				"$set": bson.M{
+					bsonutil.GetDottedKeyName(perfAnalysisKey, perfAnalysisChangePointsKey, "$", perfChangePointTriageKey, perfTriageInfoStatusKey):    status,
+					bsonutil.GetDottedKeyName(perfAnalysisKey, perfAnalysisChangePointsKey, "$", perfChangePointTriageKey, perfTriageInfoTriagedOnKey): time.Now(),
+				},
+			}
+			res, err := env.GetDB().Collection(perfResultCollection).UpdateOne(sc, filter, update)
+			fmt.Print(res)
```
tychoish, line 460: omit

rest/data/interface.go
```diff
@@ -44,6 +44,8 @@ type Connector interface {
 	// processing recalculation job has been scheduled for each type of
 	// test (project/variant/task/test combo).
 	ScheduleSignalProcessingRecalculateJobs(context.Context) error
+	// TriageChangePoints returns once all change points have been successfully marked
+	TriageChangePoints(ctx context.Context, changePoints map[string]string, status string) error
```
tychoish, line 48: what is this changepoint map?

rest/data/perf.go
```diff
@@ -265,6 +265,30 @@ func (dbc *DBConnector) ScheduleSignalProcessingRecalculateJobs(ctx context.Cont
...
+	PerfResultID string `bson:"perf_result_id" json:"perf_result_id" yaml:"perf_result_id"`
+	Measurement  string `bson:"measurement" json:"measurement" yaml:"measurement"`
+}
+
+func (dbc *DBConnector) TriageChangePoints(ctx context.Context, changePoints map[string]string, status string) error {
+	ts := model.TriageStatus(status)
+	err := ts.Validate()
+	if err != nil {
+		return gimlet.ErrorResponse{
+			StatusCode: http.StatusUnprocessableEntity,
```
tychoish, line 278: is this something that users should retry?

rest/data/perf.go
```diff
@@ -265,6 +265,30 @@ func (dbc *DBConnector) ScheduleSignalProcessingRecalculateJobs(ctx context.Cont
...
 
+type ChangePointStub struct {
+	PerfResultID string `bson:"perf_result_id" json:"perf_result_id" yaml:"perf_result_id"`
+	Measurement  string `bson:"measurement" json:"measurement" yaml:"measurement"`
+}
+
+func (dbc *DBConnector) TriageChangePoints(ctx context.Context, changePoints map[string]string, status string) error {
+	ts := model.TriageStatus(status)
+	err := ts.Validate()
+	if err != nil {
```
tychoish, line 276: combine 275/276

rest/data/perf.go
```diff
@@ -265,6 +265,30 @@ func (dbc *DBConnector) ScheduleSignalProcessingRecalculateJobs(ctx context.Cont
...
+		return gimlet.ErrorResponse{
+			StatusCode: http.StatusUnprocessableEntity,
+			Message:    fmt.Sprintf("Invalid triage status: %s", status),
+		}
+	}
+	err = model.TriageChangePoints(ctx, dbc.env, changePoints, ts)
+	if err != nil {
+		return gimlet.ErrorResponse{
+			StatusCode: http.StatusInternalServerError,
+			Message:    "Could not triage change points",
```
tychoish, line 286: do you want this message to annotate/log something about the cause of the error passed?

rest/perf_routes.go
```diff
@@ -358,7 +358,70 @@ func (h *perfSignalProcessingRecalculateHandler) Run(ctx context.Context) gimlet
...
+// POST /perf/change_points/triage/mark
+
+type changePointStub struct {
+	PerfResultID string `bson:"perf_result_id" json:"perf_result_id" yaml:"perf_result_id"`
+	Measurement  string `bson:"measurement" json:"measurement" yaml:"measurement"`
+}
+
+type changePointMarkRequest struct {
+	Status       string            `bson:"status" json:"status" yaml:"status"`
+	ChangePoints []changePointStub `bson:"change_points" json:"change_points" yaml:"change_points"`
```
tychoish, line 379: why stub? can we use an existing type from elsewhere?

rest/perf_routes.go
```diff
@@ -358,7 +358,70 @@ func (h *perfSignalProcessingRecalculateHandler) Run(ctx context.Context) gimlet
...
+	}
+	return nil
+}
+
+func (h *perfChangePointTriageMarkHandler) Run(ctx context.Context) gimlet.Responder {
+	cps := map[string]string{}
+	for _, stub := range h.req.ChangePoints {
+		cps[stub.PerfResultID] = stub.Measurement
+	}
+	err := h.sc.TriageChangePoints(ctx, cps, h.req.Status)
```
tychoish, line 418: I'd be inclined to pass the slice rather than a map around.

model/change_point_detection.go
ancostas: I think the build barons typically triage all of them in a given version using multiselect, and I'm guessing ~10 per version?

So typically probably ~10 at a time, unless they select multiple versions --> then it would be 10*verions, which for a given page size they are using would max out around ~10*page_size. Page size would probably be 50, so maximally ~500 in typical situations.

model/change_point_detection.go
ancostas: whoops!

model/change_point_detection.go
ancostas: My take on this is we'd be retrying due to connection issues --> and that's something the driver itself should manage, right? Otherwise if the transaction simply fails to commit, there is not much we can do, or?

rest/data/interface.go
ancostas: A map of perf result id to measurement being triaged. I tried this with a slice of anonymous structs, but was having some problems having the datatypes match up. The spirit of this is to:
* Pass the information around in a way that's agnostic of which layer you are in
* Not have to map the data over to new structs in each passthrough layer

How do you typically do that?

rest/data/perf.go
ancostas: Not automatically; this is basically invalid input and they need to fix their request.

rest/data/perf.go
ancostas: Updated to the following:
* If one of the results was not found for update, tell client which one w/ 404 response so they can fixup request; don't pollute logs with client request errors
* If this is not client error, log it, and only give generic information to client under 500 status code (don't leak implementation details)

ancostas: Only thing that I find a little janky is doing strings.contains on the error message. Surely there's a better way?

rest/perf_routes.go
ancostas: I don't think anything has this interface? It's meant to only be used to decode client request.

rest/perf_routes.go
ancostas: So you mentioned this earlier:
* I don't want to pass a slice of structs in this package down to model package (circular dep)
* I had issues with passing slices of an anonymous struct

Was I just doing point 2 the wrong way?

model/change_point_detection.go
tychoish: I mean it could fail to commit because of some kind of write conflict, and those aren't automatically retried. I'm also not sure that the driver actually retries due to connection issues either

model/change_point_detection.go
```diff
@@ -428,3 +433,44 @@ func createChangePoint(ctx context.Context, env cedar.Environment, resultToUpdat
...
+			res, err := env.GetDB().Collection(perfResultCollection).UpdateOne(sc, filter, update)
+			if err != nil {
+				err2 := session.AbortTransaction(ctx)
+				if err2 != nil {
+					return errors.Wrap(err, "Failed to abort transaction during failed change point triage")
+				}
+				return errors.Errorf("Unable to triage change point <%s> for performance result %s", measurement, perfResultId)
+			}
+			if res.ModifiedCount != 1 {
+				return errors.Errorf("Could not find change point <%s> for performance result %s", measurement, perfResultId)
```
john-m-liu, line 467: You probably want to abort here as well

model/change_point_detection.go
john-m-liu: The driver doesn't automatically retry, but it should bubble up transient errors such as there not being an available write ticket within the transaction timeout. I'd expect that cedar's db to not hit this at the moment, but it's worth retrying so that it doesn't become a problem in the future

model/change_point_detection.go
```diff
@@ -428,3 +433,44 @@ func createChangePoint(ctx context.Context, env cedar.Environment, resultToUpdat
...
+		return errors.Wrap(err, "Unable to start transaction for triage update")
+	}
+	return mongo.WithSession(ctx, session, func(sc mongo.SessionContext) error {
+		for perfResultId, measurement := range changePoints {
+			filter := bson.M{
+				perfIDKey: perfResultId,
+				bsonutil.GetDottedKeyName(perfAnalysisKey, perfAnalysisChangePointsKey, perfChangePointMeasurementKey): measurement,
+			}
+			update := bson.M{
+				"$set": bson.M{
```
john-m-liu, line 453: The update is the same for all documents, so I think you could do this with a single update command without a transaction right?

rest/perf_routes.go
```diff
@@ -358,7 +358,70 @@ func (h *perfSignalProcessingRecalculateHandler) Run(ctx context.Context) gimlet
...
+//
+// POST /perf/change_points/triage/mark
+
+type changePointStub struct {
+	PerfResultID string `bson:"perf_result_id" json:"perf_result_id" yaml:"perf_result_id"`
+	Measurement  string `bson:"measurement" json:"measurement" yaml:"measurement"`
+}
+
+type changePointMarkRequest struct {
+	Status       string            `bson:"status" json:"status" yaml:"status"`
```
john-m-liu, line 378: These probably don't need bson or yaml tags

model/change_point_detection.go
ancostas: Should we wrap the driver in a client then? It seems like this retry logic should be implemented wherever driver/client is consumed, rather than separately all over the place -- right?

ancostas: On a separate not; it sounds like this improvement would apply to most of the codebase and could be done in a separate tix?

model/change_point_detection.go
ancostas: Nice catch. FMI: if we don't commit before returning, transaction will abort on its own as a safeguard, right?

model/change_point_detection.go
ancostas: I think a transaction is needed in order to abort if we don't modify everything we expected to find. Additionally, it seems like reporting 404 for a miss is harder if we don't do these individually.

rest/data/interface.go
ancostas: *Update* Anonymous structs don't work since this is a separate package.

An option would be to define something in the `model` package; but defining a struct the REST API is expecting in `model` seems very wrong. And mapping over structs in each layer seems inefficient/a lot of boilerplate.

rest/perf_routes.go
ancostas: Update above.

model/change_point_detection.go
john-m-liu: That's reasonable, but I just wanted to double check - the only way for each update to fail is if that measurement variable that is passed in does not exist for the ID passed in. Under what scenarios would that happen?

ancostas:
in the middle of trying to make this work with an updatemany -- in case someone reviews in the meantime

model/change_point_detection.go
```diff
@@ -428,3 +428,60 @@ func createChangePoint(ctx context.Context, env cedar.Environment, resultToUpdat
...
+					if cp.Measurement == stub.Measurement {
+						goto cont
+					}
+				}
+			}
+		}
+		return errors.Errorf("Could not find change point <%s> for performance result %s", stub.Measurement, stub.PerfResultID)
+	cont:
+		continue
+	}
```
ancostas, line 473: Something like this is needed since performance results can have multiple change points being triaged at once.

model/change_point_detection.go
```diff
@@ -428,3 +428,60 @@ func createChangePoint(ctx context.Context, env cedar.Environment, resultToUpdat
...
+
+	update := bson.M{
+		"$set": bson.M{
+			bsonutil.GetDottedKeyName(perfAnalysisKey, perfAnalysisChangePointsKey, "$", perfChangePointTriageKey, perfTriageInfoStatusKey):    status,
+			bsonutil.GetDottedKeyName(perfAnalysisKey, perfAnalysisChangePointsKey, "$", perfChangePointTriageKey, perfTriageInfoTriagedOnKey): time.Now(),
+		},
+	}
+	for _, cond := range conditions {
+		if _, err := env.GetDB().Collection(perfResultCollection).UpdateOne(ctx, cond, update); err != nil {
```
ancostas, line 482: I can't see how to make this an updatemany.

```
replset:PRIMARY> filter
{
	"$or" : [
		{
			"_id" : "82d9f0c1a0355dc54995ff88656af1c8a3fb14f1",
			"analysis.change_points.measurement" : "measurement"
		},
		{
			"_id" : "b87a1d2a7c19c263cad984e245f7ae222e31cb29",
			"analysis.change_points.measurement" : "measurement"
		},
		{
			"_id" : "fd2cf39bbc6d68bf1c6e462586abf73d8668db29",
			"analysis.change_points.measurement" : "measurement"
		},
		{
			"_id" : "ccdb2eb4ee3da56ee2d37cc4c77136d559fd5b99",
			"analysis.change_points.measurement" : "measurement"
		},
		{
			"_id" : "a8d4723706b194924df79c47fb32a81a4a9d4b33",
			"analysis.change_points.measurement" : "measurement"
		}
	]
}
replset:PRIMARY> update
{
	"$set" : {
		"analysis.change_points.$.triage.triage_status" : "true_positive",
		"analysis.change_points.$.triage.triaged_on" : "2020-04-21T13:30:39.326068-04:00"
	}
}
replset:PRIMARY> db.perf_results.updateMany(filter, update)
2020-04-21T13:48:42.454-0400 E  QUERY    [js] WriteError({
	"index" : 0,
	"code" : 2,
	"errmsg" : "The positional operator did not find the match needed from the query.",
	"op" : {
		"q" : {
			"$or" : [
				{
					"_id" : "82d9f0c1a0355dc54995ff88656af1c8a3fb14f1",
					"analysis.change_points.measurement" : "measurement"
				},
				{
					"_id" : "b87a1d2a7c19c263cad984e245f7ae222e31cb29",
					"analysis.change_points.measurement" : "measurement"
				},
				{
					"_id" : "fd2cf39bbc6d68bf1c6e462586abf73d8668db29",
					"analysis.change_points.measurement" : "measurement"
				},
				{
					"_id" : "ccdb2eb4ee3da56ee2d37cc4c77136d559fd5b99",
					"analysis.change_points.measurement" : "measurement"
				},
				{
					"_id" : "a8d4723706b194924df79c47fb32a81a4a9d4b33",
					"analysis.change_points.measurement" : "measurement"
				}
			]
		},
		"u" : {
			"$set" : {
				"analysis.change_points.$.triage.triage_status" : "true_positive",
				"analysis.change_points.$.triage.triaged_on" : "2020-04-21T13:30:39.326068-04:00"
			}
		},
		"multi" : true,
		"upsert" : false
	}
}) :
WriteError({
	"index" : 0,
	"code" : 2,
	"errmsg" : "The positional operator did not find the match needed from the query.",
	"op" : {
		"q" : {
			"$or" : [
				{
					"_id" : "82d9f0c1a0355dc54995ff88656af1c8a3fb14f1",
					"analysis.change_points.measurement" : "measurement"
				},
				{
					"_id" : "b87a1d2a7c19c263cad984e245f7ae222e31cb29",
					"analysis.change_points.measurement" : "measurement"
				},
				{
					"_id" : "fd2cf39bbc6d68bf1c6e462586abf73d8668db29",
					"analysis.change_points.measurement" : "measurement"
				},
				{
					"_id" : "ccdb2eb4ee3da56ee2d37cc4c77136d559fd5b99",
					"analysis.change_points.measurement" : "measurement"
				},
				{
					"_id" : "a8d4723706b194924df79c47fb32a81a4a9d4b33",
					"analysis.change_points.measurement" : "measurement"
				}
			]
		},
		"u" : {
			"$set" : {
				"analysis.change_points.$.triage.triage_status" : "true_positive",
				"analysis.change_points.$.triage.triaged_on" : "2020-04-21T13:30:39.326068-04:00"
			}
		},
		"multi" : true,
		"upsert" : false
	}
})
WriteError@src/mongo/shell/bulk_api.js:458:48
mergeBatchResults@src/mongo/shell/bulk_api.js:855:49
executeBatch@src/mongo/shell/bulk_api.js:919:13
Bulk/this.execute@src/mongo/shell/bulk_api.js:1163:21
DBCollection.prototype.updateMany@src/mongo/shell/crud_api.js:666:17
@(shell):1:1
replset:PRIMARY> db.perf_results.update(filter["$or"][0], update)
WriteResult({ "nMatched" : 1, "nUpserted" : 0, "nModified" : 1 })
```

rest/data/interface.go
ancostas: I've moved the struct into `model` package; it still seems janky for the REST layer to know about db models.

model/change_point_detection.go
ancostas: If we can't do an updatemany, I think we might need to revert back to the transactional implementation.

tychoish: you can do a BulkWrite api with many UpdateOnes  (ordered=true) which isn't trasnactional, but will do the thing you want in a single network round trip

I think it's safe to avoid the transaction if the system can recover for a partially applied updatemany

model/change_point_detection.go
```diff
@@ -434,43 +429,59 @@ func createChangePoint(ctx context.Context, env cedar.Environment, resultToUpdat
 	return errors.Wrap(err, "Unable to create change point")
 }
 
-func TriageChangePoints(ctx context.Context, env cedar.Environment, changePoints map[string]string, status TriageStatus) error {
-	session, err := env.GetDB().Client().StartSession()
+type ChangePointStub struct {
```
tychoish, line 432: Why is this called stub? can it just be ChangePointInfo or similar?

ancostas:
@tychoish updated. Idk that there's any recovery mode other than BB's unmarking change points through the API. They'll probably be more interested in marking the ones that couldn't be applied in the first place.

model/change_point_detection.go
```diff
@@ -478,10 +480,15 @@ func TriageChangePoints(ctx context.Context, env cedar.Environment, changePoints
...
 	for _, cond := range conditions {
-		if _, err := env.GetDB().Collection(perfResultCollection).UpdateOne(ctx, cond, update); err != nil {
-			return errors.Wrap(err, "Could not perform triaging update")
-		}
+		operations = append(operations, &mongo.UpdateOneModel{Filter: cond, Update: update})
+	}
+
+	bulkOptions := options.BulkWriteOptions{}
+	bulkOptions.SetOrdered(true)
+	if _, err := env.GetDB().Collection(perfResultCollection).BulkWrite(ctx, operations, &bulkOptions); err != nil {
```
tychoish, line 490: I'd just do `BulkWrite(ctx, ops, options.BulkWrite().SetOrdered(true))`

tychoish:
Just a little nit, plus the merge.

My question about recovery is as much about the operational story if something goes wrong (or say we want to clear change points and recalculate, or we do some kind of step-back action, and need to remove previous changepoints to better find the point in question, I'm sort of making things up here,) I want to make sure that "fixing things" doesn't require changing something in the database. Otherwise I think this will be fine.

ancostas:
@tychoish fixed, and dropped a meeting on your cal to go over the recovery stuff.

model/change_point_detection.go
```diff
@@ -462,6 +464,65 @@ func createChangePoint(ctx context.Context, env cedar.Environment, resultToUpdat
...
+	if err := cur.All(ctx, &results); err != nil {
+		return errors.Wrap(err, "Could not decode performance results for triage")
+	}
+
+	for _, stub := range changePoints {
+		for _, res := range results {
+			if res.ID == stub.PerfResultID {
+				for _, cp := range res.Analysis.ChangePoints {
+					if cp.Measurement == stub.Measurement {
+						goto cont
```
john-m-liu, line 500: avoid using gotos since they make it really hard to tell where in the execution stack you are. Instead add a loop label to the change points loop and continue to that here

ancostas: Neat, I didn't know you could do that in Go!

zamj:
I'm a little scared by the recovery case as it stands in this Pr. Can you let me know the outcome of your meeting @ancostas ?

PR #208, opened 2020-04-21
MAKE-1240: Implement the HistoricalTestData model

julianedwards:


model/historical_test_data.go
```diff
@@ -0,0 +1,69 @@
...
+	LastUpdate      time.Time `bson:"last_update"`
+
+	env       cedar.Environment
+	populated bool
+}
+
+// CreateHistoricalTestData is an entry point for creating a new
+// HistoricalTestData.
+func CreateHistoricalTestData(info HistoricalTestDataInfo) (*HistoricalTestData, error) {
+	if err := info.validate(); err != nil {
```
tychoish, line 26: why is this private?

model/historical_test_data.go
```diff
@@ -0,0 +1,69 @@
...
+	Project   string    `bson:"project"`
+	Variant   string    `bson:"variant"`
+	TaskName  string    `bson:"task_name"`
+	TestName  string    `bson:"test_name"`
+	Requester string    `bson:"requester"`
+	Date      time.Time `bson:"date"`
+}
+
+func (i *HistoricalTestDataInfo) validate() error {
+	var emptyField bool
```
tychoish, line 49: I'd do a catcher with NewWhen calls and the conditionals so the error message makes it clear what's missing.

model/historical_test_data.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// HistoricalTestDataInfo describes information unique to a single test
+// statistics document.
+type HistoricalTestDataInfo struct {
+	Project   string    `bson:"project"`
+	Variant   string    `bson:"variant"`
+	TaskName  string    `bson:"task_name"`
+	TestName  string    `bson:"test_name"`
+	Requester string    `bson:"requester"`
```
tychoish, line 44: requester is a weird evergreen term that maybe we want to elide/cover up?

model/historical_test_data.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/pkg/errors"
+)
+
+// HistoricalTestData describes aggregated test result data for a given date
+// range.
+type HistoricalTestData struct {
+	Info            HistoricalTestDataInfo
+	NumPass         int       `bson:"num_pass"`
+	NumFail         int       `bson:"num_fail"`
+	AverageDuration float64   `bson:"average_duration"`
```
tychoish, line 16: do we want to store total time in some units? also units are unclear.

model/historical_test_data.go
```diff
@@ -0,0 +1,69 @@
...
+	LastUpdate      time.Time `bson:"last_update"`
+
+	env       cedar.Environment
```
tychoish, line 19: I don't think you're using this (yet? ever?)

model/historical_test_data.go
julianedwards: Would you rather store totalTime and calculate average duration on the fly using NumPass + NumFail?

model/historical_test_data.go
julianedwards: I will because I need to get the configuration to get the test results bucket

julianedwards: I broke this into small tickets as a productivity hackl

model/historical_test_data.go
julianedwards: it doesn't have to be but I didn't see the need to make it public?

model/historical_test_data.go
julianedwards: its used as part of the identity... would you rather call it something else?

julianedwards: I don't think we can get rid of it though

model/historical_test_data.go
tychoish: Just to clarify, this is going to be stored in s3? If that's true (and it probably ought to be?) then we should probably store everything (components and average?)

model/historical_test_data.go
tychoish: nocie.

model/historical_test_data.go
tychoish: We should definitely call it something else, and maybe have a maineline boolean in addition to  some better kind of name.

model/historical_test_data.go
julianedwards: Don't think there is anything else to store here? Average will be calculated as test results come in, don't see a compelling reason to store each individual test duration.

julianedwards: Unless we are planning to do more complex rollups?

model/historical_test_data.go
julianedwards: What is a good substitute? `request_type`? Historical test statistics uses requester as part of the id and its not binary (like mainline), so we need to keep an equivalent field I think.

PR #209, opened 2020-04-22
MAKE-1210: Add percent_change to change points

zamj:


units/recalculate_change_points.go
```diff
@@ -105,24 +105,48 @@ func (j *recalculateChangePointsJob) Run(ctx context.Context) {
...
+		for i, pointIndex := range result {
+			startFirstPartition := 0
+			if i != 0 {
+				startFirstPartition = result[i-1]
+			}
+			endSecondPartition := len(floatSeries)
+			if i != len(result)-1 {
+				endSecondPartition = result[i+1]
+			}
+			percentChange := calculatePercentChange(startFirstPartition, pointIndex, endSecondPartition, floatSeries)
```
ancostas, line 123: I feel like this would be clearer:
```
for i, changePoint := range result {
	var lowerBound int
	var upperBound int

	if i == 0 {
		lowerBound = 0
	} else {
		lowerBound = result[i-1]
	}

	if i == len(result)-1 {
		upperBound = len(floatSeries)
	} else {
		upperBound = result[i+1]
	}

	windowBeforeAverage := getAverageValue(floatSeries[lowerBound:changePoint])
	windowAfterAverage := getAverageValue(floatSeries[changePoint:upperBound])

	change := windowAfterAverage / windowBeforeAverage

	percentChange := 100 * (change - 1)
         mapped := ...etc
}
```

What do you think?

zamj: I've changed it to hopefully be a little more clear and taken most of what you've suggested. What do you think of this?

`			var lowerBound int
			var upperBound int

			if i == 0 {
				lowerBound = 0
			} else {
				lowerBound = result[i-1]
			}

			if i == len(result)-1 {
				upperBound = len(floatSeries)
			} else {
				upperBound = result[i+1]
			}

			lowerWindow := floatSeries[lowerBound:pointIndex]
			upperWindow := floatSeries[pointIndex:upperBound]
			percentChange := calculatePercentChange(lowerWindow, upperWindow)`

units/recalculate_change_points_test.go
```diff
@@ -129,15 +134,47 @@ func makePerfResultsWithChangePoints(unique string, seed int64) ([]testResultsAn
...
+				upperBound = cpIndices[i+1]
+			}
+
+			lowerWindow := convertIntArrayToFloat64(currentTimeSeries[lowerBound:pointIndex])
+			upperWindow := convertIntArrayToFloat64(currentTimeSeries[pointIndex:upperBound])
+
+			percentChange := calculatePercentChange(lowerWindow, upperWindow)
+			cpsWithIndexAndPercentChange[i] = changePointIndexPercentChange{index: pointIndex, percentChange: percentChange}
+		}
+		changePoints[measurementName] = append(changePoints[measurementName], cpsWithIndexAndPercentChange...)
```
ancostas, line 164: This will work. One thing to note is that we are basically checking that we get the percent change in the same way the job itself gets the percent change. It's possible that our average calculations are bugged though, and this test would pass (since it uses the same procedure).

Should we change this to be more of a sanity check? Our test data's stable regions are constant/have no noise, and cps have a minimum spacing in the test, so we could do something like:
```
# No need to average over windows, since they are constant --> just use points before/after the change
percentChange := 100*((currentTimeSeries[pointIndex]/currentTimeSeries[pointIndex-1])-1)
cpsWithIndexAndPercentChange[i] = changePointIndexPercentChange{index: pointIndex, percentChange: percentChange}
```
And drop all of the windowing stuff.

What do you think?

units/recalculate_change_points.go
```diff
@@ -113,24 +113,58 @@ func (j *recalculateChangePointsJob) Run(ctx context.Context) {
...
 		result, err := j.changePointDetector.DetectChanges(ctx, floatSeries)
+		if err != nil {
+			j.AddError(errors.Wrapf(err, "Unable to detect change points in time perfData %s", j.PerformanceResultId))
+			return
+		}
 
 		var changePoints []model.ChangePoint
-		for _, pointIndex := range result {
-			mapped := model.CreateChangePoint(pointIndex+latestTriagedChangePointIndex, perfData.Measurement, j.changePointDetector.Algorithm().Name(), j.changePointDetector.Algorithm().Version(), algorithmConfigurationToOptions(j.changePointDetector.Algorithm().Configuration()))
+		for i, pointIndex := range result {
```
tychoish, line 122: can/should this math be a method on the change point?

units/recalculate_change_points.go
```diff
@@ -113,24 +113,58 @@ func (j *recalculateChangePointsJob) Run(ctx context.Context) {
...
 }
 
+func calculatePercentChange(lowerWindow, upperWindow []float64) float64 {
+	avgLowerWindow := getAverageValue(lowerWindow)
+	avgUpperWindow := getAverageValue(upperWindow)
+
+	return 100 * ((avgUpperWindow / avgLowerWindow) - 1)
+}
+
+func getAverageValue(values []float64) float64 {
```
tychoish, line 160: it seems like these math functions should use library functions, I think we already vendor the following which has a function for this: https://godoc.org/github.com/aclements/go-moremath/stats

We also have/can have gonum and other math libraries and there's no real reason to avoid using one (in most cases.)

units/recalculate_change_points_test.go
```diff
@@ -129,15 +134,47 @@ func makePerfResultsWithChangePoints(unique string, seed int64) ([]testResultsAn
...
 
 	// time to map over change points
-	changePoints := map[string][]int{}
+	changePoints := map[string][]changePointIndexPercentChange{}
 	for measurement, cpIndices := range timeSeriesChangePoints {
 		measurementName := "measurement_" + strconv.Itoa(measurement)
-		changePoints[measurementName] = append(changePoints[measurementName], cpIndices...)
+		cpsWithIndexAndPercentChange := make([]changePointIndexPercentChange, len(cpIndices))
+		currentTimeSeries := timeSeries[measurement]
+		for i, pointIndex := range cpIndices {
```
tychoish, line 142: this feels like we're duplicating production logic in a test.

units/recalculate_change_points_test.go
zamj: Seems good to me, since the regions are stable. I've added your change

units/recalculate_change_points_test.go
zamj: This was changed

units/recalculate_change_points.go
zamj: No I don't think so. This for loop is calculating multiple ChangePoints so adding this whole calculation as a method on a single ChangePoint struct doesn't seem to be a good match. 

Adding just the percent_change calculation as a method doesn't seem to be a good match either as the calculation requires a lot of state to do the calculation for each region. So the actual signature for the method would be pretty unwieldy and wouldn't add much benefit imo.

model/change_point_detection.go
```diff
@@ -23,23 +23,25 @@ var (
...
+	Index         int           `bson:"index" json:"index" yaml:"index"`
+	Measurement   string        `bson:"measurement" json:"measurement" yaml:"measurement"`
+	CalculatedOn  time.Time     `bson:"calculated_on" json:"calculated_on" yaml:"calculated_on"`
+	Algorithm     AlgorithmInfo `bson:"algorithm" json:"algorithm" yaml:"algorithm"`
+	Triage        TriageInfo    `bson:"triage" json:"triage" yaml:"triage"`
+	PercentChange float64       `bson:"percent_change" json:"percent_change" yaml:"percent_change"`
 }
 
-func CreateChangePoint(index int, measurement string, algorithmName string, algorithmVersion int, algoOptions []AlgorithmOption) ChangePoint {
+func CreateChangePoint(index int, measurement string, algorithmName string, algorithmVersion int, algoOptions []AlgorithmOption, percentChange float64) ChangePoint {
```
julianedwards, line 34: Apologize for being late in the game here, but what is the point of having this `CreateChangePoint` function? Aside from setting `CalculatedOn` to `time.Now()` there is no other setup, nor are there private fields. I think it would be cleaner to create instances of struct directly.

zamj: This function already existed and I was just adding to it. I don't have any hard opinion on it. I can remove it and instantiate the struct directly, unless there was a specific reason for this existing, @tychoish @ancostas ?

ancostas: This is a constructor to ensure `CalculatedOn` and `Triage` are set up in a consistent way.

julianedwards: Seems reasonable, I think its mostly how many parameters passed in that bother me. It's too late anyway to make any significant changes to the model. I will mark as resolved.

units/recalculate_change_points.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"sort"
 
+	"github.com/aclements/go-moremath/stats"
+
```
tychoish, line 9: extra blank line?

PR #210, opened 2020-04-24
MAKE-1273: Implement HistoricalTestData Find method

julianedwards:


model/config.go
```diff
@@ -158,10 +158,11 @@ var (
...
 type BucketConfig struct {
-	AWSKey            string `bson:"aws_key" json:"aws_key" yaml:"aws_key"`
-	AWSSecret         string `bson:"aws_secret" json:"aws_secret" yaml:"aws_secret"`
-	BuildLogsBucket   string `bson:"build_logs_bucket" json:"build_logs_bucket" yaml:"build_logs_bucket"`
-	TestResultsBucket string `bson:"test_results_bucket" json:"test_results_bucket" yaml:"test_results_bucket"`
+	AWSKey                    string `bson:"aws_key" json:"aws_key" yaml:"aws_key"`
+	AWSSecret                 string `bson:"aws_secret" json:"aws_secret" yaml:"aws_secret"`
+	BuildLogsBucket           string `bson:"build_logs_bucket" json:"build_logs_bucket" yaml:"build_logs_bucket"`
+	TestResultsBucket         string `bson:"test_results_bucket" json:"test_results_bucket" yaml:"test_results_bucket"`
+	HistoricalTestStatsBucket string `bson:"historical_test_stats_bucket" json:"historical_test_stats_bucket" yaml:"historical_test_stats_bucket"`
```
tychoish, line 165: do we want these in seperate buckets or in different prefixes within the same bucket?

model/historical_test_data.go
```diff
@@ -57,3 +135,10 @@ func (i *HistoricalTestDataInfo) validate() error {
 
 	return catcher.Resolve()
 }
+
+func (i *HistoricalTestDataInfo) getPath(artifactType PailType) string {
+	if artifactType == PailLocal {
+		return filepath.Join(i.Project, i.Variant, i.TaskName, i.TestName, i.RequestType, fmt.Sprintf("%d", i.Date.Unix()))
```
tychoish, line 141: shouldn't we make both of these strings.Join([]string{<list>}, "/")

model/historical_test_data.go
```diff
@@ -16,25 +26,93 @@ type HistoricalTestData struct {
...
+		ctx,
+		d.env,
+		conf.Bucket.HistoricalTestStatsBucket,
+		"",
+		string(pail.S3PermissionsPrivate),
+		true,
+	)
+	if err != nil {
+		return errors.Wrap(err, "problem creating bucket")
+	}
```
tychoish, line 87: can we factor out bucket setup (and maybe cache config somewhere?) into a method?

model/config.go
julianedwards: I don't know if it matters... Do you have a preference?

model/historical_test_data.go
julianedwards: I used filepath because for windows the slash is the other way

model/config.go
tychoish: probably within one bucket just because bucket creation is a bit onerous

model/historical_test_data.go
tychoish: oh, I see.

model/test_results.go
```diff
@@ -201,22 +203,9 @@ func (t *TestResults) Download(ctx context.Context) ([]TestResult, error) {
...
-		t.env,
-		conf.Bucket.TestResultsBucket,
-		t.Artifact.Prefix,
-		string(pail.S3PermissionsPrivate),
-		false,
-	)
+	bucket, prefix, err := t.getBucket(ctx)
 	if err != nil {
-		return nil, errors.Wrap(err, "problem creating bucket")
+		return nil, err
```
tychoish, line 208: why not wrap this?

model/test_results.go
```diff
@@ -289,6 +278,38 @@ func (t *TestResults) Close(ctx context.Context) error {
...
+		t.bucket,
+		prefix,
+		string(pail.S3PermissionsPrivate),
+		true,
+	)
+	if err != nil {
+		return nil, "", errors.Wrap(err, "problem creating bucket")
+	}
+
+	return bucket, prefix, nil
```
tychoish, line 310: why do we need to return a prefix if its baked into the bucket?

model/historical_test_data.go
```diff
@@ -113,6 +101,31 @@ func (d *HistoricalTestData) Find(ctx context.Context) error {
 	return nil
 }
 
+func (d *HistoricalTestData) getBucket(ctx context.Context) (pail.Bucket, error) {
+	if d.bucket == "" {
+		conf := &CedarConfig{}
```
tychoish, line 106: I wonder if we should cache some of this? Maybe worth opening a ticket.

julianedwards: I cached the bucket, I don't think theres anything else to cache here. Unless you want to globally cache it like in evergreen?

model/test_results.go
julianedwards: I wrap it in the helper func

model/test_results.go
julianedwards: for logging I want to get the right prefix

model/historical_test_data.go
tychoish: We can't cache it globally because of the cycle. we can do some rearranging if need be to cache it in between requests if it becomes an issue.

PR #212, opened 2020-04-28
MAKE-1286: disambiguate perf routes

tychoish:


rest/service.go
```diff
@@ -214,6 +214,7 @@ func (s *Service) addRoutes() {
 	s.app.AddRoute("/admin/users/key").Version(1).Post().Get().Handler(s.fetchUserToken)
 	s.app.AddRoute("/admin/users/certificate").Version(1).Post().Get().Handler(s.fetchUserCert)
 	s.app.AddRoute("/admin/users/certificate/key").Version(1).Post().Get().Handler(s.fetchUserCertKey)
+	s.app.AddRoute("/admin/perf/change_points").Version(1).Post().Wrap(checkUser).RouteHandler(makePerfSignalProcessingRecalculate(s.sc))
```
ancostas, line 217: We can do this. I'm still wondering about two things though:

* Why are redirecting POST to GET? I feel like GET wouldn't be able to handle a request body.
* Why no error when registering conflicting routes?

tychoish: These are good questions, without good answers. 
- The "GET doesn't have request bodies" is a convention, and servers and clients rarely enforce this area of the specification (and ours is not unusual in this regard), and we try and abide by it, but we fail sometimes.
- It's actually quite complicated to enforce this, and we'd have to enforce it at runtime, and all of our routing is provided by a third party library that we don't maintain, and that doesn't return errors for this kind of problem. It's possible that rather than `{id}` we can specify a more specific regular expression that wouldn't match `change_points` so we could keep the current naming, but I think that's less clear. Any validation that we add would have to check all routes after they've been registered and would have to account for potential regular expression and that seems onerous relative to its value.

ancostas: Let's leave the routing conflict notification for now. As you mention, this is something the library either handles or doesn't. I'm just surprised it doesn't.

For the other issue, I'm still not sure I understand why a POST gets redirected to a GET, even if the GET could theoretically service it by diverging from common practice. It seems simpler to just return a 404 if there is no route registered for a POST.

PR #213, opened 2020-04-28
MAKE-1274: Implement HistoricalTestData SaveNew method

julianedwards:


model/historical_test_data.go
```diff
@@ -101,6 +102,37 @@ func (d *HistoricalTestData) Find(ctx context.Context) error {
...
+		return errors.New("cannot save unpopulated historical test data")
+	}
+	if d.env == nil {
+		return errors.New("cannot find with a nil environment")
+	}
+
+	data, err := bson.Marshal(d)
+	if err != nil {
+		return errors.Wrap(err, "problem marshalling historical test data")
+	}
```
tychoish, line 116: I'd do this on line 128? We might also want to wrap this in something that returns a reader? 

Do we expect that we'll have more than one document at a time? Should we have some way to batch this? 

Also does it make sense to use bson or json for the historical rollups?

julianedwards: I thought we decided on bson? Does it matter?

tychoish: I think BSON for the raw data, but maybe JSON for the rollups? eh, bson is fine.

julianedwards: It would be one document at a time, these will be created and updated in amboy jobs as results come in.

PR #214, opened 2020-04-28
MAKE-1276: Implement HistoricalTestData Remove method

julianedwards:


model/historical_test_data.go
```diff
@@ -136,6 +136,21 @@ func (d *HistoricalTestData) SaveNew(ctx context.Context) error {
...
+	if d.env == nil {
+		return errors.New("cannot find with a nil environment")
+	}
+
+	bucket, err := d.getBucket(ctx)
+	if err != nil {
+		return err
+	}
+
+	return errors.Wrap(bucket.Remove(ctx, d.getPath()), "problem removing historical test data file")
```
julianedwards, line 151: This returns an error if the file DNE, should we check if the file already DNE before calling `bucket.Remove` and noop ?

tychoish: I mean, we could also just handle the error and ignore DNE errors without needing to make double the calls.

julianedwards: Problem is the bucket can be backed by any implementation in pail, so it would be complicated to ignore the DNE errors.

tychoish: see how amboy implements duplicate job errors. It's a bit annoying, but not terrible.

PR #218, opened 2020-04-30
MAKE-1217, MAKE-1289: Add filtering by variant, task, test, version, measurement, thread_level to get change points by version API, and test using mdb 4.2

ancostas:


model/change_point_detection.go
```diff
@@ -369,10 +369,9 @@ func ReplaceChangePoints(ctx context.Context, env cedar.Environment, performance
 	return catcher.Resolve()
 }
 
-func GetTotalPagesForChangePointsGroupedByVersion(ctx context.Context, env cedar.Environment, projectId string, pageSize int) (int, error) {
-	countKey := "count"
-	pipe := appendAfterBaseGetChangePointsByVersionAgg(projectId, bson.M{
-		"$count": countKey,
+func GetTotalPagesForChangePointsGroupedByVersion(ctx context.Context, env cedar.Environment, projectId string, pageSize int, variantRegex, versionRegex, taskRegex, testRegex, measurementRegex string, threadLevels []int) (int, error) {
+	pipe := appendAfterBaseGetChangePointsByVersionAgg(projectId, variantRegex, versionRegex, taskRegex, testRegex, measurementRegex, threadLevels, bson.M{
+		"$count": "count",
```
ancostas, line 374: No need for a variable.

model/change_point_detection.go
```diff
@@ -389,14 +388,44 @@ func GetTotalPagesForChangePointsGroupedByVersion(ctx context.Context, env cedar
 		}
 		return int(math.Ceil(float64(res.Count) / float64(pageSize))), nil
 	}
-	return 0, errors.New("Not able to get count of total changepoints matching query")
+	return 0, nil
```
ancostas, line 391: If you don't get any results, I don't think its an error. There are just no results.

model/change_point_detection.go
```diff
@@ -389,14 +388,44 @@ func GetTotalPagesForChangePointsGroupedByVersion(ctx context.Context, env cedar
...
+		bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoVersionKey):  bson.M{"$regex": versionRegex},
+		bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoTaskNameKey): bson.M{"$regex": taskRegex},
+		bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoTestNameKey): bson.M{"$regex": testRegex},
+	}
+
+	if threadLevels != nil {
+		matchStage[bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoArgumentsKey, "thread_level")] = bson.M{
+			"$in": threadLevels,
+		}
+	}
```
ancostas, line 407: Looks like threadlevels will need to be a csv list of integers, rather than a regex, since they are persisted as ints.

rest/change_point_routes.go
```diff
@@ -49,19 +56,33 @@ func (h *perfGetChangePointsByVersionHandler) Parse(_ context.Context, r *http.R
 	} else {
 		h.page = 0
 	}
-	pageSize := vals.Get("page_size")
+	pageSize := vals.Get("pageSize")
```
ancostas, line 59: all the same I think, but usually these query parameters are camel case.

rest/model/change_points.go
```diff
@@ -5,13 +5,13 @@ import dbmodel "github.com/evergreen-ci/cedar/model"
...
 type APIChangePointGroupedByVersionResult struct {
-	Versions   []APIChangePointsWithVersion `json:"versions"`
+	Versions   []APIVersionWithChangePoints `json:"versions"`
 	Page       int                          `json:"page"`
 	PageSize   int                          `json:"page_size"`
 	TotalPages int                          `json:"total_pages"`
 }
 
-type APIChangePointsWithVersion struct {
+type APIVersionWithChangePoints struct {
```
ancostas, line 14: I think this is clearer.

model/change_point_detection.go
```diff
@@ -369,10 +369,9 @@ func ReplaceChangePoints(ctx context.Context, env cedar.Environment, performance
 	return catcher.Resolve()
 }
 
-func GetTotalPagesForChangePointsGroupedByVersion(ctx context.Context, env cedar.Environment, projectId string, pageSize int) (int, error) {
-	countKey := "count"
-	pipe := appendAfterBaseGetChangePointsByVersionAgg(projectId, bson.M{
-		"$count": countKey,
+func GetTotalPagesForChangePointsGroupedByVersion(ctx context.Context, env cedar.Environment, projectId string, pageSize int, variantRegex, versionRegex, taskRegex, testRegex, measurementRegex string, threadLevels []int) (int, error) {
```
julianedwards, line 372: lets use an arg struct here

ancostas: Which package/s should this arg struct reside in?

It seems like this coerces either:
* mapping over structs in each layer (inefficient)
* direct dependency of top level packages on bottom level packages (design antipattern)

julianedwards: The arg struct would reside in the model package. What do you mean by mapping over struct in each layer?

julianedwards: Here's an example of an arg struct: https://github.com/evergreen-ci/cedar/blob/master/rest/data/interface.go#L97

ancostas:
Can we just cut a ticket to fix perf graph lookups? These things break all the time and should probably:
* Work in v4.2 (live in prod)
* Be isolated and repeatable

outside the scope of this ticket...

model/change_point_detection.go
```diff
@@ -389,14 +388,44 @@ func GetTotalPagesForChangePointsGroupedByVersion(ctx context.Context, env cedar
 		}
 		return int(math.Ceil(float64(res.Count) / float64(pageSize))), nil
 	}
-	return 0, errors.New("Not able to get count of total changepoints matching query")
+	return 0, nil
 }
 
-func appendAfterBaseGetChangePointsByVersionAgg(projectId string, additionalSteps ...bson.M) []bson.M {
+func appendAfterBaseGetChangePointsByVersionAgg(projectId, variantRegex, versionRegex, taskRegex, testRegex, measurementRegex string, threadLevels []int, additionalSteps ...bson.M) []bson.M {
+	matchStage := bson.M{
```
julianedwards, line 395: should you be checking if the regexs are empty before adding them to the match stage?

model/change_point_detection.go
ancostas: If the arg struct resides in the model package, then the rest package needs to know about that package, and that defeats the purpose of providing interfaces in the data package that are meant to abstract away the data persistence layer.

ancostas: if you add an arg struct in the data package to allow rest package to be agnostic of the model package, then you will need to map over from one arg struct to another.

model/change_point_detection.go
ancostas: The empty regex does nothing regardless.

model/change_point_detection.go
ancostas: Regarding indexes, there appear to be some in place for project, variant, version, task, test.

julianedwards: The rest/data package already uses the model package so it doesn't matter. You can either map over from one arg struct to another, or you can alias the model's arg struct in the rest/data package so that the rest package is still agnostic to the model package.

ancostas: I'd argue the rest package should be refactored at some point to *not* use the model package directly. In any case, I didn't know about type aliases, and that solves the problem!

model/change_point_detection.go
```diff
@@ -369,10 +369,21 @@ func ReplaceChangePoints(ctx context.Context, env cedar.Environment, performance
 	return catcher.Resolve()
 }
 
-func GetTotalPagesForChangePointsGroupedByVersion(ctx context.Context, env cedar.Environment, projectId string, pageSize int) (int, error) {
-	countKey := "count"
-	pipe := appendAfterBaseGetChangePointsByVersionAgg(projectId, bson.M{
-		"$count": countKey,
+type GetChangePoinsGroupedByVersionArgs struct {
```
julianedwards, line 372: This is misspelled, also I would change it to `ChangePointsByVersionOpts`. (We call these arg structs opts instead of args).

rest/change_point_routes.go
```diff
@@ -38,37 +37,50 @@ func (h *perfGetChangePointsByVersionHandler) Factory() gimlet.RouteHandler {
...
+	h.args.TaskRegex = vals.Get("taskRegex")
+	h.args.TestRegex = vals.Get("testRegex")
+	h.args.MeasurementRegex = vals.Get("measurementRegex")
+	tls := vals.Get("threadLevels")
+	if tls != "" {
+		tlslice := strings.Split(tls, ",")
+		for _, tl := range tlslice {
+			intTl, err := strconv.Atoi(tl)
+			catcher.Add(err)
+			h.args.ThreadLevels = append(h.args.ThreadLevels, intTl)
```
zamj, line 69: I think you need to initialize the ThreadLevels slice before using it

model/change_point_detection.go
```diff
@@ -369,10 +369,21 @@ func ReplaceChangePoints(ctx context.Context, env cedar.Environment, performance
...
+type GetChangePointsGroupedByVersionOpts struct {
+	ProjectId        string
+	Page             int
+	PageSize         int
+	VariantRegex     string
+	VersionRegex     string
+	TaskRegex        string
+	TestRegex        string
+	MeasurementRegex string
+	ThreadLevels     []int
```
tychoish, line 381: I'm a bit confused how thread_level gets here: thread level is a higher-order construct that is specific to (some) server tests, and isn't a first-class attribute of the perf system in Cedar: there's a "test options" map in the info section, and I presume we're just mapping against that, but it makes it unclear what happens if we want to handle other kinds of arguments to tests. 

Maybe making this a map[string][]int named Arguments or Parameters would make more sense here.

model/change_point_detection.go
```diff
@@ -387,16 +398,46 @@ func GetTotalPagesForChangePointsGroupedByVersion(ctx context.Context, env cedar
...
+		bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoProjectKey):  args.ProjectId,
+		bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoVariantKey):  bson.M{"$regex": args.VariantRegex},
+		bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoVersionKey):  bson.M{"$regex": args.VersionRegex},
+		bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoTaskNameKey): bson.M{"$regex": args.TaskRegex},
+		bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoTestNameKey): bson.M{"$regex": args.TestRegex},
+	}
+
+	if args.ThreadLevels != nil {
+		matchStage[bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoArgumentsKey, "thread_level")] = bson.M{
+			"$in": args.ThreadLevels,
```
tychoish, line 417: see above about thread level.

model/change_point_detection.go
```diff
@@ -387,16 +398,46 @@ func GetTotalPagesForChangePointsGroupedByVersion(ctx context.Context, env cedar
...
 
-func appendAfterBaseGetChangePointsByVersionAgg(projectId string, additionalSteps ...bson.M) []bson.M {
+func appendAfterBaseGetChangePointsByVersionAgg(args GetChangePointsGroupedByVersionOpts, additionalSteps ...bson.M) []bson.M {
+	matchStage := bson.M{
+		bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoProjectKey):  args.ProjectId,
+		bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoVariantKey):  bson.M{"$regex": args.VariantRegex},
+		bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoVersionKey):  bson.M{"$regex": args.VersionRegex},
+		bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoTaskNameKey): bson.M{"$regex": args.TaskRegex},
+		bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoTestNameKey): bson.M{"$regex": args.TestRegex},
+	}
```
tychoish, line 413: do we have anything to assert that these regexes are valid or anchored? are these queries well indexed?

model/change_point_detection.go
```diff
@@ -369,10 +369,21 @@ func ReplaceChangePoints(ctx context.Context, env cedar.Environment, performance
 	return catcher.Resolve()
 }
 
-func GetTotalPagesForChangePointsGroupedByVersion(ctx context.Context, env cedar.Environment, projectId string, pageSize int) (int, error) {
-	countKey := "count"
-	pipe := appendAfterBaseGetChangePointsByVersionAgg(projectId, bson.M{
-		"$count": countKey,
+type GetChangePointsGroupedByVersionOpts struct {
```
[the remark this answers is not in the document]
ancostas: If Opts are usually persisted, should we perhaps just do Opts/Args. IMO this would make clear that these are just arguments for an API.

model/change_point_detection.go
```diff
@@ -369,10 +369,21 @@ func ReplaceChangePoints(ctx context.Context, env cedar.Environment, performance
...
+	ProjectId        string
```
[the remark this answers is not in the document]
ancostas: GAH!

model/change_point_detection.go
ancostas: How it got here is simple: consumer wants to filter/sort by thread level! So I'd request that the customer facing API support `?thread_levels=2,4,16` for example, but how this is represented after parsing and in lower levels is up to you. Let me know what your prefer!

model/change_point_detection.go
ancostas: We have no requirements regarding the regex as of yet. One hopes that the ~10 users of this API who work for us won't be too naughty. Are there certain cases that worry you specifically that we'd want to handle here right now? All that comes to mind based on my brief reading of $regex in mdb is that you wouldn't want case insensitive regexes or such, but we're not allowing user to pass options anyway.

W.r.t. indices, last I checked we have indices for project id, variant, version, task, and test.

model/change_point_detection.go
ancostas: I meant for the rest API --> we wanted to pass a regex initially but since they are persisted as integers we instead should just pass a slice of ints to match agains (which is a csv list of ints when invoking the API).

model/change_point_detection.go
ancostas: Ditto. We want to filter on them, but how we do it/the representation is up for review.

model/change_point_detection.go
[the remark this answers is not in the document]
tychoish: If we're persisting this structure to the db (which I don't follow, because it seems just like an argument structure passed between different internal APIs, but maybe I'm reading it wrong) then it needs bson tags, and the like

model/change_point_detection.go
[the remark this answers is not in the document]
tychoish: sorry :(

model/change_point_detection.go
tychoish: Right but, tests and users can specify arbirary string-to-int (slice) mapping and the back-end doesn't need to know about them internally.

model/change_point_detection.go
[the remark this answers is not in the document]
ancostas: No I mean: we're not persisting to DB -- would renaming Opts-->Args so that its clear this is just a struct to pass arguments, rather than some kind of options on a domain model?

model/change_point_detection.go
ancostas: i.e. `?arguments=<insert json here>`?

model/change_point_detection.go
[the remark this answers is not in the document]
tychoish: I think if you name it "Options" its clear enough. I think having a panic'ing marshaler is a good way to ensure it doesn't get stuffed in the DB, but that's optional.

model/change_point_detection.go
tychoish: Sorry for being unclear earlier. I'm not objecting to filtering it, just the wat that this implementation institutionalizes thread_level over other kinds of arguments/parameters on tests.

model/change_point_detection.go
tychoish: Hope springs eternal! Also thanks for checking on the indexes. 

I believe if you do not anchor the regexes (e.g. have them start with ^) then the queries can't use the indexes,) so if we don't validate for that in the application then we could end up impacting the DB with a query, which seems not-ideal.

model/change_point_detection.go
tychoish: Is it viable to just parse unknown query strings as things that go into arguments? 

json encoding query parameters seems... fun?

ancostas: Tix to handle this stuff since ui also needs an update https://jira.mongodb.org/browse/MAKE-1299

model/change_point_detection.go
ancostas: Just did a quick check here: https://docs.mongodb.com/manual/reference/operator/query/regex/

Seems like as long as it's not case insensitive, it will leverage an index; Left anchoring/prefixes allow for an additional optimization of only traversing the range of that index with such a prefix.

ancostas: I think it's well enough that it uses the index without requiring a prefix match -- e.g. its probably a common use case to search for "latency" and want to find things like "average_latency_90th_percentile".

PR #220, opened 2020-04-30
MAKE-1242: Create test results protobuf

julianedwards:


test_results.proto
```diff
@@ -0,0 +1,61 @@
...
+  string version = 2;
+  string variant = 3;
+  string task_name = 4;
+  string task_id = 5;
+  int32 execution = 6;
+  string request_type = 7;
+  bool mainline = 8;
+}
+
+enum TestResultsStorage {
```
tychoish, line 25: does it makes sense to allow this to be user specifiable?

julianedwards: You're right, it doesn't seem to make sense. We allow it to be user specifiable for logs as well which doesn't seem right either.

julianedwards: But the client of this would be evergreen, so do we want to take that control away?

PR #221, opened 2020-05-01
MAKE-1243: Implement gRPC service CreateTestResultsRecord

julianedwards:


rpc/internal/test_results.ext.go
```diff
@@ -0,0 +1,50 @@
...
+		TaskID:      t.TaskId,
+		Execution:   int(t.Execution),
+		RequestType: t.RequestType,
+		Mainline:    t.Mainline,
+	}
+}
+
+// Export exports TestResult to the corresponding TestResult type in the model
+// package.
+func (t TestResult) Export() (model.TestResult, error) {
```
tychoish, line 26: should this return a pointer?

rpc/internal/test_results_service.go
```diff
@@ -0,0 +1,58 @@
...
+
+	record := model.CreateTestResults(info.Export(), conf.Bucket.TestResultsBucketType)
+	record.Setup(s.env)
+	return &TestResultsResponse{TestResultsRecordId: record.ID},
+		newRPCError(codes.Internal, errors.Wrap(record.SaveNew(ctx), "problem saving test results record"))
+}
+
+// AddTestResults adds test results to an existing test results record.
+func (s *testResultsService) AddTestResults(ctx context.Context, results *TestResults) (*TestResultsResponse, error) {
+	return nil, nil
```
tychoish, line 44: can we return errors till tehy're implemented?

rpc/internal/test_results.ext.go
julianedwards: No because the model functions don't take pointers

rpc/internal/test_results_service.go
julianedwards: sure

PR #222, opened 2020-05-01
MAKE-1243: Implement test results grpc AddTestResults, StreamTestResults, and CloseTestResultsRecord

julianedwards:


rpc/internal/test_results_service.go
```diff
@@ -41,18 +43,73 @@ func (s *testResultsService) CreateTestResultsRecord(ctx context.Context, info *
...
+	for i, result := range results.Results {
+		exportedResult, err := result.Export()
+		if err != nil {
+			return nil, newRPCError(codes.InvalidArgument, errors.Wrapf(err, "problem exporting test result"))
+		}
+		exportedResults[i] = exportedResult
+	}
+
+	return &TestResultsResponse{TestResultsRecordId: record.ID},
+		newRPCError(codes.Internal, errors.Wrapf(record.Append(ctx, exportedResults), "problem appending test results for '%s'", results.TestResultsRecordId))
```
tychoish, line 65: Is it correct to return a valid response and an error?

rpc/internal/test_results_service.go
```diff
@@ -41,18 +43,73 @@ func (s *testResultsService) CreateTestResultsRecord(ctx context.Context, info *
...
 }
 
 // StreamTestResults adds test results via client-side streaming to an existing
 // test results record.
 func (s *testResultsService) StreamTestResults(stream CedarTestResults_StreamTestResultsServer) error {
-	return newRPCError(codes.Unimplemented, errors.New("not implemented"))
+	ctx := stream.Context()
+	id := ""
```
tychoish, line 72: I'd `var id string` for zero-value initializations.

rpc/internal/test_results_service.go
```diff
@@ -41,18 +43,73 @@ func (s *testResultsService) CreateTestResultsRecord(ctx context.Context, info *
...
+		if err := ctx.Err(); err != nil {
+			return newRPCError(codes.Aborted, err)
+		}
+
+		results, err := stream.Recv()
+		if err == io.EOF {
+			return stream.SendAndClose(&TestResultsResponse{TestResultsRecordId: id})
+		}
+		if err != nil {
+			return err
```
tychoish, line 84: should we wrap?

rpc/internal/test_results_service.go
```diff
@@ -41,18 +43,73 @@ func (s *testResultsService) CreateTestResultsRecord(ctx context.Context, info *
...
+	id := ""
+
+	for {
+		if err := ctx.Err(); err != nil {
+			return newRPCError(codes.Aborted, err)
+		}
+
+		results, err := stream.Recv()
+		if err == io.EOF {
+			return stream.SendAndClose(&TestResultsResponse{TestResultsRecordId: id})
```
tychoish, line 81: what if id is still uninitialized.

rpc/internal/test_results_service.go
```diff
@@ -41,18 +43,73 @@ func (s *testResultsService) CreateTestResultsRecord(ctx context.Context, info *
...
+	record.Setup(s.env)
+	if err := record.Find(ctx); err != nil {
+		if db.ResultsNotFound(err) {
+			return nil, newRPCError(codes.NotFound, err)
+		}
+		return nil, newRPCError(codes.Internal, errors.Wrapf(err, "problem finding test results record for '%s'", info.TestResultsRecordId))
+	}
+
+	return &TestResultsResponse{TestResultsRecordId: record.ID},
+		newRPCError(codes.Internal, errors.Wrapf(record.Close(ctx), "problem closing test results with id %s", record.ID))
```
tychoish, line 114: same question about non-nil+error

rpc/internal/test_results_service.go
julianedwards: Interesting point... well it wouldn't be an error technically. Just means the user closed it before sending any information.

rpc/internal/test_results_service.go
julianedwards: the stream.Recv() func returns an rpc error code, should we still wrap it?

rpc/internal/test_results_service.go
```diff
@@ -110,6 +114,8 @@ func (s *testResultsService) CloseTestResultsRecord(ctx context.Context, info *T
 		return nil, newRPCError(codes.Internal, errors.Wrapf(err, "problem finding test results record for '%s'", info.TestResultsRecordId))
 	}
 
-	return &TestResultsResponse{TestResultsRecordId: record.ID},
-		newRPCError(codes.Internal, errors.Wrapf(record.Close(ctx), "problem closing test results with id %s", record.ID))
+	if err := record.Close(ctx); err != nil {
```
tychoish, line 117: if the context is canceled do we want to not close the record?

PR #223, opened 2020-05-04
MAKE-1299: Refactor expanded metrics front+back end not to rely on thread levels directly

ancostas:


ancostas:
Failing tests are accounted for with other tix from what i can see.

rest/change_point_routes.go
```diff
@@ -42,36 +43,56 @@ func (h *perfGetChangePointsByVersionHandler) Parse(_ context.Context, r *http.R
...
 	h.args.MeasurementRegex = vals.Get("measurementRegex")
-	tls := vals.Get("threadLevels")
-	if tls != "" {
-		tlslice := strings.Split(tls, ",")
-		for _, tl := range tlslice {
-			intTl, err := strconv.Atoi(tl)
-			catcher.Add(err)
-			h.args.ThreadLevels = append(h.args.ThreadLevels, intTl)
+	delete(vals, "measurementRegex")
+	h.args.Arguments = map[string][]int{}
```
zamj, line 71: I have a similar sentiment here as my other comment. It worries me to re-type these, it seems an easy place to accidentally break in the future. What would you think about declaring these queryParams as constants or something like that so they could be re-used?

rest/change_point_routes.go
```diff
@@ -35,43 +36,73 @@ func (h *perfGetChangePointsByVersionHandler) Factory() gimlet.RouteHandler {
...
+				catcher.Add(err)
+				h.args.Arguments[key] = append(h.args.Arguments[key], intVal)
+			}
 		}
 	}
 	return catcher.Resolve()
 }
 
+var matchFirstCap = regexp.MustCompile("(.)([A-Z][a-z]+)")
+var matchAllCap = regexp.MustCompile("([a-z0-9])([A-Z])")
```
tychoish, line 98: I'd avoid the global variables and just put these in the utility function!

rest/change_point_routes_test.go
```diff
@@ -0,0 +1,56 @@
...
+func (s *ChangePointSuite) TestGetVersionWithChangePointsRouteParsing() {
+	rh := s.rh["change_points_by_version"].(*perfGetChangePointsByVersionHandler)
+	url := "/perf/project/some_project/change_points_by_version"
+	url += "?page=0&pageSize=10"
+	url += "&variantRegex=some_variant"
+	url += "&versionRegex=some_version"
+	url += "&taskRegex=some_task"
+	url += "&testRegex=some_test"
+	url += "&measurementRegex=some_measurement"
+	url += "&threadLevel=1,2,3"
```
tychoish, line 41: I generally do strings.Join() for things like this but it's totally fine.

rest/change_point_routes.go
tychoish: (I might also make it return an error if there's a problem parsing the regex or the conversion?)

PR #225, opened 2020-05-05
MAKE-1304: Get version with change points not working

ancostas:


model/change_point_detection.go
```diff
@@ -464,6 +464,15 @@ func GetChangePointsGroupedByVersion(ctx context.Context, env cedar.Environment,
...
 		},
+		// It is sad, that $group can't handle $skips and $limits :(
+		{
+			"$lookup": bson.M{
+				"from":         perfResultCollection,
+				"localField":   "perf_results",
+				"foreignField": "_id",
+				"as":           "perf_results",
+			},
+		},
```
ancostas, line 475: @tychoish @zamj I feel dirty, but it works. A self join in a nosql database.... 😬

model/change_point_detection.go
```diff
@@ -464,8 +464,17 @@ func GetChangePointsGroupedByVersion(ctx context.Context, env cedar.Environment,
 		{
 			"$limit": args.PageSize,
 		},
+		// It is sad, that $group can't handle $skips and $limits :(
```
tychoish, line 467: take this out. 

aggregation works like a pipeline, so might be the fact that the skip/limit happens after $group? is there a way to project more of the document in the group to get out of this?

ancostas:
@jimoleary Have you encountered the need to self join before?

model/change_point_detection.go
ancostas: projecting more of the documents inside the $group is what doesn't scale (since you have to group over the whole collection).

ancostas: IMO agg pipeline could have an optimization such that $group takes $skip and $limit downstream into account. But I digress, will remove the comment.

model/change_point_detection.go
zamj: I think we need to rethink how we're dealing with these change points. If this works, I'm fine for the short term to let us keep moving, but I'd hate to keep this long term.

jimoleary: The $group looks as efficient as it can be  (see [here](https://docs.mongodb.com/master/core/aggregation-pipeline/#pipeline-operators-and-indexes)) although I am suspicious of whether an index can be used after it.

Skip and limit for pagination is pretty much always inefficient in mongodb (since the document sizes vary you can't reallly skip).

This pipeline always seems to sort by ```order``` and then ```skip```, would it be possible to insert an order $lte or $gte into the initial match stage in appendAfterBaseGetChangePointsByVersionAgg (with a supporting index)?

This should jump directly to the page you are looking for and might allow you to remove the skip stage and maybe the sort stage (depending on the location of order in the index). In this case you only ever process one page worth of documents.

I also suspect that you might be able to remove the group stage altogether and merge it into the lookup (but you would need to filter the older results afterwards to avoid a pipeline and keep it efficient, plus I'm not sure if appendAfterBaseGetChangePointsByVersionAgg is used elsewhere).

**However, has anyone looked at an explain plan for this pipeline or the logs?**

@ancostas: Self joins aren't unkown but I think in principle this is not really a self join since you are looking up a different type of document that just happens to be stored in the same collection. The lookup should be efficient as it must be indexed and is not a pipeline but are there any known issues with 1 to many joins? You could roll them up **but** this may lead to other issues.

ancostas: Regarding using `$gte`/`$lte` for the pagination, wouldn't you need to first `$group` by version, then `$skip` and `$limit`, in order to find the page bounds in terms of `order`? Also, I think we might be over-optimizing when getting into these kinds of bucketing patterns: users in the UI can only go back and forth between pages, so I don't see them ever getting to page 50 of 500 (They would just refine their sort+filter before that).

ancostas: Also, since you asked:
Query
```
[
        {
                "$match" : {
                        "info.project" : "sys-perf",
                        "analysis.change_points.0" : {
                                "$exists" : true
                        }
                }
        },
        {
                "$group" : {
                        "_id" : "$info.version",
                        "order" : {
                                "$first" : "$info.order"
                        },
                        "perf_results" : {
                                "$push" : "$_id"
                        }
                }
        },
        {
                "$lookup" : {
                        "from" : "perf_results",
                        "localField" : "_id",
                        "foreignField" : "info.version",
                        "as" : "perf_results"
                }
        }
]
```

Explain
```
{
        "stages" : [
                {
                        "$cursor" : {
                                "query" : {
                                        "info.project" : "sys-perf",
                                        "analysis.change_points.0" : {
                                                "$exists" : true
                                        }
                                },
                                "fields" : {
                                        "info.order" : 1,
                                        "info.version" : 1,
                                        "_id" : 1
                                },
                                "queryPlanner" : {
                                        "plannerVersion" : 1,
                                        "namespace" : "cedar.perf_results",
                                        "indexFilterSet" : false,
                                        "parsedQuery" : {
                                                "$and" : [
                                                        {
                                                                "info.project" : {
                                                                        "$eq" : "sys-perf"
                                                                }
                                                        },
                                                        {
                                                                "analysis.change_points.0" : {
                                                                        "$exists" : true
                                                                }
                                                        }
                                                ]
                                        },
                                        "queryHash" : "492DF875",
                                        "planCacheKey" : "2C5BD197",
                                        "winningPlan" : {
                                                "stage" : "FETCH",
                                                "filter" : {
                                                        "analysis.change_points.0" : {
                                                                "$exists" : true
                                                        }
                                                },
                                                "inputStage" : {
                                                        "stage" : "IXSCAN",
                                                        "keyPattern" : {
                                                                "info.project" : 1
                                                        },
                                                        "indexName" : "info.project_1",
                                                        "isMultiKey" : false,
                                                        "multiKeyPaths" : {
                                                                "info.project" : [ ]
                                                        },
                                                        "isUnique" : false,
                                                        "isSparse" : false,
                                                        "isPartial" : false,
                                                        "indexVersion" : 2,
                                                        "direction" : "forward",
                                                        "indexBounds" : {
                                                                "info.project" : [
                                                                        "[\"sys-perf\", \"sys-perf\"]"
                                                                ]
                                                        }
                                                }
                                        },
                                        "rejectedPlans" : [
                                                {
                                                        "stage" : "FETCH",
                                                        "filter" : {
                                                                "analysis.change_points.0" : {
                                                                        "$exists" : true
                                                                }
                                                        },
                                                        "inputStage" : {
                                                                "stage" : "IXSCAN",
                                                                "keyPattern" : {
                                                                        "info.project" : NumberLong(1),
                                                                        "info.variant" : NumberLong(1),
                                                                        "info.task_name" : NumberLong(1),
                                                                        "info.test_name" : NumberLong(1),
                                                                        "info.mainline" : NumberLong(1),
                                                                        "info.order" : NumberLong(1)
                                                                },
                                                                "indexName" : "info.project_1_info.variant_1_info.task_name_1_info.test_name_1_info.mainline_1_info.order_1",
                                                                "isMultiKey" : false,
                                                                "multiKeyPaths" : {
                                                                        "info.project" : [ ],
                                                                        "info.variant" : [ ],
                                                                        "info.task_name" : [ ],
                                                                        "info.test_name" : [ ],
                                                                        "info.mainline" : [ ],
                                                                        "info.order" : [ ]
                                                                },
                                                                "isUnique" : false,
                                                                "isSparse" : false,
                                                                "isPartial" : false,
                                                                "indexVersion" : 2,
                                                                "direction" : "forward",
                                                                "indexBounds" : {
                                                                        "info.project" : [
                                                                                "[\"sys-perf\", \"sys-perf\"]"
                                                                        ],
                                                                        "info.variant" : [
                                                                                "[MinKey, MaxKey]"
                                                                        ],
                                                                        "info.task_name" : [
                                                                                "[MinKey, MaxKey]"
                                                                        ],
                                                                        "info.test_name" : [
                                                                                "[MinKey, MaxKey]"
                                                                        ],
                                                                        "info.mainline" : [
                                                                                "[MinKey, MaxKey]"
                                                                        ],
                                                                        "info.order" : [
                                                                                "[MinKey, MaxKey]"
                                                                        ]
                                                                }
                                                        }
                                                }
                                        ]
                                }
                        }
                },
                {
                        "$group" : {
                                "_id" : "$info.version",
                                "order" : {
                                        "$first" : "$info.order"
                                },
                                "perf_results" : {
                                        "$push" : "$_id"
                                }
                        }
                },
                {
                        "$lookup" : {
                                "from" : "perf_results",
                                "as" : "perf_results",
                                "localField" : "_id",
                                "foreignField" : "info.version"
                        }
                }
        ],
        "serverInfo" : {
                "host" : "alex-costas-linux-desktop",
                "port" : 27017,
                "version" : "4.2.6",
                "gitVersion" : "20364840b8f1af16917e4c23c1b5f5efd8b352f8"
        },
        "ok" : 1
}
```

Issue isn't so much the performance though, its more that $group scans the entire db, which exhaust the 100MB memory limit for that stage if you are `$push`ing `$$ROOT` into the groups...

PR #227, opened 2020-05-18
MAKE-1334: Add Exhausted check to LogIterator interface

julianedwards:


model/log_iterator.go
```diff
@@ -23,9 +23,12 @@ import (
...
 // logs.
 type LogIterator interface {
-	// Next returns true if the iterator has not yet been exhausted, false
-	// otherwise.
+	// Next returns true if the iterator has not yet been exhausted or
+	// closed, false otherwise.
 	Next(context.Context) bool
+	// Exhausted returns true if the iterator has not yet been exhausted,
+	// regardless if it has been closed or not.
+	Exhausted() bool
```
tychoish, line 31: I think I'd call this HasNext(), which seems more consistent with iterator idioms I've seen other places.

julianedwards: I think it should not be called `HasNext` because it is different in that `HasNext` should return whether `Next` is going to return `true`, whereas this just returns that you reached the end of the iterator. This is necessary because `Next` can return `false` before the iterator is exhausted in the case where the iterator is closed.

PR #230, opened 2020-06-01
MAKE-1331: Make signal processing jobs in cedar report time series to Performance Analysis and Triage Service and remove all other tooling

zamj:


zamj:
This is the first of two PR's. The next one will remove all the change point related things this one didn't

units/time_series_update.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+		j.performanceAnalysisService = perf.NewPerformanceAnalysisService(j.conf.ChangeDetector.URI, j.conf.ChangeDetector.User, j.conf.ChangeDetector.Token)
+	}
+	performanceData, err := model.GetPerformanceData(ctx, j.env, j.PerformanceResultId)
+	if err != nil {
+		j.AddError(errors.Wrapf(err, "Unable to aggregate time perfData %s", j.PerformanceResultId.String()))
+		return
+	}
+	if performanceData == nil {
+		j.AddError(model.MarkPerformanceResultsAsAnalyzed(ctx, j.env, j.PerformanceResultId))
```
ancostas, line 94: I think we may want to mark as analyzed below -- i.e. once we've successfully reported the update -- as well. This should prevent the same time series being picked up by the periodic job over and over, right?

We may also consider whether we would like to remove that periodic job and this marking altogether in the subsequent PR.

ancostas:
Also, some of the tests relating to our other stuff have broken with these changes; I'd say just delete those tests and do the rest of the cleanup in subsequent PR.

units/time_series_update.go
zamj: Makes sense, I've marked them as analyzed. Sure, we can discuss that in the second PR.

zamj:
> Looks pretty good, barring the one comment relating to marking series as analyzed.
> 
> A couple things outside the scope of the PR:
> 
> 1. Have we tried this out against staging?
> 2. Do we need any DevOps instructions for cedar? I think not, since the baseurl and credentials for the service _should_ not be different. Point 1. should provide a sanity check for that.

Yeah, I don't think we need to do anything devops related. Staging is working with this change

zamj:
The failing test in this patch is going to be worked on in
https://jira.mongodb.org/browse/MAKE-1290
and doesn't seem related to the changes in this PR.

julianedwards:
Can we make a ticket to add documentation to exported fields?

zamj:
> Can we make a ticket to add documentation to exported fields?

Done
https://jira.mongodb.org/browse/MAKE-1359

julianedwards:
Thanks!

perf/time_series_updater.go
```diff
@@ -0,0 +1,133 @@
...
+	return nil
+}
+
+func (spc *performanceAnalysisAndTriageClient) doRequest(method, route string, ctx context.Context, in interface{}) error {
+	body, err := json.Marshal(in)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	conf := utility.HTTPRetryConfiguration{
```
julianedwards, line 76: I would use the default HTTP config, and just add/change the things you want to configure.

zamj: Most of this file is the same as `perf/change_point_detector.go` was. For whatever reason, my `git mv` didn't take so it looks like all the new code. That part is the same as the old though. I'm a little hesitant to change the config just because we had issues with networking in the past and seems to be working well now.

julianedwards: sounds fair

PR #239, opened 2020-06-26
EVG-9700: Fix execution filtering and add testing for buildlogger

julianedwards:
These changes:
- Fix the model and rest code to filter execution correctly (if execution is not set, the latest execution is fetched).
- Adds testing for the above change.
- Cleans up some of the code.

[EVG-9700](https://jira.mongodb.org/browse/EVG-9700)

### DIFF
--- a/model/buildlogger.go
+++ b/model/buildlogger.go
@@ -397,29 +397,13 @@ type Logs struct {
 	timeRange TimeRange
 }
 
-// EmptyLogInfo allows querying of null or missing fields.
-type EmptyLogInfo struct {
-	Project     bool
-	Version     bool
-	Variant     bool
-	TaskName    bool
-	TaskID      bool
-	Execution   bool
-	TestName    bool
-	Trial       bool
-	ProcessName bool
-	Format      bool
-	Tags        bool
-	Arguments   bool
-	ExitCode    bool
-}
-
 // LogFindOptions describes the search criteria for the Find function on Logs.
 type LogFindOptions struct {
-	TimeRange TimeRange
-	Info      LogInfo
-	Empty     EmptyLogInfo
-	Limit     int64
+	TimeRange       TimeRange
+	Info            LogInfo
+	EmptyTestName   bool
+	LatestExecution bool
+	Limit           int64
 }
 
 // Setup sets the environment for the logs. The environment is required for
@@ -445,7 +429,10 @@ func (l *Logs) Find(ctx context.Context, opts LogFindOptions) error {
 	if opts.Limit > 0 {
 		findOpts.SetLimit(opts.Limit)
 	}
-	findOpts.SetSort(bson.D{{Key: logCreatedAtKey, Value: -1}})
+	findOpts.SetSort(bson.D{
+		{Key: logInfoExecutionKey, Value: -1},
+		{Key: logCreatedAtKey, Value: -1},
+	})
 	it, err := l.env.GetDB().Collection(buildloggerCollection).Find(ctx, createFindQuery(opts), findOpts)
 	if err != nil {
 		return errors.WithStack(err)
@@ -464,8 +451,23 @@ func (l *Logs) Find(ctx context.Context, opts LogFindOptions) error {
 	} else {
 		return mongo.ErrNoDocuments
 	}
+	if opts.LatestExecution {
+		l.filterLatestExecution()
+	}
 
 	return nil
+}
+
+func (l *Logs) filterLatestExecution() {
+	execution := 0
+	for i, log := range l.Logs {
+		if log.Info.Execution < execution {
+			l.Logs = l.Logs[:i]
+			break
+		} else {
+			execution = log.Info.Execution
+		}
+	}
 }
 
 func createFindQuery(opts LogFindOptions) map[string]interface{} {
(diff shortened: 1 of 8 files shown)


model/buildlogger.go
```diff
@@ -445,7 +429,10 @@ func (l *Logs) Find(ctx context.Context, opts LogFindOptions) error {
 	if opts.Limit > 0 {
 		findOpts.SetLimit(opts.Limit)
 	}
-	findOpts.SetSort(bson.D{{Key: logCreatedAtKey, Value: -1}})
+	findOpts.SetSort(bson.D{
```
Kimchelly, line 432: Is this sort covered by an index?

model/buildlogger.go
```diff
@@ -464,87 +451,75 @@ func (l *Logs) Find(ctx context.Context, opts LogFindOptions) error {
...
 		return mongo.ErrNoDocuments
 	}
+	if opts.LatestExecution {
+		l.filterLatestExecution()
+	}
 
 	return nil
 }
 
+func (l *Logs) filterLatestExecution() {
```
Kimchelly, line 461: I don't follow the logic behind this. Is this trying to just get the single latest result?

If the results are already returned in order of decreasing execution ID, isn't it sufficient to just set `l.Logs` to the first element? And if they're not sorted by execution ID, don't you have to traverse the entire slice?

model/buildlogger.go
```diff
@@ -464,87 +451,75 @@ func (l *Logs) Find(ctx context.Context, opts LogFindOptions) error {
...
+}
+
 func createFindQuery(opts LogFindOptions) map[string]interface{} {
 	search := bson.M{
 		logCreatedAtKey:   bson.M{"$lte": opts.TimeRange.EndAt},
 		logCompletedAtKey: bson.M{"$gte": opts.TimeRange.StartAt},
 	}
 	if opts.Info.Project != "" {
 		search[bsonutil.GetDottedKeyName(logInfoKey, logInfoProjectKey)] = opts.Info.Project
-	} else if opts.Empty.Project {
```
Kimchelly, line 478: Are all of these empty field options no longer necessary (aside from test name)?

model/buildlogger_test.go
```diff
@@ -593,10 +593,21 @@ func TestBuildloggerFindLogs(t *testing.T) {
 		assert.NoError(t, db.Collection(buildloggerCollection).Drop(ctx))
 	}()
 	log1, log2 := getTestLogs()
-
-	_, err := db.Collection(buildloggerCollection).InsertOne(ctx, log1)
-	require.NoError(t, err)
-	_, err = db.Collection(buildloggerCollection).InsertOne(ctx, log2)
+	time.Sleep(time.Millisecond)
```
Kimchelly, line 596: Rather than relying on sleeping to change `time.Now()`, would it be better to just pass in a time to `getTestLogs()`?

model/buildlogger_test.go
```diff
@@ -759,48 +789,45 @@ func TestBuildloggerCreateFindQuery(t *testing.T) {
...
+		assert.Equal(t, opts.Info.Format, search[bsonutil.GetDottedKeyName(logInfoKey, logInfoFormatKey)])
+		assert.Equal(t, bson.M{"$in": opts.Info.Tags}, search[bsonutil.GetDottedKeyName(logInfoKey, logInfoTagsKey)])
+		query, ok := search[bsonutil.GetDottedKeyName(logInfoKey, logInfoArgumentsKey)]
+		require.True(t, ok)
+		args := query.(bson.M)["$in"].([]bson.M)
+		for _, arg := range args {
+			count := 0
+			for key, val := range arg {
+				assert.Equal(t, opts.Info.Arguments[key], val)
+				assert.Zero(t, count)
```
Kimchelly, line 812: If you want the loop to have exactly 0 or 1 element, can you just require that `len(args) < 2` rather than have the counter expect at most one iteration?

rest/data/buildlogger.go
```diff
@@ -36,7 +36,7 @@ func (dbc *DBConnector) FindLogByID(ctx context.Context, opts BuildloggerOptions
 	} else if err != nil {
 		return data, time.Time{}, paginated, gimlet.ErrorResponse{
 			StatusCode: http.StatusInternalServerError,
-			Message:    fmt.Sprintf("database error"),
+			Message:    errors.Wrapf(err, "problem finding log with id '%s'", opts.ID).Error(),
```
Kimchelly, line 39: Yay! 🎉

rest/data/buildlogger.go
```diff
@@ -256,17 +258,18 @@ func (dbc *DBConnector) FindGroupedLogs(ctx context.Context, opts BuildloggerOpt
...
-	it, err := dbc.findLogsByTestName(ctx, opts, false)
+	it, err := dbc.findLogsByTestName(ctx, &opts)
 	if err != nil {
 		return data, time.Time{}, paginated, err
 	}
 	its = append(its, it)
 
 	opts.TestName = ""
-	it, err = dbc.findLogsByTestName(ctx, opts, false)
+	// need to set this to false since the last call to findLogsByTestaName
```
Kimchelly, line 269: findLogsByTestName*

Also capitalize "needs" since it's a sentence.

rest/data/buildlogger.go
```diff
@@ -568,17 +584,18 @@ func (mc *MockConnector) FindGroupedLogs(ctx context.Context, opts BuildloggerOp
...
-	it, err := mc.findLogsByTestName(ctx, opts, false)
+	it, err := mc.findLogsByTestName(ctx, &opts)
 	if err != nil {
 		return data, time.Time{}, paginated, err
 	}
 	its = append(its, it)
 
 	opts.TestName = ""
-	it, err = mc.findLogsByTestName(ctx, opts, false)
+	// need to set this to false since the last call to findLogsByTestaName
```
Kimchelly, line 595: Same comment as above.

model/buildlogger.go
julianedwards: So there can be more than one log document for an execution (actually there almost always is). For example, evergreen task logs, agent logs, and system logs that all have the same task id + execution but would be separate buildlogger metadata docs.

model/buildlogger.go
julianedwards: No that's why I removed them

model/buildlogger.go
julianedwards: No I need to update the indexes

julianedwards: No I need to change the indexes

PR #244, opened 2020-07-01
EVG-9700: Correctly filter groups

julianedwards:
This ticket is part of [EVG-9700](https://jira.mongodb.org/browse/EVG-9700). It corrects the filtering by "group", which is reality just a tag in the log metadata but needs to be filtered separately in cases where there is a group and tags as the `$in` filter in mongo returns an array with any of the matching fields (rather than all).

### DIFF
--- a/model/buildlogger.go
+++ b/model/buildlogger.go
@@ -519,6 +519,8 @@ func createFindQuery(opts LogFindOptions) map[string]interface{} {
 			bson.M{bsonutil.GetDottedKeyName(logInfoKey, logInfoTagsKey): opts.Group},
 			bson.M{bsonutil.GetDottedKeyName(logInfoKey, logInfoTagsKey): bson.M{"$in": opts.Info.Tags}},
 		}
+	} else if opts.Group != "" {
+		search[bsonutil.GetDottedKeyName(logInfoKey, logInfoTagsKey)] = opts.Group
 	} else if len(opts.Info.Tags) > 0 {
 		search[bsonutil.GetDottedKeyName(logInfoKey, logInfoTagsKey)] = bson.M{"$in": opts.Info.Tags}
 	}
--- a/model/buildlogger.go
+++ b/model/buildlogger.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/sha1"
+
 	"fmt"
 	"hash"
 	"io"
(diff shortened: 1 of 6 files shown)


model/buildlogger.go
```diff
@@ -511,7 +532,9 @@ func createFindQuery(opts LogFindOptions) map[string]interface{} {
 	if opts.Info.Format != "" {
 		search[bsonutil.GetDottedKeyName(logInfoKey, logInfoFormatKey)] = opts.Info.Format
 	}
-	if len(opts.Info.Tags) > 0 {
+	if opts.Group != "" {
+		search[bsonutil.GetDottedKeyName(logInfoKey, logInfoTagsKey)] = bson.M{"$in": []string{opts.Group}}
```
Kimchelly, line 536: Isn't is sufficient to omit `$in` for the case of a singular string? I think `search[bsonutil.GetDottedKeyName(logInfoKey, logInfoTagsKey)] = opts.Group` is equivalent.

model/buildlogger.go
```diff
@@ -425,15 +426,34 @@ func (l *Logs) Find(ctx context.Context, opts LogFindOptions) error {
...
+	query := createFindQuery(opts)
+	sort := bson.D{
 		{Key: logInfoExecutionKey, Value: -1},
 		{Key: logCreatedAtKey, Value: -1},
-	})
-	it, err := l.env.GetDB().Collection(buildloggerCollection).Find(ctx, createFindQuery(opts), findOpts)
+	}
+	if opts.Group != "" && len(opts.Info.Tags) > 0 {
+		pipeline := []bson.M{
+			bson.M{"$match": query},
```
Kimchelly, line 439: nit: you don't need to declare all the `bson.M` types when initializing the slice elements since the compiler already knows they're of type `bson.M` (e.g. this could just be `{"$match": query}`).

model/buildlogger.go
```diff
@@ -425,15 +426,34 @@ func (l *Logs) Find(ctx context.Context, opts LogFindOptions) error {
...
+			bson.M{"$match": bson.M{
```
Kimchelly, line 440: Is there a reason `$match` has to be done twice in a row rather than just `$match` once? Can this find logic just be handled in `createFindQuery` by using `$and` on `logInfoTagsKey` for the tags + the group?

rest/data/buildlogger.go
```diff
@@ -188,6 +188,7 @@ func (dbc *DBConnector) FindLogsByTestName(ctx context.Context, opts Buildlogger
 		paginated bool
 	)
 
+	opts.Group = ""
```
Kimchelly, line 191: What's the significance of clearing the group for the data connectors?

julianedwards: Just in case it was set, it shouldn't be set here. Sanity thing

Kimchelly: Sure, but why is it invalid for it to be set here?

julianedwards: cause the test name doesn't allow filtering by group (there is a special group route for that)

PR #248, opened 2020-07-01
EVG-12442: fix various bugs in REST client

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-12442

* Fix bugs that made the REST client unusable from the CLI.
* Fix misspellings.

### DIFF
--- a/operations/client.go
+++ b/operations/client.go
@@ -60,7 +60,7 @@ func printStatus() cli.Command {
 			}
 
 			grip.Debug(status)
-			out, err := pretyJSON(status)
+			out, err := prettyJSON(status)
 			if err != nil {
 				return errors.WithStack(err)
 			}
@@ -71,7 +71,7 @@ func printStatus() cli.Command {
 	}
 }
 
-func pretyJSON(data interface{}) (string, error) {
+func prettyJSON(data interface{}) (string, error) {
 	out, err := json.MarshalIndent(data, "", "   ")
 	if err != nil {
 		return "", errors.Wrap(err, "problem rendering status result")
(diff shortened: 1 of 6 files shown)


operations/client.go
```diff
@@ -47,7 +47,7 @@ func printStatus() cli.Command {
 			opts := rest.ClientOptions{
 				Host:   c.Parent().String(clientHostFlag),
 				Port:   c.Parent().Int(clientPortFlag),
-				Prefix: "",
+				Prefix: "/rest",
```
julianedwards, line 50: Should this be `rest/v1`? I think it would make more sense to put the `v1` here instead of in all the rest routes in `rest/client.go` but that would be an annoying change so feel free to ignore.

operations/flags.go
```diff
@@ -100,7 +100,6 @@ func restServiceFlags(flags ...cli.Flag) []cli.Flag {
 		cli.IntFlag{
 			Name:  clientPortFlag,
 			Usage: "port for the remote cedar service.",
-			Value: 3000,
```
julianedwards, line 103: why remove this?

julianedwards, line 103: why remove this?

operations/client.go
julianedwards: On second thought, some routes may want to use a future v2 and others v1. This is fine.

operations/flags.go
Kimchelly: This is mostly a problem of what a sane default looks like. For a localhost service, it's fine, but if you want to use it as a client to production cedar, you have to specify `--host https://cedar.mongodb.com --port 0`, which is kind of weird.

julianedwards: Oh I thought remote used port 3000 for some reason...

julianedwards: Does the value not have to be set to 0 then?

Kimchelly: I changed the `BeforeFunc` a bit to handle default vs non-default behavior. It sets the port to 3000 if you don't specify the host flag (so using the default values will still connect to the default local cedar). However, if you set host (e.g. `--host=https://cedar.mongodb.com`), the port will not be set to any default (so you don't need to specify `--port=0`).

PR #270, opened 2020-08-12
EVG-12752: Add FindByTaskID method to system metrics model

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-12752

This is the first of a series of tickets to add the rest route to fetch system metrics data. This ticket:
1.) Adds a `FindByTaskID` method on the `SystemMetrics` struct in the `model` package.
2.) Cleans up some tests and comments.

### DIFF
--- a/model/system_metrics_test.go
+++ b/model/system_metrics_test.go
@@ -81,6 +81,80 @@ func TestSystemMetricsFind(t *testing.T) {
 	})
 }
 
+func TestSystemMetricsFindByTaskID(t *testing.T) {
+	env := cedar.GetEnvironment()
+	db := env.GetDB()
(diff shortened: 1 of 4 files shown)


model/system_metrics_test.go
```diff
@@ -81,6 +81,80 @@ func TestSystemMetricsFind(t *testing.T) {
 	})
 }
 
+func TestSystemMetricsFindByTaskID(t *testing.T) {
+	env := cedar.GetEnvironment()
+	db := env.GetDB()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	defer func() {
```
Kimchelly, line 89: Should this also clear the collection at the start of the test to ensure it has no data to begin with?

PR #272, opened 2020-08-14
EVG-12752: Add data fetching for system metrics in rest package 

julianedwards:


rest/data/system_metrics.go
```diff
@@ -0,0 +1,100 @@
...
+			Message:    fmt.Sprintf("system metrics for task id '%s' not found", opts.TaskID),
+		}
+	} else if err != nil {
+		return nil, gimlet.ErrorResponse{
+			StatusCode: http.StatusInternalServerError,
+			Message:    errors.Wrapf(err, "problem retrieving system metrics data for task id '%s'", opts.TaskID).Error(),
+		}
+	}
+
+	// check that the metric is valid so we can return the appropriate err
```
Kimchelly, line 34: nit: error, not err

rest/data/system_metrics.go
```diff
@@ -0,0 +1,100 @@
...
+	// code.
+	_, ok := sm.Artifact.MetricChunks[metricType]
+	if !ok {
+		return nil, gimlet.ErrorResponse{
+			StatusCode: http.StatusNotFound,
+			Message:    fmt.Sprintf("metric type '%s' for task id '%s' not found", metricType, opts.TaskID),
+		}
+	}
+	sm.Setup(dbc.env)
```
Kimchelly, line 43: Is there a reason we need to call `Setup` again when it was already called before this?

rest/data/system_metrics.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+///////////////////////////////
+// MockConnector Implementation
+///////////////////////////////
+
+func (mc *MockConnector) FindSystemMetricsByType(ctx context.Context, metricType string, opts dbModel.SystemMetricsFindOptions) (io.ReadCloser, error) {
+	var sm *dbModel.SystemMetrics
+	for key := range mc.CachedSystemMetrics {
+		val := mc.CachedSystemMetrics[key]
```
Kimchelly, line 62: Can this just iterate `for key, val := range mc.CachedSystemMetrics`?

rest/data/system_metrics.go
```diff
@@ -0,0 +1,100 @@
...
+		if opts.EmptyExecution {
+			if val.Info.TaskID == opts.TaskID && (sm == nil || val.Info.Execution > sm.Info.Execution) {
+				sm = &val
+			}
+		} else {
+			if val.Info.TaskID == opts.TaskID && val.Info.Execution == opts.Execution {
```
Kimchelly, line 68: Can this just be combined with the `else` block to be `else if`?

rest/data/system_metrics.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+	}
+	if sm == nil {
+		return nil, gimlet.ErrorResponse{
+			StatusCode: http.StatusNotFound,
+			Message:    fmt.Sprintf("system metrics for task id '%s' not found", opts.TaskID),
+		}
+	}
+
+	// check that the metric is valid so we can return the appropriate err
```
Kimchelly, line 81: nit: error, not err.

rest/data/system_metrics.go
julianedwards: If we use `key, val...` I can't use a pointer to `val` to assign it to `sm`. Iterating using `range` in go is weird.

PR #277, opened 2020-08-27
EVG-12820: Add pagination to system metrics download

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-12820

This is the first of 2 PRs for this ticket. In this PR I changed the system metrics model code to have a separate `DownloadWithPagination` method and enabled a soft size limit in the `SystemMetricsReadCloser`.

### DIFF
--- a/model/system_metrics.go
+++ b/model/system_metrics.go
@@ -326,8 +326,61 @@ func (sm *SystemMetrics) appendSystemMetricsChunkKey(ctx context.Context, metric
 // Download returns a system metrics reader for the system metrics data of the
 // specified type.
 func (sm *SystemMetrics) Download(ctx context.Context, metricType string) (io.ReadCloser, error) {
+	bucket, chunks, err := sm.download(ctx, metricType)
+	if err != nil {
+		return nil, err
+	}
+
+	return NewSystemMetricsReadCloser(ctx, SystemMetricsReadCloserOptions{
+		Bucket:    bucket,
+		Chunks:    chunks,
+		BatchSize: 2,
+	}), nil
+}
+
+// SystemMetricsDownloadOptions contains the arguments for downloading system
+// metrics with pagination.
+type SystemMetricsDownloadOptions struct {
+	MetricType string
+	PageSize   int
+	StartIndex int
+}
+
+// DownloadWithPagination downloads system metrics with a soft size limit. The
+// data along with the next unread chunk index is returned, chunks are never
+// partially read.
+func (sm *SystemMetrics) DownloadWithPagination(ctx context.Context, opts SystemMetricsDownloadOptions) ([]byte, int, error) {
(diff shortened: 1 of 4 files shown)


model/system_metrics_test.go
```diff
@@ -571,6 +576,57 @@ func TestSystemMetricsDownload(t *testing.T) {
...
+		}
+		systemMetrics2.populated = true
+		systemMetrics2.Setup(env)
+
+		readData, idx, err := systemMetrics2.DownloadWithPagination(ctx, SystemMetricsDownloadOptions{
+			MetricType: "pagination",
+			PageSize:   33,
+		})
+		require.NoError(t, err)
+		assert.Equal(t, fullData[:64], readData)
```
Kimchelly, line 607: Can we add a comment saying that the 64 comes from expecting two 32-byte chunk strings to be read?

model/test_utils.go
```diff
@@ -57,8 +57,8 @@ func GenerateTestLog(ctx context.Context, bucket pail.Bucket, size, chunkSize in
 	return chunks, lines, nil
 }
 
-// GenerateSystemMetrics is a convenience function to generate a specified number of
-// 16 byte random system metrics data chunks in the specified bucket.
+// GenerateSystemMetrics is a convenience function to generate a specified
+// number of 32 byte random system metrics data chunks in the specified bucket.
```
Kimchelly, line 61: Thank you, this would be confusing otherwise.

model/system_metrics_test.go
Kimchelly: Actually thinking on it, it might just have been clearer if `PageSize` was calculated as `len(chunk) + 1` and then this would check `fullData[:2 * len(chunk)]` (without needing a comment).

model/system_metrics.go
```diff
@@ -326,8 +326,61 @@ func (sm *SystemMetrics) appendSystemMetricsChunkKey(ctx context.Context, metric
...
+// data along with the next unread chunk index is returned, chunks are never
+// partially read.
+func (sm *SystemMetrics) DownloadWithPagination(ctx context.Context, opts SystemMetricsDownloadOptions) ([]byte, int, error) {
+	bucket, chunks, err := sm.download(ctx, opts.MetricType)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	if opts.StartIndex >= len(chunks.Chunks) {
+		return nil, opts.StartIndex, nil
```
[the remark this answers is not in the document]
julianedwards: I think typically with pagination, you return the next index no matter what. It is up to the reader of the paginated data to check that no more data is being returned and to stop reading.

model/test_utils.go
julianedwards: Yes! I was stuck for like 2 hours testing cause I though it was 16 bytes and I couldn't figure out why the pagination wasn't working! It was working, I was just off with the size of the chunks...

PR #278, opened 2020-08-27
EVG-12721: add debug logs for system metrics

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-12721

* I don't see any obvious causes of problems when testing system metrics, so I want to debug what requests are made to the gRPC service.
* Also drive-by improvement to makefile's linting.

### DIFF
--- a/rpc/internal/system_metrics_service.go
+++ b/rpc/internal/system_metrics_service.go
@@ -8,6 +8,8 @@ import (
 	"github.com/evergreen-ci/cedar"
 	"github.com/evergreen-ci/cedar/model"
 	"github.com/mongodb/anser/db"
+	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/message"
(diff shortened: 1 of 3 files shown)


rpc/internal/system_metrics_service.go
```diff
@@ -48,14 +50,30 @@ func (s *systemMetricsService) AddSystemMetrics(ctx context.Context, data *Syste
 	systemMetrics := &model.SystemMetrics{ID: data.Id}
 	systemMetrics.Setup(s.env)
 	if err := systemMetrics.Find(ctx); err != nil {
+		grip.Error(message.WrapError(err, message.Fields{
```
julianedwards, line 53: maybe add id here? or task id?

Kimchelly: Done. The task ID isn't available here, so I just used the document ID.

PR #279, opened 2020-08-31
EVG-12820: Add pagination for system metrics rest routes

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-12820

This adds pagination for system metrics to the rest package.

### DIFF
--- a/rest/data/system_metrics_test.go
+++ b/rest/data/system_metrics_test.go
@@ -91,7 +91,8 @@ func (s *systemMetricsConnectorSuite) setup() {
 		systemMetrics := dbModel.CreateSystemMetrics(taskInfo, opts)
 		systemMetrics.Setup(s.env)
 		s.Require().NoError(systemMetrics.SaveNew(s.ctx))
-		s.Require().NoError(systemMetrics.Append(s.ctx, "uptime", dbModel.FileText, []byte(fmt.Sprintf("execution %d", taskInfo.Execution))))
+		s.Require().NoError(systemMetrics.Append(s.ctx, "uptime", dbModel.FileText, []byte(fmt.Sprintf("execution %d\n", taskInfo.Execution))))
+		s.Require().NoError(systemMetrics.Append(s.ctx, "uptime", dbModel.FileText, []byte("chunk2")))
 		s.Require().NoError(systemMetrics.Find(s.ctx))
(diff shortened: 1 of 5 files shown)


rest/system_metrics_routes.go
```diff
@@ -40,16 +43,23 @@ func (h *systemMetricsGetByTypeHandler) Factory() gimlet.RouteHandler {
 func (h *systemMetricsGetByTypeHandler) Parse(_ context.Context, r *http.Request) error {
 	var err error
 
-	h.opts.TaskID = gimlet.GetVars(r)["task_id"]
-	h.metricType = gimlet.GetVars(r)["type"]
+	h.findOpts.TaskID = gimlet.GetVars(r)["task_id"]
+	h.downloadOpts.MetricType = gimlet.GetVars(r)["type"]
 	vals := r.URL.Query()
+	h.downloadOpts.PageSize = softSizeLimit
```
Kimchelly, line 49: Should users be able to specify this in their request?

rest/data/system_metrics_test.go
```diff
@@ -105,41 +106,65 @@ func (s *systemMetricsConnectorSuite) TearDownSuite() {
...
+	data, nextIdx, err = s.sc.FindSystemMetricsByType(s.ctx, findOpts, downloadOpts)
+	s.Require().NoError(err)
+	s.Equal("chunk2", string(data))
+	s.Equal(2, nextIdx)
+
+	downloadOpts.StartIndex = 2
+	data, nextIdx, err = s.sc.FindSystemMetricsByType(s.ctx, findOpts, downloadOpts)
+	s.Require().NoError(err)
+	s.Nil(data)
+	s.Equal(2, nextIdx)
```
Kimchelly, line 144: Should this also test the case where the page size is large enough to read multiple chunks (but not all of them at once)?

rest/system_metrics_routes.go
julianedwards: I don't think so, I don't want to trust users to set reasonable page sizes. We already set it for buildlogger and it seems to work fine.

rest/data/system_metrics_test.go
julianedwards: That is tested in the model package, I just want to test that it is calling the right thing.

PR #282, opened 2020-09-22
EVG-12996: fix test results REST routes

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-12996

The bucket used for the test result iterator should not include the test results collection name in the prefix. I also added a check for errors after calling `(*TestResultsIterator).Next()`.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -6,7 +6,6 @@ linters:
     - goimports
     - govet
     - ineffassign
-    - interfacer
     - misspell
     - unconvert
 
--- a/model/test_results.go
+++ b/model/test_results.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"hash"
 	"io"
-	"path/filepath"
 	"time"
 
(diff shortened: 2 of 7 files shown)


julianedwards:
Having the prefix with the test results collection is intentional. We are going to use the same bucket for historical stats data and thus need the collection prefixed so we can keep them separate.

julianedwards (changes requested):
The usage of the collection prefix is intentional, the same bucket will be used for historical stats data so we want to keep things separate. What we can do is set the full prefix in  `model.CreateTestResults`.

Kimchelly:
It seems kind of weird to have the historical stats stored with test results. Why can't we just have separate buckets?

Kimchelly:
Update: I made [BUILD-11915](https://jira.mongodb.org/browse/BUILD-11915) to request a separate bucket for historical test results.

julianedwards (approved):

PR #289, opened 2020-10-05
EVG-8914: implement historical test data job

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-8914

* Implement historical test data job.
    * It doesn't retry, so it's possible for some test results data points to be lost.
* Use historical test data bucket and remove unnecessary bucket prefix.
* Do not include unnecesary fields in historical test data in S3.
* Store time.Durations for durations instead of floats (results from the REST route are still returned as floating point seconds instead of nanoseconds).


units/historical_test_data.go
```diff
@@ -0,0 +1,206 @@
...
+	if err := conf.Find(); err != nil {
+		j.AddError(errors.Wrap(err, "finding cedar configuration"))
+		return
+	}
+	if conf.Flags.DisableHistoricalTestData {
+		return
+	}
+
+	skip, err := j.shouldNoop(ctx, conf)
+	j.AddError(errors.Wrapf(err, "checking if job needs to run"))
```
julianedwards, line 100: We want to continue on error here?

units/historical_test_data.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+}
+
+// shouldNoop checks if this job must run for this project and task, depending
+// on whether historical test data is disabled for this job's project and
+// whether this project skips tasks matching this job's task name.
+func (j *historicalTestDataJob) shouldNoop(ctx context.Context, conf *model.CedarConfig) (bool, error) {
+	settings, err := j.getProjectSettings(ctx, conf)
+	if err != nil {
+		return false, errors.Wrapf(err, "getting historical test data settings for project '%s'", j.Info.Project)
```
julianedwards, line 145: If we rely on skip in the `Run` shouldn't this be set to `true`? Unless we want to continue on error in this case.

julianedwards (changes requested):

units/historical_test_data.go
Kimchelly: Yeah, `skip` is what actually determines if the job should continue.

units/historical_test_data.go
Kimchelly: I set this to false because I think if we can't get the project settings (e.g. network failures, Evergreen is down), in the worst case, we'll end up including test results in the historical stats on an excluded task or a disabled project. it seems like it's safer to still record the test data rather than potentially drop a data point that we actually want.

julianedwards (approved):

PR #291, opened 2020-10-14
EVG-13135: log error if required indexes are missing

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-13135

* Add function to check that indexes exist (ignoring extra options such as "unique"/"sparse").
* Log error when indexes are missing. I'll add a splunk alert for missing indexes.

### DIFF
--- a/operations/service.go
+++ b/operations/service.go
@@ -165,6 +165,10 @@ func Service() cli.Command {
 				return errors.WithStack(err)
 			}
 
+			if err := model.CheckIndexes(ctx, env.GetDB(), model.GetRequiredIndexes()); err != nil {
+				grip.Error(errors.Wrap(err, "missing expected database indexes"))
+			}
+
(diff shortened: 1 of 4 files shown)


operations/service.go
```diff
@@ -165,6 +165,11 @@ func Service() cli.Command {
 				return errors.WithStack(err)
 			}
 
+			if err := model.CheckIndexes(ctx, env.GetDB(), model.GetRequiredIndexes()); err != nil {
```
julianedwards, line 168: Can we have an option to skip this? Running cedar locally for testing would fail since your local DB probably doesn't have these indexes.

julianedwards (changes requested):

operations/service.go
Kimchelly: Done. I moved the check for the DB URI.

julianedwards (approved):

PR #293, opened 2020-11-12
EVG-7782: Return 404 when test result DNE

julianedwards:


rest/test_results_routes.go
```diff
@@ -15,7 +15,7 @@ import (
 
 ///////////////////////////////////////////////////////////////////////////////
 //
-// GET /testresults/task_id/{task_id}
+// GET /test_results/task_id/{task_id}
```
Kimchelly, line 18: Doesn't the REST service route also have to be changed?

Kimchelly (changes requested):

rest/test_results_routes.go
julianedwards: I am going to change it in another ticket, I don't want to break too many things right now.

Kimchelly: I thought nothing was using the test results routes right now. What's wrong with changing it?

julianedwards: Staging, but I will change it before I merge staging. It just requires me to also change timber and revendor and I am trying to debug something else before that.

Kimchelly (approved):

PR #295, opened 2020-12-01
EVG-13301: Merge rollups when saving new perf data

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-13301

There are two ways to add rollups: (1) via `PerformanceResult.SaveNew` and (2) via `PerformanceResult.MergeRollups`. The latter ensures no duplicate rollup names are added and checks against the version of a rollup (if it exists already). This PR refactors `PerformanceResult.SaveNew` to use `MergeRollups` instead of inserting the rollups as they are.

### DIFF
--- a/model/perf.go
+++ b/model/perf.go
@@ -139,6 +139,9 @@ func (result *PerformanceResult) SaveNew(ctx context.Context) error {
 		result.ID = result.Info.ID()
 	}
 
+	rollups := result.Rollups.Stats
+	result.Rollups.Stats = []PerfRollupValue{}
+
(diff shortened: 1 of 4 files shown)


ancostas (changes requested):
Looks good modulo: can we add a test showing it does error when user attempts to save duplicate rollups?

Kimchelly (approved):

julianedwards:
Yeah lint is failing on master, I am going to look into it separately. So the design of `MergeRollups` isn't that it will fail but it instead just take the most recent rollup with the given name (hence, the version check).

ancostas (approved):
Ok

PR #298, opened 2020-12-08
EVG-13411: Rough POC of what the cedar subscription service may look like

julianedwards:
This is for https://jira.mongodb.org/browse/PM-2004. Let me know what you think!

Essentially, there is a collection for each "topic" registered in the `topicRegistry` in `model`. Internally, cedar can create and publish whatever topics are necessary. The topic collection really only holds a foreign id, pointing to the relevant document in another collection (maybe this should be a list of ids in the event a topic needs to fetch multiple ids?). The `Topic` interface would then need to implement the `ExtractMessage` function which would take that id and look it up in the database and return the relevant data for that topic (for example, rollups for perf metrics).


model/message_broker.go
```diff
@@ -0,0 +1,135 @@
...
+	data := make(chan MessageEntry, 100)
+	go func() {
+		defer close(data)
+
+		for cs.Next(ctx) {
+			resumeToken = []byte(cs.ResumeToken())
+			entry := MessageEntry{}
+			entry.ResumeToken = resumeToken
+
+			if err = cs.Decode(cs); err != nil {
```
Kimchelly, line 88: Is this supposed to decode into the entry?

model/message_broker.go
```diff
@@ -0,0 +1,135 @@
...
+				entry.Err = errors.Wrapf(err, "problem decoding message entry for topic %s", topic.Name())
+				data <- entry
+				break
+			}
+			data <- entry
+		}
+
+		catcher := grip.NewBasicCatcher()
+		catcher.Add(cs.Err())
+		catcher.Add(cs.Close(nil))
```
Kimchelly, line 98: Can we pass a context to this, even if it's just `context.Background()`?

model/message_broker.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// ResumeToken and Error should be checked when watching a topic.
+	ResumeToken []byte
+	Err         error
+}
+
+// PublishToTopic creates a new message entry with the given id in the
+// corresponding collection for the given topic. If the topic does not exist in
+// the global registry, an error is returned.
+func PublishToTopic(ctx context.Context, env cedar.Environment, name string, id string) error {
```
Kimchelly, line 29: I guess it's premature since it isn't called anywhere, but how are we going to know when to publish to a topic?

Kimchelly (changes requested):

model/message_broker.go
julianedwards: The cursor automatically sets `context.Background()` if its nil is passed in.

model/message_broker.go
julianedwards: When people ask us to. We would have to create a new implementation of the `Topic` interface and then publish wherever the new information is coming in (probably in the rpc service code).

model/message_broker.go
julianedwards: Yes, thank you!

model/message_broker.go
```diff
@@ -0,0 +1,135 @@
...
+		defer close(data)
+
+		for cs.Next(ctx) {
+			resumeToken = []byte(cs.ResumeToken())
+			entry := MessageEntry{}
+			entry.ResumeToken = resumeToken
+
+			if err = cs.Decode(cs); err != nil {
+				entry.Err = errors.Wrapf(err, "problem decoding message entry for topic %s", topic.Name())
+				data <- entry
```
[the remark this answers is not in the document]
julianedwards: `cs.Next` checks for `ctx.Done()` so it will return `false` when that happens.

model/message_broker.go
Kimchelly: I know it does, but the context docs specifically state not to do this: https://golang.org/pkg/context/

Do not pass a nil Context, even if a function permits it.

model/message_broker.go
```diff
@@ -0,0 +1,135 @@
...
+	return data, nil
+}
+
+// Topic is an interface representing a topic for publishing and subscribing
+// via the message broker.
+type Topic interface {
+	// Name returns the unique name of the topic.
+	Name() string
+	// ExtractMessage searches the database and returns a consistently
+	// structured message extracted from the document with the given id.
```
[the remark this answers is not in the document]
julianedwards: Like a topic will have a consistent structure for each message. Maybe uniform is a better word?

model/message_broker.go
```diff
@@ -0,0 +1,135 @@
...
+		opts = opts.SetResumeAfter(resumeToken)
+	}
+	cs, err := env.GetDB().Collection(topic.Name()).Watch(ctx, mongo.Pipeline{}, opts)
+	if err != nil {
+		return nil, errors.Wrapf(err, "problem getting change streams for topic %s", topic.Name())
+	}
+
+	// TODO: figure out appropriate buffer size.
+	data := make(chan MessageEntry, 100)
+	go func() {
```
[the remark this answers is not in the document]
julianedwards: Yes

model/message_broker.go
```diff
@@ -0,0 +1,135 @@
...
+
+// MessageEntry represents an entry for a "message" of a topic in the database.
+// ForeignID is the id of a document located in another collection used by the
+// Topic interface to extract the actual message. The id of each message entry
+// is an int sequentially incremented for each message published to the topic.
+type MessageEntry struct {
+	ForeignID string `bson:"foreign_id"`
+
+	// ResumeToken and Error should be checked when watching a topic.
+	ResumeToken []byte
```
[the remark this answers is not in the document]
julianedwards: Nope, they don't get saved to the DB. They are for the clients when watching a topic.

model/message_broker.go
```diff
@@ -0,0 +1,135 @@
...
+
+	return watchTopic(ctx, env, topic, resumeToken)
+}
+
+// publishToTopic creates a new message entry with the foreign ID in the
+// corresponding collection for the given topic.
+func publishToTopic(ctx context.Context, env cedar.Environment, topic Topic, foreignID string) error {
+	entry := &MessageEntry{ForeignID: foreignID}
+	// TODO: should there be a collection per topic? Or, rather, one
+	// collection for all topics, indexed by topic names.
```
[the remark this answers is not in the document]
julianedwards: Yeah, then we can just index by topic name. I will change that.

model/message_broker.go
```diff
@@ -0,0 +1,135 @@
...
+		"collection":   topic.Name(),
+		"insertResult": insertResult,
+		"op":           "publish to a topic",
+	})
+
+	return errors.Wrapf(err, "problem publishing to topic %s", topic.Name())
+}
+
+func watchTopic(ctx context.Context, env cedar.Environment, topic Topic, resumeToken []byte) (chan MessageEntry, error) {
+	// TODO: Maybe need to set options FullDocument -> options.UpdateLookup.
```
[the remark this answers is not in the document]
julianedwards: There won't be, these documents should never change. We just add new ones. I actually don't think we need to do this.

model/message_broker.go
julianedwards: Haha ok I will change it.

model/message_broker.go
Kimchelly: Okay, I was just checking that every time we have to publish something for a new topic, we have to change the code. Instead of putting it in the gRPC code, isn't there a concern that this is fragile or should be put in the DB operation code? It basically means that anytime we do a database op, we'll have to check if the change is supposed to be published to some topic.

rpc/internal/subscriber_service.go
```diff
@@ -0,0 +1,61 @@
...
+			msg, err = topic.ExtractMessage(msgEntry.ForeignID)
+			if err != nil {
+				msgEntry.Err = errors.Wrap(err, "problem extracting message")
+			}
+		}
+
+		err = srv.Send(&SubscribeResponse{
+			Topic:       req.Topic,
+			ResumeToken: resumeToken,
+			Msg:         msg,
```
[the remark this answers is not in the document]
julianedwards: I imagine we would convert it to JSON for the user, this would happen in `ExtractMessage`.

rpc/internal/subscriber_service.go
```diff
@@ -0,0 +1,61 @@
...
+
+func (s *subscriberService) Subscribe(req *SubscribeRequest, srv Subscriber_SubscribeServer) error {
+	ctx := srv.Context()
+
+	topic, err := model.GetTopic(req.Topic)
+	if err != nil {
+		return newRPCError(codes.NotFound, err)
+	}
+
+	data, err := model.WatchTopic(ctx, s.env, req.Topic, req.ResumeToken)
```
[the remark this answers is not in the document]
julianedwards: Not really, I am going to look into that.

model/message_broker.go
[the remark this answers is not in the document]
Kimchelly: It does, but we still have to ensure that data won't hang when it sends to the channel. If the loop successfully gets the next change and then the context cancels, then the outer function (in this case `Subscribe()`) will return with an error when sending the message to the client because the context was cancelled. The data stream won't close until it hits the deferred close, so the goroutine will be stuck sending to a channel that will never receive.

model/message_broker.go
julianedwards: Yeah it would be a code change, but hopefully with this interface not too hard or complicated. The `Topic` implementation would be put in the `model` packge and the `Publish` call would be in the `rpc` package (most likely, or maybe in an amboy job who knows).

julianedwards: But if we are publishing something that may be updated or changed, then yes the call to `Publish` would probably make more sense in the database code.

Kimchelly: Alright, I'll leave it alone until I see what the publishing changes look like.

model/message_broker.go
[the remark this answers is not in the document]
Kimchelly: So is the "consistent" part referring to the data format of the message (e.g. the message is always BSON)?

model/message_broker.go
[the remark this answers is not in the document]
Kimchelly: Do we need `omitempty` or does BSON assume they're omitted by default?

model/message_broker.go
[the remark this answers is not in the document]
julianedwards: Ah right, changing it now.

model/message_broker.go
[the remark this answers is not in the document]
julianedwards: Yeah and the layout of the data, so you can expect the same fields.

model/message_broker.go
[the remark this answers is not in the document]
julianedwards: I forgot, I have to do `bson:"-"` for these.

model/message_broker.go
```diff
@@ -0,0 +1,164 @@
...
+// publishToTopic creates a new message entry with the foreign ID in the
+// corresponding collection for the given topic.
+func publishToTopic(ctx context.Context, env cedar.Environment, topic Topic, foreignID string) error {
+	entry := &MessageEntry{
+		Topic:     topic.Name(),
+		ForeignID: foreignID,
+	}
+	insertResult, err := env.GetDB().Collection(topicsCollection).InsertOne(ctx, entry)
+	grip.DebugWhen(err == nil, message.Fields{
+		"collection":   topic.Name(),
```
Kimchelly, line 74: Should this be `topic` instead of `collection`?

Kimchelly (changes requested):

model/message_broker.go
```diff
@@ -0,0 +1,164 @@
...
+	go func() {
+		defer recovery.LogStackTraceAndContinue(fmt.Sprintf("watching topic %s", topic.Name()))
+		defer close(data)
+
+		for cs.Next(ctx) {
+			resumeToken = []byte(cs.ResumeToken())
+			entry := MessageEntry{}
+			entry.ResumeToken = resumeToken
+
+			if err = cs.Decode(&entry); err != nil {
```
ybrill, line 111: I don't understand how this works: I thought Decode would try to unmarshal into an event document.

rpc/internal/subscriber_service.go
ybrill, line 32: is there a solution for the case where the resume token is no longer in the oplog? Like if I subscribed and then went down for a week, is there a way for me to catch up?

model/message_broker.go
```diff
@@ -0,0 +1,164 @@
...
+
+const topicsCollection = "topics"
+
+// MessageEntry represents an entry for a "message" of a topic in the database.
+type MessageEntry struct {
+	// Topic is the name of the topic.
+	Topic string `bson:"topic"`
+	// ForeignID is the id of a document located in another collection used
+	// by the Topic interface to extract the actual message. The id of each
+	// message entry is an int sequentially incremented for each message
```
ybrill, line 26: where do we set the ID of the message entries? Why do we want an increasing int id?

model/message_broker.go
```diff
@@ -0,0 +1,164 @@
...
+func watchTopic(ctx context.Context, env cedar.Environment, topic Topic, resumeToken []byte) (chan MessageEntry, error) {
+	opts := options.ChangeStream()
+	if len(resumeToken) > 0 {
+		opts = opts.SetResumeAfter(resumeToken)
+	}
+	cs, err := env.GetDB().Collection(topicsCollection).Watch(
+		ctx,
+		mongo.Pipeline{{{
+			Key: "$match", Value: bson.D{{
+				Key: messageEntryTopicKey, Value: topic.Name()},
```
ybrill, line 91: [the docs imply](https://docs.mongodb.com/manual/changeStreams/#modify-change-stream-output) the key needs to be `"fullDocument." + messageEntryTopicKey`

model/message_broker.go
julianedwards: That is an outdated comment, going to update it.

model/message_broker.go
julianedwards: What do you mean by an event document?

model/message_broker.go
```diff
@@ -0,0 +1,164 @@
...
+					break
+				}
+			case <-ctx.Done():
+				break
+			}
+		}
+
+		catcher := grip.NewBasicCatcher()
+		catcher.Add(cs.Err())
+		catcher.Add(cs.Close(context.Background()))
```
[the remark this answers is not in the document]
julianedwards: I would rather keep it here just in case there is an error.

model/message_broker.go
ybrill: To me [the docs](https://docs.mongodb.com/manual/changeStreams/#open-a-change-stream) imply that iterating the change stream cursor returns [a change event document](https://docs.mongodb.com/manual/reference/change-events/#change-stream-output).

> To retrieve the data change event from the cursor, iterate the change stream cursor. For information on the change stream event, see Change Events.

rpc/internal/subscriber_service.go
julianedwards: Once the resume token is no longer on the oplog, resuming from that token would fail.

model/message_broker.go
julianedwards: Oh right, yeah I would need to get the full document here.

model/message_broker.go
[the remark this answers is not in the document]
ybrill: what do you think about calling it in both places? [The docs](https://godoc.org/go.mongodb.org/mongo-driver/mongo#ChangeStream.Close) assert the call is idempotent.

rpc/internal/subscriber_service.go
ybrill: I was thinking a solution could be for messages to have an increasing int id (as the old comment implied). Instead of round tripping the resume token with the client, the client could tell us the latest _message number_ they received. We'd query the db for all the messages with a greater id and write them into the stream before watching the change stream.

julianedwards: Yeah I originally wrote exactly that. The code got really ugly and messy with multiple cursors, so for the sake of the POC I made it simpler. I think it is a valid option to think about though.

Kimchelly (approved):

bsamek:
Need this really be served by a change stream? If a reader has read the last place it has read to, then it's possible for that reader to re-query the application and get what is available since the last time it read.

PR #299, opened 2020-12-16
EVG-13447: Move historical test data to DB

julianedwards:

As discussed, this ticket moves storing historical test data in S3 to the database. I removed the durations array since we can calculate the average duration without it.


model/historical_test_data.go
```diff
@@ -1,39 +1,51 @@
...
 // range.
 type HistoricalTestData struct {
-	Info            HistoricalTestDataInfo `bson:"-"`
+	ID              string                 `bson:"_id,omitempty"`
+	Info            HistoricalTestDataInfo `bson:"info"`
 	NumPass         int                    `bson:"num_pass"`
 	NumFail         int                    `bson:"num_fail"`
-	Durations       []time.Duration        `bson:"durations"`
-	AverageDuration time.Duration          `bson:"average_duration"`
+	AverageDuration float64                `bson:"average_duration"`
```
Kimchelly, line 30: Would it be too hard to store it as a `time.Duration` instead so we know it's nanoseconds and have a native Go time type? It might reduce the amount of rounding error that will result each time we add a data point. It seems like this could be done by just removing the `.Seconds()` rounding in the update pipeline.

model/historical_test_data.go
```diff
@@ -61,130 +72,150 @@ func (d *HistoricalTestData) Setup(e cedar.Environment) { d.env = e }
...
+	pipeline := []bson.M{
+		{"$set": bson.M{historicalTestDataLastUpdateKey: "$$NOW"}},
+	}
+	switch result.Status {
+	case "pass":
+		pipeline = append(
+			pipeline,
+			bson.M{"$set": bson.M{
+				historicalTestDataAverageDurationKey: bson.M{"$cond": bson.M{
+					"if": bson.M{"$gte": []interface{}{fmt.Sprintf("$%s", historicalTestDataAverageDurationKey), 1}},
```
Kimchelly, line 128: Should we use `bson.D` types instead of `[]interface{}`?

model/historical_test_data.go
```diff
@@ -61,130 +72,150 @@ func (d *HistoricalTestData) Setup(e cedar.Environment) { d.env = e }
 // IsNil returns if the HistoricalTestData is populated or not.
 func (d *HistoricalTestData) IsNil() bool { return !d.populated }
 
-// Find searches the globally configured pail bucket for the
-// HistoricalTestData. The enviromemt should not be nil.
+// Find searches the database for the HistoricalTestData. The enviromemt should
+// not be nil.
 func (d *HistoricalTestData) Find(ctx context.Context) error {
```
Kimchelly, line 77: Can we specify in the doc comment that this only finds by ID?

model/historical_test_data.go
```diff
@@ -61,130 +72,150 @@ func (d *HistoricalTestData) Setup(e cedar.Environment) { d.env = e }
...
 
 	return nil
 }
 
-// Save saves HistoricalTestData to the Pail backed storage. The
-// HistoricalTestData should be populated and the environment should not be
-// nil.
-func (d *HistoricalTestData) Save(ctx context.Context) error {
+// Update updates the HistoricalTestData with the new data. If the
+// HistoricalTestData, it is created. The HistoricalTestData should be
```
Kimchelly, line 99: If the HistoricalTestData does not exist, it is created.

Kimchelly (changes requested):

model/config.go
```diff
@@ -172,15 +172,13 @@ var (
...
 // Credentials and other configuration information for pail Bucket usage.
 type BucketConfig struct {
-	AWSKey                       string   `bson:"aws_key" json:"aws_key" yaml:"aws_key"`
-	AWSSecret                    string   `bson:"aws_secret" json:"aws_secret" yaml:"aws_secret"`
-	BuildLogsBucket              string   `bson:"build_logs_bucket" json:"build_logs_bucket" yaml:"build_logs_bucket"`
-	SystemMetricsBucket          string   `bson:"system_metrics_bucket" json:"system_metrics_bucket" yaml:"system_metrics_bucket"`
-	SystemMetricsBucketType      PailType `bson:"system_metrics_bucket_type" json:"system_metrics_bucket_type" yaml:"system_metrics_bucket_type"`
-	TestResultsBucket            string   `bson:"test_results_bucket" json:"test_results_bucket" yaml:"test_results_bucket"`
-	TestResultsBucketType        PailType `bson:"test_results_bucket_type" json:"test_results_bucket_type" yaml:"test_results_bucket_type"`
-	HistoricalTestDataBucket     string   `bson:"historical_test_data_bucket" json:"historical_test_data_bucket" yaml:"historical_test_data_bucket"`
```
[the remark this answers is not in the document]
julianedwards: Created https://jira.mongodb.org/browse/BUILD-12429

model/historical_test_data.go
```diff
@@ -1,39 +1,51 @@
...
+	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
+const historicalTestDataCollection = "historical_test_data"
+
 // HistoricalTestData describes aggregated test result data for a given date
 // range.
 type HistoricalTestData struct {
-	Info            HistoricalTestDataInfo `bson:"-"`
+	ID              string                 `bson:"_id,omitempty"`
```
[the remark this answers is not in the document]
julianedwards: Yeah, so if it is empty, mongo creates one.

model/historical_test_data.go
julianedwards: I don't think it matters. I hate `bson.D` cause it adds so many curly braces and makes it less readable.

model/historical_test_data.go
julianedwards: The problem with storing it as `time.Duration` is that we are doing the calculation in mongo, so it gets stored as a `float64`, and decoding fails when we want to go back to `time.Duration`.

julianedwards: What if we keep it as a `float64` but store it as nanoseconds in the database, and convert to seconds in `rest/model/historical_test_data.go`?

julianedwards: I updated the code to store ns in the database.

model/historical_test_data.go
julianedwards: Done.

model/historical_test_data.go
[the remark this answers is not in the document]
Kimchelly: It seems like the only way to create document is through upsertion in `Update()` and it always set the `_id`, which is relied upon in `Find()`. Doesn't that make omitempty unnecessary?

model/historical_test_data.go
Kimchelly: Can we unmarshal it as `time.Duration` then? It seems like if it's already in nanoseconds, it could be converted to the correct type (from float to long) by using [$toLong](https://docs.mongodb.com/manual/reference/operator/aggregation/toLong).

julianedwards: Done.

Kimchelly (approved):

PR #300, opened 2020-12-17
EVG-13536: Add perf rollups based on counter.n

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-13536

This adds two rollups based on `n` (number of documents) from the raw ftdc data, `DocumentThroughput` and `DocumentsTotal`.


aggrand (approved):
The total rollup looks great! Seems to perfectly parallel OperationsTotal and OperationThroughput which is what we want. LGTM as far as that's concerned.

ybrill (approved):

PR #302, opened 2021-01-22
EVG-13447: GetHistoricalStatsData aggregation

julianedwards:

This PR copies over the aggregation to fetch historical test data from evergreen.


model/historical_test_data.go
```diff
@@ -258,3 +269,484 @@ func (i *HistoricalTestDataInfo) validate() error {
...
+	Date    time.Time
+	Variant string
+	Task    string
+	Test    string
+}
+
+func (s *HTDStartAt) validate(groupBy HTDGroupBy) error {
+	catcher := grip.NewBasicCatcher()
+	if s == nil {
+		catcher.New("StartAt should not be nil")
```
Kimchelly, line 375: nit: can we just use `catcher.NewWhen()` for these if statements?

model/historical_test_data.go
```diff
@@ -258,3 +269,484 @@ func (i *HistoricalTestDataInfo) validate() error {
...
+	GroupNumDays int
+	GroupBy      HTDGroupBy
+	StartAt      *HTDStartAt
+	Limit        int
+	Sort         HTDSort
+}
+
+func (f *HistoricalTestDataFilter) validate() error {
+	catcher := grip.NewBasicCatcher()
+	if f == nil {
```
Kimchelly, line 419: Same question about using `catcher.NewWhen()`.

model/historical_test_data.go
```diff
@@ -258,3 +269,484 @@ func (i *HistoricalTestDataInfo) validate() error {
...
+	catcher.Add(f.GroupBy.validate())
+	catcher.Add(f.validateDates())
+
+	return catcher.Resolve()
+}
+
+func (f *HistoricalTestDataFilter) validateDates() error {
+	catcher := grip.NewBasicCatcher()
+	if !f.AfterDate.Equal(utility.GetUTCDay(f.AfterDate)) {
+		catcher.New("invalid AfterDate value")
```
Kimchelly, line 449: Can we make it more obvious that the AfterDate/BeforeDate is only valid if it's equal to midnight UTC of a day?

model/historical_test_data.go
```diff
@@ -258,3 +269,484 @@ func (i *HistoricalTestDataInfo) validate() error {
...
+		},
+	}
+	switch f.GroupBy {
+	case HTDGroupByVariant:
+		fields = append(fields, htdPaginationField{
+			field:  bsonutil.GetDottedKeyName(historicalTestDataInfoKey, historicalTestDataInfoVariantKey),
+			strict: true,
+			value:  f.StartAt.Variant,
+		})
+		fallthrough
```
Kimchelly, line 622: Just to check - is the assumption that, if you group by variant, you also have to specify a task and a test? And similarly, if you group by task, you also have to specify a test?

model/historical_test_data.go
```diff
@@ -258,3 +269,484 @@ func (i *HistoricalTestDataInfo) validate() error {
...
+			strict: false,
+			value:  f.StartAt.Test,
+		})
+	}
+
+	return buildPaginationOrBranches(fields)
+}
+
+// buildPaginationOrBranches builds and returns the $or branches of the
+// pagination constraints. fields is an array of field names, they must be in
```
Kimchelly, line 642: ...array of field names, and they must be...

model/historical_test_data.go
```diff
@@ -258,3 +269,484 @@ func (i *HistoricalTestDataInfo) validate() error {
...
+	} else {
+		return bson.M{
+			"$gte": pf.value,
+			"$lt":  pf.nextValue,
+		}
+	}
+}
+
+// getNextExpression returns an expression that can be used to match the
+// documents which have a field value greater or smaller than the this
```
Kimchelly, line 723: ...smaller than this...

model/historical_test_data.go
```diff
@@ -258,3 +269,484 @@ func (i *HistoricalTestDataInfo) validate() error {
...
+	case HTDSortLatestFirst:
+	default:
+		return errors.Errorf("invalid HTDSort value: %s", s)
+	}
+
+	return nil
+}
+
+// HTDStartAt represents parameters that allow a search query to resume at a
+// specific point. Used for pagination.
```
Kimchelly, line 364: This seems to have an encoded assumption that it's hierarchical, as in, if you specify a variant, you must specify task and test. Can we just add a short comment on this so anyone who reads this has less confusion next time?

Kimchelly (changes requested):

model/historical_test_data.go
```diff
@@ -258,3 +269,484 @@ func (i *HistoricalTestDataInfo) validate() error {
 
 	return catcher.Resolve()
 }
+
+///////////////////
+// Find aggregation
+///////////////////
+
+// AggregatedHistoricalTestData represents aggregated test execution data.
+type AggregatedHistoricalTestData struct {
```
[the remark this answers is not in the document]
julianedwards: Only `TaskName` and `Variant` could be empty, I will at omitempty to those lines.

model/historical_test_data.go
julianedwards: You always have to specify a task and test (see the validation functions above). But if you group by variant, it will also group by task and test, if you group by task it also groups by test, and if you group by test it just groups by test. That is always why variant and task can be returned emtpy.

model/historical_test_data.go
julianedwards: So test must always be specified (I think that is a given since this is the historical test data), but the only time variant and task must be specified is if `HTDGroupByVariant` is set (then variant and task must be set) or `HTDGroupByTask` is set (in which case task must be set).

model/historical_test_data.go
julianedwards: That is done in user facing documentation: https://github.com/evergreen-ci/evergreen/wiki/REST-V2-Usage#teststats

Kimchelly (approved):

PR #303, opened 2021-01-26
EVG-13447: Implement historical test routes in cedar

julianedwards:

This is the final PR for this ticket. It actually adds the REST route to fetch historical test data using the aggregation merged with a previous PR.

### DIFF
--- a/rest/historical_test_data_routes.go
+++ b/rest/historical_test_data_routes.go
@@ -107,7 +107,9 @@ func (h *historicalTestDataHandler) Run(ctx context.Context) gimlet.Responder {
 				"problem paginating response"))
 		}
 	}
-	resp.AddData(data[:lastIndex])
+	if err = resp.AddData(data[:lastIndex]); err != nil {
+		return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "problem creating response"))
+	}
 
 	return resp
 }
@@ -145,6 +147,9 @@ func (h *htdFilterHandler) parse(vals url.Values) error {
 	}
 
 	h.filter.StartAt, err = h.readStartAt(vals.Get("start_at"))
+	if err != nil {
+		return err
+	}
 
 	h.filter.GroupBy, err = h.readGroupBy(vals.Get("group_by"))
 	if err != nil {
@@ -236,7 +241,7 @@ func (h *htdFilterHandler) readRequesters(requesters []string) ([]string, error)
 		case htdAPIRequesterAdhoc:
 			requesterValues = append(requesterValues, cedar.AdHocRequester)
 		default:
-			return nil, errors.Errorf("Invalid requester value %v", requester)
+			return nil, errors.Errorf("invalid requester value %v", requester)
 		}
 	}
 	return requesterValues, nil
@@ -266,7 +271,7 @@ func (h *htdFilterHandler) readInt(intString string, min, max, defaultValue int)
 	}
 
 	if value < min || value > max {
-		return 0, errors.New("Invalid int parameter value")
+		return 0, errors.New("invalid int parameter value")
 	}
 	return value, nil
 }
--- a/rest/historical_test_data_routes.go
+++ b/rest/historical_test_data_routes.go
@@ -50,7 +50,7 @@ const (
 
 ///////////////////////////////////////////////////////////////////////////////
 //
-// GET /historical_test_data/<project_id>
+// GET /historical_test_data/{project_id}
 
 type historicalTestDataHandler struct {
 	sc data.Connector
--- a/rest/historical_test_data_routes.go
+++ b/rest/historical_test_data_routes.go
@@ -1,0 +1,343 @@
+package rest
+
+import (
+	"context"
+	"net/http"
+	"net/url"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/evergreen-ci/cedar"
+	"github.com/evergreen-ci/cedar/model"
+	"github.com/evergreen-ci/cedar/rest/data"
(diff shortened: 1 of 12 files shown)


rest/historical_test_data_routes.go
```diff
@@ -0,0 +1,348 @@
...
+	switch groupByValue {
+	case htdAPITestGroupByVariant:
+		return model.HTDGroupByVariant, nil
+	case htdAPITestGroupByTask:
+		return model.HTDGroupByTask, nil
+	case htdAPITestGroupByTest:
+		return model.HTDGroupByTest, nil
+	// We no longer store distro, but do not want to error if it is
+	// present.
+	case "", htdAPITestGroupByDistro:
```
Kimchelly, line 309: Is this a backwards compatibility case? Is it not possible to do this distro->variant translation in the Evergreen REST route rather than preserve its clunky behavior here?

rest/historical_test_data_routes.go
```diff
@@ -0,0 +1,348 @@
...
+	if err != nil {
+		return err
+	}
+
+	h.filter.GroupBy, err = h.readGroupBy(vals.Get("group_by"))
+	if err != nil {
+		return err
+	}
+
+	h.filter.Limit, err = h.readInt(vals.Get("limit"), 1, htdAPIMaxLimit, htdAPIMaxLimit)
```
Kimchelly, line 159: Is the min supposed to be 1 instead of `htdAPIMaxLimit`?

rest/historical_test_data_routes.go
```diff
@@ -0,0 +1,348 @@
...
+	if err != nil {
+		return gimlet.ErrorResponse{
+			Message:    "invalid limit value",
+			StatusCode: http.StatusBadRequest,
+		}
+	}
+	// Add 1 for pagination.
+	h.filter.Limit++
```
Kimchelly, line 167: Just to check my understanding - is the reason we need to add 1 to the limit to know if there's more data or not (i.e. at least one more data point) and is why you check `len(data)` against `requestLimit` above?

rest/service.go
```diff
@@ -331,5 +331,7 @@ func (s *Service) addRoutes() {
 	s.app.AddRoute("/test_results/task_id/{task_id}").Version(1).Get().RouteHandler(makeGetTestResultsByTaskId(s.sc))
 	s.app.AddRoute("/test_results/test_name/{task_id}/{test_name}").Version(1).Get().RouteHandler(makeGetTestResultByTestName(s.sc))
 
+	s.app.AddRoute("/historical_test_data/<project_id>").Version(1).Get().RouteHandler(makeGetHistoricalTestData(s.sc))
```
Kimchelly, line 334: Is the parameter supposed to be `{project_id}`?

rest/historical_test_data_routes.go
```diff
@@ -0,0 +1,348 @@
...
+	htdAPIMaxNumTasks     = 50
+	htdAPIMaxLimit        = 1000
+
+	// Format used to encode dates in the API.
+	htdAPIDateFormat = "2006-01-02"
+)
+
+///////////////////////////////////////////////////////////////////////////////
+//
+// GET /historical_test_data/<project_id>
```
Kimchelly, line 53: `{project_id}`?

Kimchelly (changes requested):

rest/historical_test_data_routes_test.go
```diff
@@ -0,0 +1,117 @@
+package rest
```
Kimchelly, line 1: I know that the REST route mostly just devolves into parsing and calling the DB connector, but given that the parsing layer is quite involved, should we also integration test that the REST route behaves as expected when calling the DB connector (i.e. it has the same behavior as the DB connector tests as long as you pass in the right REST query parameters?)

rest/historical_test_data_routes.go
julianedwards: The min is 1, the second to last value is the max and the last value is the default value (which is also the max possible value).

rest/historical_test_data_routes.go
```diff
@@ -0,0 +1,348 @@
...
+}
+
+// readStartAt parses a start_at key value and returns the corresponding
+// HTDStartAt struct.
+func (h *htdFilterHandler) readStartAt(startAtValue string) (*model.HTDStartAt, error) {
+	if startAtValue == "" {
+		return nil, nil
+	}
+
+	elements := strings.Split(startAtValue, "|")
```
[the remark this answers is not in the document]
julianedwards: It is for pagination

rest/historical_test_data_routes_test.go
julianedwards: I thought the whole point of the mock connector was for the REST route tests. I think it would be overkill to test using the DB connector here as well.

rest/historical_test_data_routes.go
julianedwards: Yeah it is for backwards compatibility, we can still do that translation in the Evergreen REST route but I think it's fine to keep it just in case.

rest/historical_test_data_routes.go
julianedwards: Yes, so in the `model` package the actual max limit is `1001`. We always want to check one ahead to know for pagination.

rest/historical_test_data_routes.go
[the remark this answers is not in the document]
Kimchelly: Is it actually necessary for the variant/task/test to be part of the pagination? It seems like the only thing that would change with each page in the request would be the date, but I could be wrong on that.

Kimchelly (approved):

PR #305, opened 2021-02-01
EVG-13812: Add size limit to ftdc files we process

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-13812

Limits ftdc processing size limit to 4GB or less.

### DIFF
--- a/perf/ftdc_rollups.go
+++ b/perf/ftdc_rollups.go
@@ -71,6 +71,11 @@ func CreatePerformanceStats(dx *ftdc.ChunkIterator) (*PerformanceStatistics, err
 					perfStats.timers.extractedDurations,
 					extractValues(convertToFloats(metric.Values), lastValue)...,
 				)
+				// In order to avoid memory panics, reject
+				// anything larger than 4GB.
+				if len(perfStats.timers.extractedDurations) > 500000000 {
+					return nil, errors.New("size of ftdc file exceeds 4GB")
+				}
 				lastValue = float64(metric.Values[len(metric.Values)-1])
 				perfStats.timers.durationTotal = time.Duration(metric.Values[len(metric.Values)-1])
 			case "timers.total":
--- a/perf/ftdc_rollups.go
+++ b/perf/ftdc_rollups.go
@@ -8,6 +8,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+const maxDurationsSize = 250000000
+
 type PerformanceStatistics struct {
 	counters struct {
 		operationsTotal int64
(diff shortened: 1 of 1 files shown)


perf/ftdc_rollups.go
```diff
@@ -71,6 +71,11 @@ func CreatePerformanceStats(dx *ftdc.ChunkIterator) (*PerformanceStatistics, err
 					perfStats.timers.extractedDurations,
 					extractValues(convertToFloats(metric.Values), lastValue)...,
 				)
+				// In order to avoid memory panics, reject
+				// anything larger than 4GB.
+				if len(perfStats.timers.extractedDurations) > 500000000 {
+					return nil, errors.New("size of ftdc file exceeds 4GB")
```
ybrill, line 77: if we return an error here won't we be wasting time on the next two attempts?

perf/ftdc_rollups.go
```diff
@@ -71,6 +71,11 @@ func CreatePerformanceStats(dx *ftdc.ChunkIterator) (*PerformanceStatistics, err
 					perfStats.timers.extractedDurations,
 					extractValues(convertToFloats(metric.Values), lastValue)...,
 				)
+				// In order to avoid memory panics, reject
+				// anything larger than 4GB.
+				if len(perfStats.timers.extractedDurations) > 500000000 {
```
ybrill, line 76: is it still possible to run into issues if multiple jobs with large files were to land on the same app server? Would you consider using [MemStats](https://golang.org/pkg/runtime/#MemStats) to find how much memory the application is using in total?

perf/ftdc_rollups.go
```diff
@@ -71,6 +71,11 @@ func CreatePerformanceStats(dx *ftdc.ChunkIterator) (*PerformanceStatistics, err
 					perfStats.timers.extractedDurations,
 					extractValues(convertToFloats(metric.Values), lastValue)...,
 				)
+				// In order to avoid memory panics, reject
+				// anything larger than 4GB.
```
ybrill, line 75: why did you decide on 4GB?

julianedwards: To be safe, it was panicking around 5GB.

perf/ftdc_rollups.go
julianedwards: Yeah, I think its fine for it to run twice more since there are not many jobs this large and this is a temporary solution.

perf/ftdc_rollups.go
ybrill: Still, what about what we do with it downstream: ([copying and then sorting](https://github.com/evergreen-ci/cedar/blob/301d41222f3ccd6ac661a13f25395465a8222801/perf/rollup_factory.go#L278-L279)). 

- Wouldn't copying would put us over again.
- Is the sort in-place? Will it take forever on such a large file?

perf/ftdc_rollups.go
ybrill, line 76: should we make this a constant?

perf/ftdc_rollups.go
julianedwards: Are you suggesting we do half? Say, 2 or 2.5GB?

ybrill: Maybe even that would be too much? 
But do we have a sense of what the typical size of these files is? Like, if we set it at 500MB, for instance, we be disabling other workloads that are currently successful?

julianedwards: I think most are quite small, but I am not sure, there are so many of these tests running.

ybrill: Logging the final size and alerting when we get something too big are probably worthwhile: 
- Logging the size: it would be good to have stats on how big these files are.
- Alerting: we wouldn't have noticed this error if it wasn't for the jobs stopping up the queue. If we're going to just be erroring and moving on points will be missing from the trend charts and we won't know it.

ybrill: followed up "in person": we're going to open a separate PR for logging when this happens.

PR #306, opened 2021-02-03
EVG-13722: db auth framework

hhoke:


bsamek:
I'd like to take a look at this before merging. Adding @ybrill as well. (I'd like to be cautious about these changes, since they could cause the app not to start.)

ybrill:
Given the overlap with https://github.com/evergreen-ci/evergreen/pull/4361 my plan is to hold off on reviewing this PR until that review closes out.
@hhoke let me know if it would be helpful for you to arrange this differently (like if you wanted this in place to let you set up staging cedar, for instance)

hhoke:
@ybrill I want to run this first, and test on cedar staging, before the evergreen PR. I have tested this with the authless version here, in evergreen, and have tested this with an auth file on my mac, using a local mongodb with auth enabled.

operations/flags.go
```diff
@@ -118,6 +119,12 @@ func dbFlags(flags ...cli.Flag) []cli.Flag {
 			Usage:  "specify a database name to use",
 			Value:  "cedar",
 			EnvVar: "CEDAR_DATABASE_NAME",
+		},
+		cli.StringFlag{
+			Name:   dbCredsFileFlag,
+			Usage:  "specify a DB credential file location",
+			Value:  "",
```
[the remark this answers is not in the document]
hhoke: Default should be the zero value for now. Otherwise we might accidentally break prod if we don't set the environment variable to "".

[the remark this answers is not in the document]
ybrill: Why would it break? 
I would think it would look for the file, not find it, grip.Error, and keep going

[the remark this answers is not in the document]
hhoke: Right now there's a unified chef environment for cedar staging and cedar prod, so the yaml files will show up on prod once we turn that on. We'll have to run prod without auth and staging with auth for some time. Once we migrate cedar prod db to atlas, I think it's fine to set this default to our yaml file.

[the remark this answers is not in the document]
ybrill: ah I get it now.
still, we can omit setting `Value:  ""`: it's implied.

operations/setup.go
```diff
@@ -130,12 +136,50 @@ func (c *serviceConf) setup(ctx context.Context) error {
...
+	if err := decoder.Decode(&creds); err != nil {
+		return nil, err
+	}
+
+	return creds, nil
+}
+
+func newServiceConf(numWorkers int, localQueue bool, mongodbURI, bucket, dbName string, dbCredFile string) *serviceConf {
+
+	// should I return an error?
```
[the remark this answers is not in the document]
ybrill: would you mind to remove the comment?

ybrill (approved):

environment.go
```diff
@@ -53,11 +53,21 @@ func NewEnvironment(ctx context.Context, name string, conf *Configuration) (Envi
...
 			SetServerSelectionTimeout(conf.SocketTimeout).
-			SetMonitor(apm.NewLoggingMonitor(ctx, time.Minute, apm.NewBasicMonitor(&apm.MonitorConfig{AllTags: true})).DriverAPM()))
+			SetMonitor(apm.NewLoggingMonitor(ctx, time.Minute, apm.NewBasicMonitor(&apm.MonitorConfig{AllTags: true})).DriverAPM())
+
+		if conf.HasAuth() {
+			credential := options.Credential{
+				Username: conf.DBUser,
+				Password: conf.DBPwd,
+			}
+
```
julianedwards, line 67: Nit: remove blank line.

environment.go
```diff
@@ -53,11 +53,21 @@ func NewEnvironment(ctx context.Context, name string, conf *Configuration) (Envi
...
 	env.conf = conf
 	if env.client == nil {
-		env.client, err = mongo.NewClient(options.Client().ApplyURI(conf.MongoDBURI).
+		opts := options.Client().ApplyURI(conf.MongoDBURI).
 			SetConnectTimeout(conf.MongoDBDialTimeout).
 			SetSocketTimeout(conf.SocketTimeout).
 			SetServerSelectionTimeout(conf.SocketTimeout).
-			SetMonitor(apm.NewLoggingMonitor(ctx, time.Minute, apm.NewBasicMonitor(&apm.MonitorConfig{AllTags: true})).DriverAPM()))
+			SetMonitor(apm.NewLoggingMonitor(ctx, time.Minute, apm.NewBasicMonitor(&apm.MonitorConfig{AllTags: true})).DriverAPM())
+
```
julianedwards, line 61: Nit: remove blank line.

julianedwards (approved):
LGTM modulo small nits

operations/setup.go
```diff
@@ -130,12 +136,49 @@ func (c *serviceConf) setup(ctx context.Context) error {
...
+	DBUser string `yaml:"mdb_database_username"`
+	DBPwd  string `yaml:"mdb_database_password"`
+}
+
+func loadCredsFromYAML(filePath string) (*dbCreds, error) {
+	creds := &dbCreds{}
+
+	file, err := os.Open(filePath)
+	if err != nil {
+		grip.Error(err)
```
bsamek, line 149: nit: It seems incorrect to log the error in this case but not in the case that YAML errors. I would suggest not logging at all and instead relying on the caller.

operations/setup.go
```diff
@@ -130,12 +136,49 @@ func (c *serviceConf) setup(ctx context.Context) error {
...
+
+	return creds, nil
+}
+
+func newServiceConf(numWorkers int, localQueue bool, mongodbURI, bucket, dbName string, dbCredFile string) *serviceConf {
+
+	creds := &dbCreds{}
+	var err error
+	if dbCredFile != "" {
+
```
bsamek, line 168: nit: delete this empty line

operations/setup.go
```diff
@@ -130,12 +136,49 @@ func (c *serviceConf) setup(ctx context.Context) error {
...
+		creds, err = loadCredsFromYAML(dbCredFile)
```
bsamek, line 169: I think this can be `creds, err :=`. I think it's unnecessary to initialize those two variables separately.

operations/setup.go
```diff
@@ -130,12 +136,49 @@ func (c *serviceConf) setup(ctx context.Context) error {
...
+		if err != nil {
```
bsamek, line 170: nit: You don't need an if statement here. You can just `grip.Error(err)`. If err is not nil, this will noop.

bsamek (changes requested):
just nits

operations/setup.go
hhoke: Good catch, this was cruft from a time of extremely indiscriminate debug logging

operations/setup.go
hhoke: Creds needs to be initialized outside the if loop or they go out of scope and can't be used in return. If I don't initialize err and use a walrus, it creates a shadowed creds. As written is the only way I was able to get build to work.

bsamek (approved):

PR #308, opened 2021-02-12
EVG-13831: Change task level logs to use tags instead of process_name

julianedwards:
JIRA: Change task level logs to use tags instead of process_name

This is temporary code for the task log migration.

### DIFF
--- a/model/buildlogger.go
+++ b/model/buildlogger.go
@@ -630,7 +630,7 @@ func findOutdatedTaskLogs(ctx context.Context, env cedar.Environment) ([]Log, er
 			"$nin": []string{AgentLog, TaskLog, SystemLog},
 		},
 	}
-	opts := options.Find().SetLimit(conf.LogMigrationLimit)
+	opts := options.Find().SetLimit(limit)
 	it, err := env.GetDB().Collection(buildloggerCollection).Find(ctx, query, opts)
 	if err != nil {
 		return nil, errors.WithStack(err)
--- a/model/buildlogger.go
+++ b/model/buildlogger.go
@@ -650,6 +650,10 @@ func findOutdatedTaskLogs(ctx context.Context, env cedar.Environment) ([]Log, er
 }
 
 func updateOutdatedTaskLogs(ctx context.Context, env cedar.Environment, ids []string, logType string) error {
+	if len(ids) == 0 {
+		return nil
+	}
+
 	match := bson.M{"_id": bson.M{"$in": ids}}
 	update := bson.M{"$push": bson.M{
 		bsonutil.GetDottedKeyName(logInfoKey, logInfoTagsKey): logType,
--- a/model/buildlogger.go
+++ b/model/buildlogger.go
@@ -616,6 +616,11 @@ func findOutdatedTaskLogs(ctx context.Context, env cedar.Environment) ([]Log, er
 	if err := conf.Find(); err != nil {
 		return nil, errors.WithStack(err)
 	}
+	limit := conf.LogMigrationLimit
+	if limit <= 0 {
+		// Default limit to 1000.
+		limit = 1000
+	}
 
 	query := bson.M{
 		bsonutil.GetDottedKeyName(logInfoKey, logInfoProcessNameKey): bson.M{
--- a/model/buildlogger.go
+++ b/model/buildlogger.go
@@ -575,12 +575,12 @@ const (
 // FindAndUpdateOutdatedTaskLogs finds evergreen task logs in the database
 // with the log type (agent, system, or task) in the info.proc_name field and
 // updates those documents to have the log type in the tags array.
-func FindAndUpdateOutdatedTaskLogs(ctx context.Context, env cedar.Environment) error {
+func FindAndUpdateOutdatedTaskLogs(ctx context.Context, env cedar.Environment, limit int64) error {
 	if env == nil {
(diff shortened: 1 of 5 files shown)


model/buildlogger.go
```diff
@@ -563,3 +563,102 @@ func (l *Logs) Merge(ctx context.Context) (LogIterator, error) {
...
+func findOutdatedTaskLogs(ctx context.Context, env cedar.Environment) ([]Log, error) {
+	conf := &CedarConfig{}
+	conf.Setup(env)
+	if err := conf.Find(); err != nil {
+		return nil, errors.WithStack(err)
+	}
+	limit := conf.LogMigrationLimit
+	if limit <= 0 {
+		// Default limit to 1000.
+		limit = 1000
```
Kimchelly, line 622: Did you test that 1000 is an acceptably performant default to update once per minute?

Can we make it so that there's no default if it's unset? Just in case we want to turn off the migration.

model/buildlogger.go
```diff
@@ -563,3 +563,102 @@ func (l *Logs) Merge(ctx context.Context) (LogIterator, error) {
...
+	}
+	opts := options.Find().SetLimit(limit)
+	it, err := env.GetDB().Collection(buildloggerCollection).Find(ctx, query, opts)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	var logs []Log
+	if err = it.All(ctx, &logs); err != nil {
+		catcher := grip.NewBasicCatcher()
```
Kimchelly, line 641: nit: instead of doing this if-else, can we just initialize a catcher on L639, catch the error from `All()` and `Close()`, then check if `catcher.HasErrors()` before we return?

units/migrate_task_logs.go
```diff
@@ -0,0 +1,74 @@
...
+
+	j.SetDependency(dependency.NewAlways())
+	return j
+}
+
+func NewMigrateTaskLogsJob() amboy.Job {
+	j := makeMigrateTaskLogsJob()
+
+	ts := utility.RoundPartOfMinute(0).Format(tsFormat)
+	j.SetID(fmt.Sprintf("%s.%s", migrateTaskLogsJobName, ts))
```
Kimchelly, line 57: Does this job need a scope in case the doc update is slow to avoid another migration job running concurrently?

units/migrate_task_logs.go
```diff
@@ -0,0 +1,74 @@
...
+	return j
+}
+
+func (j *migrateTaskLogsJob) Run(ctx context.Context) {
+	defer j.MarkComplete()
+
+	if j.env == nil {
+		j.env = cedar.GetEnvironment()
+	}
+	if j.queue == nil {
```
Kimchelly, line 68: Do we still need this?

units/migrate_task_logs.go
```diff
@@ -0,0 +1,74 @@
...
+		j.queue = j.env.GetRemoteQueue()
+	}
+
+	j.AddError(errors.Wrap(model.FindAndUpdateOutdatedTaskLogs(ctx, j.env),
```
Kimchelly, line 72: Should there be some kind of alert when it finishes the migration?

model/buildlogger.go
```diff
@@ -563,3 +563,102 @@ func (l *Logs) Merge(ctx context.Context) (LogIterator, error) {
...
+	return logs, nil
+}
+
+func updateOutdatedTaskLogs(ctx context.Context, env cedar.Environment, ids []string, logType string) error {
+	if len(ids) == 0 {
+		return nil
+	}
+
+	match := bson.M{"_id": bson.M{"$in": ids}}
+	update := bson.M{"$push": bson.M{
```
Kimchelly, line 658: Should we do `$addToSet` just to be safe?

Kimchelly (changes requested):

model/buildlogger.go
julianedwards: I assume `$addToSet` is going to be slower than `$push` since it has to iterate through each array to check for duplicate values. The find query has a `$nin` clause to search for docs without the log type in the tags array. Also, having duplicate values in the tag array wouldn't be the end of the world.

units/migrate_task_logs.go
julianedwards: Done.

units/migrate_task_logs.go
julianedwards: I added grip logging to the function in the `model` package.

model/buildlogger.go
julianedwards: Yes, the I ran the `find` query with the limit set to 1000 multiple times and it ran on the order of 10s of ms. It does an index scan so it is extremely efficient.

Kimchelly (approved):

units/migrate_task_logs.go
```diff
@@ -0,0 +1,78 @@
...
+	j.Limit = limit
+
+	return j, nil
+}
+
+func (j *migrateTaskLogsJob) Run(ctx context.Context) {
+	defer j.MarkComplete()
+
+	if j.env == nil {
+		j.env = cedar.GetEnvironment()
```
ybrill, line 73: Is there a way for the job to fetch the settings from the db?
Like this I think we'd need to restart for it to take effect.

ybrill: oh, nvm.
I get it now.

ybrill (approved):

PR #309, opened 2021-02-16
EVG-13831: Remove proc_name in migration

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-13831

Turns out the index was not being used as we expected, unfortunately `$ne` and `$nin` are not selective and thus indexes cannot be used with them. So as the number of docs with `agent_log`, `task_log`, and `system_log` in the `info.tags` array increased, the time the query took to filter out those docs increased drastically. 

This code changes the update to remove the `info.proc` field as way to mark that doc as processed.


Kimchelly (approved):
LGTM mod question - should we add the number of logs as a separate field in the grip log?

julianedwards:
I don't think we really need to, the count is already in the other logs.

ybrill (approved):

PR #310, opened 2021-02-16
EVG-13831: Output with non-UTF-8 encoding prevents task logs from being sent

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-13464

Use `bytes` instead of `string` in the buildlogger protobuf to enable non UTC-8 encodings, [protobuf strings only enable UTF-8 encoding](https://developers.google.com/protocol-buffers/docs/proto3#scalar).

Also, I am using protobuf 3.6 so it changed the test results protobuf slightly.


julianedwards:
Yeah, unfortunately it is not backwards compatible and will not be very pretty. I think we need to do this on a weekend, if possible.

Kimchelly (approved):

PR #311, opened 2021-02-18
EVG-13637: Remove amboy info and stats from status endpoint

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-13637


ybrill (approved):

Kimchelly (commented):
This seems fine, but would it be better to just to pass a short context rather than remove it? The environment context won't cancel until the app server shuts down, but this shouldn't be a problem you put a short context on the queue stats.

ybrill:
I argued there was no benefit to including amboy stats in the status endpoint: it has no consumers.

Kimchelly (approved):

PR #312, opened 2021-02-19
EVG-13536: Limit outdated rollup job to 1000 results per job

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-13536

The backfill job for outdated perf rollup jobs is already limited to update results created within the past 90 days. For a change that affects every single perf result (like adding a new rollup) this can be overwhelming for the amboy queue, creating duplicate jobs and backing it up. This change reverts the revert of a commit I made recently to add a new perf rollup and adds limits a the backfill job to 1000 results per job (which runs every hour) to avoid overwhelming the queue again. The database has about ~240,000 perf results that were created in the past 90 days. Not all of these results have ftdc data and thus will not need to be updated. But even all ~240,000 results needed to be updated, this would only take about 10 days. I chose 1000 because it seems reasonable and that is the threshold for long cedar amboy queue alerts in splunk.

### DIFF
--- a/units/find_outdated_rollups.go
+++ b/units/find_outdated_rollups.go
@@ -17,6 +17,7 @@ import (
 
 const (
 	findOutdatedRollupsJobName = "find-outdated-rollups"
+	maxOutdatedPerfResults     = 1000
 )
 
 type findOutdatedRollupsJob struct {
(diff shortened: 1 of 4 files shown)


units/find_outdated_rollups.go
```diff
@@ -107,6 +108,14 @@ func (j *findOutdatedRollupsJob) Run(ctx context.Context) {
 			for _, result := range results.Results {
 				if _, ok := j.seenIDs[result.ID]; !ok {
 					j.createFTDCRollupsJobs(ctx, factories[i:], result)
+
+					count += 1
+					if count >= 1000 {
```
Kimchelly, line 113: nit: can we make this a constant, just like `maxDurationsSize`?

Kimchelly (approved):
LGTM mod nit.

units/find_outdated_rollups.go
julianedwards: Done.

PR #315, opened 2021-02-24
EVG-13419: New Find methods for test results

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-13419

This changes the `TestResults.FindByTaskID` method to `FindTestResults` and enables searching by `DisplayTaskName`.

### DIFF
--- a/model/test_results.go
+++ b/model/test_results.go
@@ -17,6 +17,7 @@ import (
 	"github.com/mongodb/grip/message"
 	"github.com/pkg/errors"
 	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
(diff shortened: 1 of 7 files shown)


model/test_results.go
```diff
@@ -378,3 +334,108 @@ type TestResult struct {
...
+		return nil, errors.Wrap(err, "finding test results record(s)")
+	}
+	if err := cur.All(ctx, &results); err != nil {
+		return nil, errors.Wrap(err, "decoding test result record(s)")
+	}
+	if len(results) == 0 {
+		return nil, errors.Wrap(mongo.ErrNoDocuments, opts.createErrorMessage())
+	}
+
+	execution := results[0].Info.Execution
```
Kimchelly, line 409: Should we wrap this block with a check for `if opts.EmptyExecution`, since it should only be necessary when execution is not specified?

Kimchelly (approved):
LGTM mod question.

model/test_results.go
julianedwards: We still need to fill set `populated` and `env` either way.

PR #316, opened 2021-02-26
EVG-13419, EVG-14119: Add LogURL and DisplayTaskID fields to test results and create route to fetch by DisplayTaskID

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-13419, https://jira.mongodb.org/browse/EVG-14119

This updates the protobuf, which will require updates in the client (Timber) and downstream (Evergreen).


model/test_results.go
```diff
@@ -338,10 +342,11 @@ type TestResult struct {
 // TestResultsFindOptions allows for querying for test results with or without
 // an execution value.
 type TestResultsFindOptions struct {
-	TaskID          string
-	DisplayTaskName string
-	Execution       int
-	EmptyExecution  bool
+	TaskID         string
+	DisplayTaskID  string
+	Variant        string
```
Kimchelly, line 347: If we're removing the ability to search by display task name, shouldn't we also get rid of variant? Or should we keep display task name + variant as an optional means to query data?

model/test_results.go
```diff
@@ -390,8 +395,8 @@ func FindTestResults(ctx context.Context, env cedar.Environment, opts TestResult
 		return nil, errors.New("cannot find with a nil environment")
 	}
 
-	if opts.TaskID == "" && opts.DisplayTaskName == "" {
-		return nil, errors.New("cannot find without a task_id or display_task_name")
+	if opts.TaskID == "" && opts.DisplayTaskID == "" {
```
Kimchelly, line 398: Should this also check for both being set, since [they're mutually exclusive](https://github.com/evergreen-ci/cedar/pull/316/files#diff-98447d1f255e456b4bdd74b07e213d03b003349d7113febe787beebeca269047R363-R366)?

rest/data/interface.go
```diff
@@ -92,14 +92,17 @@ type Connector interface {
...
-	// FindTestResultsByTaskId queries the database to find all test
+	// FindTestResultsByTaskID queries the database to find all test
 	// results with the given options. If the execution is not specified,
 	// this will return the test results from the most recent execution.
-	FindTestResultsByTaskId(context.Context, TestResultsOptions) ([]model.APITestResult, error)
+	FindTestResultsByTaskID(context.Context, TestResultsOptions) ([]model.APITestResult, error)
+	// FindTestResultsByDisplayTaskID queries the database to find all test
+	// results with the given options. If the execution is not specified,
+	// this will return the test results from the most recent execution.
+	FindTestResultsByDisplayTaskID(context.Context, TestResultsOptions) ([]model.APITestResult, error)
```
Kimchelly, line 102: Can't we just make one `FindTestResults` method for `ByTaskID` and `ByDisplayTaskID` since they have the exact same signature and the same code?

rest/data/test_results.go
```diff
@@ -106,50 +103,55 @@ func (mc *MockConnector) FindTestResultsByTaskId(ctx context.Context, opts TestR
 	if testResults == nil {
 		return nil, gimlet.ErrorResponse{
 			StatusCode: http.StatusNotFound,
-			Message:    fmt.Sprintf("test results with task id '%s' not found", opts.TaskID),
+			Message:    fmt.Sprintf("test results with display_task_id '%s' not found", opts.TaskID),
```
Kimchelly, line 106: I think this should still be task ID.

rest/data/test_results.go
```diff
@@ -235,11 +258,27 @@ func getAPITestResultFromBucket(ctx context.Context, bucket pail.Bucket, testNam
 	return apiResult, ctx.Err()
 }
 
-func convertToDBOpts(opts TestResultsOptions) dbModel.TestResultsFindOptions {
+func getTestResultsBucket(ctx context.Context, path, prefix string) (pail.Bucket, error) {
```
Kimchelly, line 261: Can this be called `getMockTestResultsBucket` so nobody tries to use it in the real DB connector?

rest/test_results_routes.go
```diff
@@ -65,6 +65,59 @@ func (h *testResultsGetByTaskIdHandler) Run(ctx context.Context) gimlet.Responde
...
+}
+
+// Factory returns a pointer to a new testResultsGetByDisplayTaskIDHandler.
+func (h *testResultsGetByDisplayTaskIDHandler) Factory() gimlet.RouteHandler {
+	return &testResultsGetByDisplayTaskIDHandler{
+		sc: h.sc,
+	}
+}
+
+// Parse fetches the task_id from the http request.
```
Kimchelly, line 90: display_task_id

rest/test_results_routes.go
```diff
@@ -65,6 +65,59 @@ func (h *testResultsGetByTaskIdHandler) Run(ctx context.Context) gimlet.Responde
...
+func (h *testResultsGetByDisplayTaskIDHandler) Parse(_ context.Context, r *http.Request) error {
+	var err error
+
+	h.opts.DisplayTaskID = gimlet.GetVars(r)["display_task_id"]
+	vals := r.URL.Query()
+	if len(vals[execution]) > 0 {
+		h.opts.Execution, err = strconv.Atoi(vals[execution][0])
+		return err
+	} else {
```
Kimchelly, line 99: nit: since you already return in the if clause, you don't need to have else here.

Kimchelly (changes requested):

model/test_results.go
julianedwards: No, I forgot to delete from an earlier iteration of the code. Removing.

model/test_results.go
julianedwards: If both are set, it will use `TaskID`. Would you rather it error?

rest/data/test_results.go
julianedwards: Made it a function on the `MockConnector`.

model/test_results.go
Kimchelly: I think it would make more sense to error since it basically just ignores the display task ID.

Kimchelly (approved):
LGTM mod comment about find with both task ID and display task ID.

PR #319, opened 2021-03-05
EVG-14097: fix flaky tests

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-14097

Also found a real bug in the `CreateHistoricalTestData` function. Ignore the reverts, I by accidentally checked this branch out from another branch that is currently in CR.

### DIFF
--- a/model/buildlogger_test.go
+++ b/model/buildlogger_test.go
@@ -378,7 +378,7 @@ func TestBuildloggerAppend(t *testing.T) {
 		}
 		var filenames map[string]bool
 		var actualData []byte
-		for i := 0; i < 10; i++ {
+		for i := 0; i < 20; i++ {
 			filenames = map[string]bool{}
 			actualData = []byte{}
 			iter, err := testBucket.List(ctx, log.ID)
--- a/model/historical_test_data.go
+++ b/model/historical_test_data.go
@@ -51,13 +51,7 @@ func CreateHistoricalTestData(info HistoricalTestDataInfo) (*HistoricalTestData,
 		return nil, err
 	}
 
-	info.Date = time.Date(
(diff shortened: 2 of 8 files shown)


model/buildlogger_test.go
```diff
@@ -378,7 +378,7 @@ func TestBuildloggerAppend(t *testing.T) {
 		}
 		var filenames map[string]bool
 		var actualData []byte
-		for i := 0; i < 10; i++ {
+		for i := 0; i < 20; i++ {
```
julianedwards, line 381: This is annoying, but even with the 10 retries it fails quite often on macs and once in a while on linux.

Kimchelly: Do you know why it's slow? As far as I see, Append is synchronous so I would imagine it would be written to the filesystem without needing to periodically retry.

Kimchelly (approved):
LGTM mod question

julianedwards:
That is what I thought, but it gives us trouble. If you look at [this test](https://evergreen.mongodb.com/test_log/603d11283e8e8669cc29b01f#L65) for example, only half the data is written. It failed way more often before I put the original retries.

Kimchelly:
I'm aware of that, but it does seem kind of weird and breaks my mental model that something that should theoretically be a synchronous write to the filesystem isn't. Should we bother looking into why, or should we just assume that writing to the local Pail bucket will be asynchronous?

julianedwards:
My assumption is that it is some slowness with the filesystem itself. But seems worth it to investigate the pail side of things. I will open a ticket.

julianedwards:
New ticket: https://jira.mongodb.org/browse/EVG-14223

PR #320, opened 2021-03-08
EVG-14128: remove cost reporting features

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-14128

* Remove cost package, its related DB models, Amboy jobs, and client functionality.
    * Move the Evergreen REST client into a separate package.
* Update README since it's outdated.


README.rst
```diff
@@ -5,58 +5,47 @@
...
 
 Components
 ~~~~~~~~~~
 
-Cedar has three core components:
+- A simple JSON-only REST web service to data storage and access (see the ``rest`` package).
 
-- a simple JSON-only REST web service to data storage and access. (See
-  the ``rest`` sub-package.)
+- A gRPC service as an alternative to the REST API, which also supports data storage and access (see the ``rpc``
```
julianedwards, line 22: This actually does not support access, just data storage.

README.rst
```diff
@@ -5,58 +5,47 @@
...
+Build operations generate data, and Evergreen stores most of this data directly (e.g. the outcome of tasks, artifacts
+produced by builds). There's other data, specific to particular projects, that is less naturally structured or less
+specifically linked to specific execution events. Additionally, Cedar provides a platform to collect and analyze data
+*about* an Evergreen deployment in the large.
 
 Components
 ~~~~~~~~~~
 
-Cedar has three core components:
+- A simple JSON-only REST web service to data storage and access (see the ``rest`` package).
```
julianedwards, line 20: This only supports access.

evergreen/client.go
```diff
@@ -1,4 +1,4 @@
-package cost
+package evergreen
```
julianedwards, line 1: Do we need this still?

model/evergreen_config.go
```diff
@@ -0,0 +1,28 @@
+package model
```
julianedwards, line 1: We can get rid of this, it is a duplicate of https://github.com/evergreen-ci/cedar/blob/main/model/config.go#L65-L73.

julianedwards (changes requested):
We can get rid of the evergreen connection info config and use the existing evergreen config.

README.rst
Kimchelly: Fixed.

README.rst
Kimchelly: Fixed.

evergreen/client.go
Kimchelly: It's required for the historical test stats job.

model/evergreen_config.go
Kimchelly: I moved it to the evergreen package because we still need an options struct.

PR #321, opened 2021-03-08
EVG-14242: Parallelize downloading test results

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-14242

The test results rest routes have long latencies (~5s) due to the fact that they must iterator through a pail bucket and download each file individually. To speed up the process I added some parallelization.

### DIFF
--- a/model/test_results_iterator.go
+++ b/model/test_results_iterator.go
@@ -3,12 +3,17 @@ package model
 import (
 	"context"
 	"io/ioutil"
+	"runtime"
+	"sync"
 
 	"github.com/evergreen-ci/pail"
 	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/message"
+	"github.com/mongodb/grip/recovery"
 	"go.mongodb.org/mongo-driver/bson"
 )
+
+const testResultsIteratorBatchSize = 100
 
 // TestResultsIterator is an interface that enables iterating over test results.
 type TestResultsIterator interface {
@@ -23,7 +28,9 @@ type TestResultsIterator interface {
 
 type testResultsIterator struct {
 	bucket      pail.Bucket
-	iter        pail.BucketIterator
+	bucketItems []pail.BucketItem
+	currentIdx  int
+	items       chan TestResult
 	currentItem TestResult
 	exhausted   bool
 	closed      bool
@@ -34,6 +41,7 @@ type testResultsIterator struct {
 func NewTestResultsIterator(bucket pail.Bucket) TestResultsIterator {
 	return &testResultsIterator{
 		bucket:  bucket,
+		items:   make(chan TestResult, testResultsIteratorBatchSize),
 		catcher: grip.NewBasicCatcher(),
 	}
 }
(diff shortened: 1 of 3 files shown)


model/test_results_iterator.go
```diff
@@ -49,47 +57,103 @@ func (i *testResultsIterator) Next(ctx context.Context) bool {
...
-		return false
+	close(idxs)
+	var wg sync.WaitGroup
+	catcher := grip.NewBasicCatcher()
+
+	for j := 0; j < runtime.NumCPU(); j++ {
+		wg.Add(1)
+		go func() {
+			defer func() {
+				var err error
```
Kimchelly, line 114: Is this supposed to be declared here? This seems equivalent to just passing nil for the error to `recovery.HandlePanicWithError`.

I noticed this is also done in the log iterator.

model/test_results_iterator.go
```diff
@@ -49,47 +57,103 @@ func (i *testResultsIterator) Next(ctx context.Context) bool {
...
+				catcher.Add(recovery.HandlePanicWithError(recover(), err))
```
Kimchelly, line 115: Can we add a message as the 3rd argument to `recovery.HandlePanicWithError` so we have context for the source of the error (e.g. "iterating test results in worker")?

model/test_results_iterator.go
```diff
@@ -49,47 +57,103 @@ func (i *testResultsIterator) Next(ctx context.Context) bool {
...
+				wg.Done()
+			}()
+
+			for idx := range idxs {
+				if err := ctx.Err(); err != nil {
+					catcher.Add(err)
+					return
+				} else {
```
Kimchelly, line 123: nit: since the if clause returns, the else is unnecessary.

Kimchelly (changes requested):

model/test_results_iterator.go
julianedwards: Yeah it should just be nil, copy and paste error. I will change it in both places.

model/test_results_iterator.go
julianedwards: More copying and pasting, I also changed it in the log iterator code.

model/test_results_iterator.go
```diff
@@ -3,13 +3,18 @@ package model
...
+	"sync"
 
 	"github.com/evergreen-ci/pail"
 	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/message"
+	"github.com/mongodb/grip/recovery"
 	"go.mongodb.org/mongo-driver/bson"
 )
 
+const testResultsIteratorBatchSize = 100
```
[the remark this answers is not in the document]
Kimchelly: I did think about this, but there's a few other places where configuring batch size would also be useful but it's hard-coded (e.g. the system metrics batch size, log iterator batch size). I would make it a separate ticket to add config options for batch sizing where it's currently hard-coded constants.

Kimchelly (approved):

PR #324, opened 2021-03-17
EVG-14223: Fix flaky buildlogger append test

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-14223

This took me hours to figure out and ended up being really stupid: occasionally the pail file is overwritten by the second call to append. The pail key is a string of the unix timestamp in milliseconds of `time.Now` and since the appends are called sequentially, this would sometimes result in the same millisecond precision unix timestamp and the first log file would get overwritten.

### DIFF
--- a/model/buildlogger_test.go
+++ b/model/buildlogger_test.go
@@ -364,6 +364,9 @@ func TestBuildloggerAppend(t *testing.T) {
 	t.Run("AppendToBucketAndDB", func(t *testing.T) {
 		log.Setup(env)
 		require.NoError(t, log.Append(ctx, chunk1))
+		// We need to sleep for a ms here to avoid a key name collision
+		// in the pail-backed blog storage since key names are the
+		// current time as a unix timestamp with millisecond precision.
 		time.Sleep(time.Millisecond)
 		require.NoError(t, log.Append(ctx, chunk2))
 		expectedData := []byte{}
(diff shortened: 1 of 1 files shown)


model/buildlogger_test.go
```diff
@@ -327,7 +326,7 @@ func TestBuildloggerAppend(t *testing.T) {
 		{
 			Priority:  level.Info,
 			Timestamp: time.Now().Add(-56 * time.Minute).Round(time.Millisecond).UTC(),
-			Data:      "Buildogger logging logs.",
```
Kimchelly, line 330: buildogger 🐶

model/buildlogger_test.go
```diff
@@ -365,42 +364,29 @@ func TestBuildloggerAppend(t *testing.T) {
 	t.Run("AppendToBucketAndDB", func(t *testing.T) {
 		log.Setup(env)
 		require.NoError(t, log.Append(ctx, chunk1))
+		time.Sleep(time.Millisecond)
```
Kimchelly, line 367: Can you add a comment explaining that appends must not overlap so they must occur in different miliseconds?

Kimchelly (approved):
LGTM mod comment

model/buildlogger_test.go
julianedwards: Done.

PR #326, opened 2021-03-19
EVG-14234: always compile with go build

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-14234

* Always recompile when `build/cedar` is the target or dependency. This is more correct than just assuming we don't have to compile if the binary exists, even if it's outdated (what it does now).
    * As an alternative, we could do what Evergreen does, which is run `find` to locate all the Go source files, then use those as Make dependencies for compilation. However, this is inefficient (slower than just letting Go recompile using its build cache) and also runs unconditionally (even for targets that are unrelated to compiling).
* Make the `clean` target simpler.

### DIFF
--- a/makefile
+++ b/makefile
@@ -58,17 +58,17 @@ coverageHtmlOutput := $(foreach target,$(packages),$(buildDir)/output.$(target).
 
 
 # implementation details for building the binary and creating a
-# convienent link in the working directory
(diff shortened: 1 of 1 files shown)


julianedwards (approved):
Thank you for doing this!

PR #328, opened 2021-03-22
EVG-14237: delete dagger-related features

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-14237

* Delete dagger-related code:
    * REST routes
    * DB models
    * CLI commands
* Skip an test that relies on S3 credentials in the `model` package using `SKIP_INTEGRATION_TESTS`.


model/pail_storage_test.go
```diff
@@ -14,6 +16,9 @@ import (
 )
 
 func TestGetDownloadURL(t *testing.T) {
+	if skip, _ := strconv.ParseBool(os.Getenv("SKIP_INTEGRATION_TESTS")); skip {
```
julianedwards, line 19: This will still run in the waterfall, correct?

julianedwards (approved):

model/pail_storage_test.go
Kimchelly: Yeah, this is just a convenience flag you can set in your local environment to skip these tests.

PR #333, opened 2021-03-25
EVG-14280: grpc health check

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-14280

I followed the grpc recommended [health check service](https://github.com/grpc/grpc/blob/master/doc/health-checking.md). I did not add the streaming health check since we don't really have a use for it. This does leave us with a flexible model though where we could add streaming or set different statuses if we ever enable disabling of services.


rpc/internal/health_service.go
```diff
@@ -0,0 +1,38 @@
...
+	services map[string]HealthCheckResponse_ServingStatus
+	env      cedar.Environment
+}
+
+// AttachHealthService attaches the health check service to the given gRPC
+// server.
+func AttachHealthService(env cedar.Environment, services map[string]HealthCheckResponse_ServingStatus, s *grpc.Server) {
+	srv := &healthService{
+		services: services,
+		env:      env,
```
Kimchelly, line 22: Is `env` necessary for this? Seems like it's unused.

Should this return unhealthy when the env context is done and Cedar is shutting down?

julianedwards: You are right, eventually we may want to ping the DB or something but for now we can leave it out.

Kimchelly (approved):

hhoke (commented):
LGTM

hhoke (approved):

PR #344, opened 2021-04-19
EVG-14439: Send historical test data with test results

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-14439

This adds two new fields to the test results protobuf to send the project level historical test data information along with test results.


rpc/internal/test_results_service.go
```diff
@@ -74,13 +82,15 @@ func (s *testResultsService) AddTestResults(ctx context.Context, results *TestRe
...
-	// for _, res := range exportedResults {
-	//     grip.Error(message.WrapError(s.env.GetRemoteQueue().Put(ctx, units.NewHistoricalTestDataJob(s.env, record.Info, res)), message.Fields{
-	//         "message":     "failed to enqueue historical test data job",
-	//         "info":        record.Info,
-	//         "test_result": res,
-	//     }))
-	// }
+	if record.Info.HistoricalTestData {
+		for _, res := range exportedResults {
+			grip.Error(message.WrapError(s.env.GetRemoteQueue().Put(ctx, units.NewHistoricalTestDataJob(s.env, record.Info, res)), message.Fields{
```
Kimchelly, line 87: Do we still need this job? I thought it was going to be deleted or just in-lined. It would be much more efficient if it was inlined given that it has to run really often.

model/test_results.go
```diff
@@ -274,17 +274,18 @@ func (t *TestResults) GetBucket(ctx context.Context) (pail.Bucket, error) {
...
+	Version            string `bson:"version,omitempty"`
+	Variant            string `bson:"variant,omitempty"`
+	TaskName           string `bson:"task_name,omitempty"`
+	DisplayTaskName    string `bson:"display_task_name,omitempty"`
+	TaskID             string `bson:"task_id,omitempty"`
+	DisplayTaskID      string `bson:"display_task_id,omitempty"`
+	Execution          int    `bson:"execution"`
+	RequestType        string `bson:"request_type,omitempty"`
+	Mainline           bool   `bson:"mainline,omitempty"`
+	HistoricalTestData bool   `bson:"historical_test_data"`
```
Kimchelly, line 287: Can we call this something more obvious like `ShouldCalculateHistoricalStats` so it's not confused with `HistoricalDataDisabled`?

rpc/internal/test_results.ext_test.go
```diff
@@ -23,16 +23,79 @@ func TestTestResultsInfoExport(t *testing.T) {
...
+		assert.Equal(t, info.Variant, modelInfo.Variant)
+		assert.Equal(t, info.TaskName, modelInfo.TaskName)
+		assert.Equal(t, info.DisplayTaskName, modelInfo.DisplayTaskName)
+		assert.Equal(t, info.TaskId, modelInfo.TaskID)
+		assert.Equal(t, int(info.Execution), modelInfo.Execution)
+		assert.Equal(t, info.RequestType, modelInfo.RequestType)
+		assert.True(t, modelInfo.HistoricalTestData)
+		assert.Equal(t, info.Mainline, modelInfo.Mainline)
+	})
+	t.Run("HistoricalTestDataDisbaled", func(t *testing.T) {
```
Kimchelly, line 40: Disabled

rpc/internal/test_results.ext.go
```diff
@@ -1,26 +1,64 @@
...
+		TaskID:             t.TaskId,
+		DisplayTaskID:      t.DisplayTaskId,
+		Execution:          int(t.Execution),
+		RequestType:        t.RequestType,
+		HistoricalTestData: !ignore,
+		Mainline:           t.Mainline,
+	}, nil
+}
+
+// historicalTestDataIngore checks whether this job's project ignores tasks
```
Kimchelly, line 44: historicalTestDataIgnore

Kimchelly (changes requested):

rpc/internal/test_results_service.go
julianedwards: I was going to get rid of the job entirely in the next  ticket, EVG-14440. Unless you think we should keep the job and just run it inline?

Kimchelly: I also would prefer to get rid of the job, but I don't think we can uncomment this block until we remove the job, or else the deploy will include this code.

Kimchelly (approved):

PR #345, opened 2021-04-20
EVG-14440: Do the historical test data update inline

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-14440

We no longer need the job, but I still want the update to be done in the background so I put it in a go routine so it does not slow down the request.


rpc/internal/test_results_service.go
```diff
@@ -83,16 +84,63 @@ func (s *testResultsService) AddTestResults(ctx context.Context, results *TestRe
...
-			// TODO: Remove this job with EVG-14440.
-			j := units.NewHistoricalTestDataJob(s.env, record.Info, res)
-			j.Run(ctx)
-			grip.Error(message.WrapError(j.Error(), message.Fields{
-				"message":     "failed to run historical test data job",
-				"info":        record.Info,
-				"test_result": res,
-			}))
-		}
+		go func() {
```
Kimchelly, line 87: Can this goroutine function be its own separate function so that `AddTestResults` is less complicated?

Kimchelly, line 87: Is there a concern that this will start too many goroutines since it has to run often? Assuming `Update` is about as performant as the other queries in `AddTestResults` like `Append`, is it necessary to have a separate goroutine for it?

rpc/internal/test_results_service.go
```diff
@@ -83,16 +84,63 @@ func (s *testResultsService) AddTestResults(ctx context.Context, results *TestRe
...
+			defer func() {
+				if err := recovery.HandlePanicWithError(recover(), nil, "historical test data update"); err != nil {
+					grip.Error(message.WrapError(err, message.Fields{
+						"message":           "failed to update historical test data",
+						"test_results_info": record.Info,
+					}))
+				}
+			}()
+
+			htdCtx, cancel := context.WithTimeout(context.Background(), time.Minute)
```
Kimchelly, line 97: Can this be moved to be immediately before `Update`, since that's the only time it's used?

rpc/internal/test_results_service.go
```diff
@@ -83,16 +84,63 @@ func (s *testResultsService) AddTestResults(ctx context.Context, results *TestRe
...
+
+			conf := model.NewCedarConfig(s.env)
+			if err := conf.Find(); err != nil {
+				grip.Error(message.WrapError(errors.Wrap(err, "finding cedar configuration"), message.Fields{
+					"message":           "failed to update historical test data",
+					"test_results_info": record.Info,
+				}))
+				return
+			}
+			if conf.Flags.DisableHistoricalTestData {
```
Kimchelly, line 108: This is effectively dropping data points in the historical test stats. Can this log when stats are disabled and what data point it's dropping?

rpc/internal/test_results_service.go
```diff
@@ -83,16 +84,63 @@ func (s *testResultsService) AddTestResults(ctx context.Context, results *TestRe
...
+				return
+			}
+
+			taskName := record.Info.DisplayTaskName
+			if taskName == "" {
+				taskName = record.Info.TaskName
+			}
```
Kimchelly, line 115: Since we've done this [task name logic before](https://github.com/evergreen-ci/cedar/blob/997a0eeb6ba85cbf227ca91b427e3d476a296afa/rpc/internal/test_results.ext.go#L17-L20), can this task name logic be a method on `TestResultsInfo`, like `HistoricalTaskName()` or something?

rpc/internal/test_results_service_test.go
```diff
@@ -231,6 +232,24 @@ func TestAddTestResults(t *testing.T) {
 					count++
 				}
 				assert.Equal(t, 3, count)
+
+				time.Sleep(time.Second)
```
Kimchelly, line 236: Is there a ticket to investigate why the Pail local bucket needs sleeps in so many of the tests? Or did you already investigate it?

Kimchelly (changes requested):
Can [this comment](https://github.com/evergreen-ci/cedar/blob/997a0eeb6ba85cbf227ca91b427e3d476a296afa/rpc/internal/test_results.ext.go#L44-L45) be updated since there's no longer a job for it?

rpc/internal/test_results_service.go
julianedwards: There will only be one goroutine per call to `AddTestResults`, previously it was blowing up in the queue because it was a job per test result (whereas this is one goroutine for all the test results). I think it should be fine. Since it does have to iterate through potentially many test results, though, I would rather keep it in a goroutine so we do not slow down requests.

rpc/internal/test_results_service_test.go
julianedwards: This is unrelated to the pail local bucket bug (which I fixed in [EVG-14223](https://jira.mongodb.org/browse/EVG-14223)). I put a sleep here cause it needs to wait for the go routine to finish updating the historical test stats.

rpc/internal/test_results_service.go
julianedwards: The other place you referenced is on a different data type (the protobuf's `TestResults` struct, not the `model.TestResults` struct) so it wouldn't actually reduce duplicate code.

rpc/internal/test_results_service.go
julianedwards: That may be excessive logging since this is an application wide flag that would result in nearly every call to `AddTestResults` logging this same thing.

rpc/internal/test_results_service.go
Kimchelly: I still have this same question.

Kimchelly (approved):
LGTM mod question.

rpc/internal/test_results.ext.go
```diff
@@ -41,8 +41,8 @@ func (t TestResultsInfo) Export() (model.TestResultsInfo, error) {
 	}, nil
 }
 
-// historicalTestDataIgnore checks whether this job's project ignores tasks
-// matching this job's task name for historical test data calculations.
+// historicalTestDataIgnore checks whether to given task name matches any of
```
Kimchelly, line 44: "the given task"

Kimchelly (approved):

PR #350, opened 2021-05-04
EVG-14557: User auth over gRPC

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-14557


rest/service.go
```diff
@@ -286,6 +280,7 @@ func (s *Service) addMiddleware() {
 
 func (s *Service) addRoutes() {
 	checkUser := gimlet.NewRequireAuthHandler()
+	depotCheck := NewCertDepotCheckMiddleware(s.Depot == nil)
```
Kimchelly, line 283: nit: can this be `checkDepot` for consistency?

rest/middleware.go
```diff
@@ -14,6 +14,25 @@ import (
...
+
+// NewCertDepotCheckMiddleware returns an implementation of gimlet.Middleware
+// that returns an error if the internal cert depot is disabled.
+func NewCertDepotCheckMiddleware(disabled bool) *certDepotCheckMiddleware {
+	return &certDepotCheckMiddleware{depotDisabled: disabled}
+}
+
+func (m *certDepotCheckMiddleware) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
+	if m.depotDisabled {
+		gimlet.WriteResponse(rw, gimlet.MakeJSONErrorResponder(errors.New("certificate depot disabled")))
```
Kimchelly, line 29: Rather than writing a bad request response, can it return something like 403 forbidden?

rest/middleware.go
```diff
@@ -23,7 +42,7 @@ type evgAuthReadLogByIDMiddleware struct {
 // gimlet.Middleware that sends a http request to Evergreen to check if the
 // user is authorized to read the log they are trying to access based on the
 // given ID.
-func NewEvgAuthReadLogByIDMiddleware(sc data.Connector, evgConf *model.EvergreenConfig) gimlet.Middleware {
+func NewEvgAuthReadLogByIDMiddleware(sc data.Connector, evgConf *model.EvergreenConfig) *evgAuthReadLogByIDMiddleware {
```
Kimchelly, line 45: Does this function) and the one below it need to  be exported? It seems like they're only used within `rest`.

rest/routes.go
```diff
@@ -568,8 +568,11 @@ func (s *Service) fetchUserCertKey(rw http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	if s.Depot == nil {
```
Kimchelly, line 571: Is this supposed to do something or is it handled by the depot-checking middleware?

rpc/service.go
```diff
@@ -21,66 +21,65 @@ import (
...
-	if c.CAName == "" {
-		return errors.New("must specify a CA name")
-	}
-	if c.ServiceName == "" {
-		return errors.New("must specify a service name")
-	}
+func (c *AuthConfig) Validate() error {
+	catcher := grip.NewBasicCatcher()
+
+	catcher.AddWhen(c.TLS && c.UserAuth, errors.New("must specify either TLS or user auth, not both"))
```
Kimchelly, line 37: Should this also check for `!c.TLS && !c.UserAuth` so we don't end up in a state with no authentication?

Kimchelly (changes requested):
In terms of deploying new agents, is it possible for gRPC clients to send both the TLS credentials _and_ the API headers temporarily so that the rollover is backward compatible (at least until agents have rolled over and cedar is still using TLS)?

rpc/service.go
julianedwards: No auth is useful for testing.

rest/routes.go
julianedwards: I forgot to remove it, done.

julianedwards:
I changed the code so that both user auth and TLS can be enabled.

Kimchelly (approved):

PR #351, opened 2021-05-11
EVG-14173: new auth tests

hhoke:
I included separate tests for reading the auth file (which is pretty straightforward) and testing the DB connection after setting up a DB with dummy auth creds (which is less so).


evergreen.yaml
```diff
@@ -132,6 +166,23 @@ tasks:
...
+    commands:
+      - func: get-project
+      - func: set-up-mongodb-auth
+      - func: set-up-curator
+      - func: run-make
+        vars: { target: "cedar" }
+      - func: run-make
+        vars: { target: "clean-results" }
+      - func: run-make
+        vars: { target: "test-cedar", RUN_TEST: "TestGlobalEnvironmentAuth", TEST_USER_ADMIN: "myUserAdmin"}
```
julianedwards, line 179: Can you format the same way you formatted the `vars` in `self-tests.yaml` in evg?

julianedwards (changes requested):

evergreen.yaml
hhoke: Yeah good call

julianedwards (approved):

PR #352, opened 2021-05-13
EVG-14514: Parallelize uploading of test results to s3

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-14514

### DIFF
--- a/model/test_results.go
+++ b/model/test_results.go
@@ -194,8 +194,10 @@ func (t *TestResults) Append(ctx context.Context, results []TestResult) error {
 			for result := range resultChan {
 				data, err := bson.Marshal(result)
 				if err != nil {
+					// We should continue on marhsalling
+					// errors so we can get as many test
+					// results in s3 as possible.
 					catcher.Wrapf(err, "problem marshalling bson for test result '%s'", result.TestName)
-					return
 				}
 
 				if err := bucket.Put(ctx, result.TestName, bytes.NewReader(data)); err != nil {
--- a/model/test_results.go
+++ b/model/test_results.go
@@ -194,11 +194,13 @@ func (t *TestResults) Append(ctx context.Context, results []TestResult) error {
 			for result := range resultChan {
 				data, err := bson.Marshal(result)
 				if err != nil {
-					catcher.Wrap(err, "problem marshalling bson")
+					catcher.Wrapf(err, "problem marshalling bson for test result '%s'", result.TestName)
+					return
 				}
 
 				if err := bucket.Put(ctx, result.TestName, bytes.NewReader(data)); err != nil {
-					catcher.Wrap(err, "problem uploading test results to bucket")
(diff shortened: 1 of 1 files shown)


model/test_results.go
```diff
@@ -173,18 +176,36 @@ func (t *TestResults) Append(ctx context.Context, results []TestResult) error {
...
+			}()
+
+			for result := range resultChan {
+				data, err := bson.Marshal(result)
+				if err != nil {
+					catcher.Wrap(err, "problem marshalling bson")
+				}
+
+				if err := bucket.Put(ctx, result.TestName, bytes.NewReader(data)); err != nil {
+					catcher.Wrap(err, "problem uploading test results to bucket")
```
Kimchelly, line 201: Should this include the test name for context?

model/test_results.go
```diff
@@ -173,18 +176,36 @@ func (t *TestResults) Append(ctx context.Context, results []TestResult) error {
...
+		go func() {
+			defer func() {
+				catcher.Add(recovery.HandlePanicWithError(recover(), nil, "test results append"))
+				wg.Done()
+			}()
+
+			for result := range resultChan {
+				data, err := bson.Marshal(result)
+				if err != nil {
+					catcher.Wrap(err, "problem marshalling bson")
```
Kimchelly, line 197: Can this include the test name for context?

Kimchelly (changes requested):

model/test_results.go
```diff
@@ -194,11 +194,13 @@ func (t *TestResults) Append(ctx context.Context, results []TestResult) error {
 			for result := range resultChan {
 				data, err := bson.Marshal(result)
 				if err != nil {
-					catcher.Wrap(err, "problem marshalling bson")
+					catcher.Wrapf(err, "problem marshalling bson for test result '%s'", result.TestName)
+					return
```
Kimchelly, line 198: Do we actually want to return in this case? It seems like this would be a continue-on-error case so that it can post as many test results as it possibly can without stopping early.

julianedwards: True, we were returning in the other case. But that makes sense.

model/test_results.go
```diff
@@ -194,8 +194,10 @@ func (t *TestResults) Append(ctx context.Context, results []TestResult) error {
 			for result := range resultChan {
 				data, err := bson.Marshal(result)
 				if err != nil {
+					// We should continue on marhsalling
+					// errors so we can get as many test
+					// results in s3 as possible.
 					catcher.Wrapf(err, "problem marshalling bson for test result '%s'", result.TestName)
```
Kimchelly, line 200: I actually meant `continue` here, because the data is most likely invalid/nil if `bson.Marshal` fails, so putting bad data in the S3 bucket won't be helpful (or might panic).

julianedwards: Sorry my head is fried after grpc debugging, I forgot to add the continue lol

Kimchelly (approved):

PR #359, opened 2021-06-08
EVG-14775: Force close gRPC service if cannot gracefully stop

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-14775

### DIFF
--- a/rpc/service.go
+++ b/rpc/service.go
@@ -6,6 +6,7 @@ import (
 	"crypto/x509"
 	"fmt"
 	"net"
+	"time"
 
 	"github.com/evergreen-ci/aviation"
 	"github.com/evergreen-ci/cedar"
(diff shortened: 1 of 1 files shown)


rpc/service.go
```diff
@@ -161,7 +162,23 @@ func RunServer(ctx context.Context, srv *grpc.Server, addr string) (WaitFunc, er
...
 		<-ctx.Done()
-		srv.GracefulStop()
+
+		gracefulStop := make(chan struct{})
+		go func() {
+			defer close(gracefulStop)
+			srv.GracefulStop()
+		}()
+
+		timer := time.NewTimer(2 * time.Minute)
```
bsamek, line 172: I think this is a problem https://github.com/10gen/kernel-tools/blob/ad3c80badf498c165c309d50e0a0bc500be36a9d/evergreen/cedar.service#L10. It should probably match Evergreen's https://github.com/10gen/kernel-tools/blob/ad3c80badf498c165c309d50e0a0bc500be36a9d/evergreen/evg-web.service#L10. Could you open a ticket or make that change under this ticket?

Kimchelly (approved):
LGTM + we should also change the systemd config.

rpc/service.go
julianedwards: I just put up a PR! Does this mean that we actually don't need to merge this code, since the timeout will be taken care of by systemd?

PR #362, opened 2021-06-21
EVG-14672: store test results more efficiently

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-14672

This stores test results as a single file in the respective s3 bucket location for the results. There really is no need to store results individually since they are tiny and the bottleneck ends up being network calls to s3.

There will be a follow up ticket to create a migration job for existing results stored the old way.

### DIFF
--- a/model/test_results.go
+++ b/model/test_results.go
@@ -7,8 +7,7 @@ import (
 	"fmt"
 	"hash"
 	"io"
-	"runtime"
-	"sync"
+	"io/ioutil"
 	"time"
 
 	"github.com/evergreen-ci/cedar"
@@ -17,7 +16,6 @@ import (
 	"github.com/mongodb/anser/db"
 	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/message"
-	"github.com/mongodb/grip/recovery"
 	"github.com/pkg/errors"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
@@ -145,7 +143,7 @@ func (t *TestResults) Remove(ctx context.Context) error {
 // environment should not be nil.
 func (t *TestResults) Append(ctx context.Context, results []TestResult) error {
 	if !t.populated {
-		return errors.New("cannot append with populated test results")
+		return errors.New("cannot append without populated test results")
 	}
 	if t.env == nil {
 		return errors.New("cannot not append test results with a nil environment")
@@ -162,7 +160,7 @@ func (t *TestResults) Append(ctx context.Context, results []TestResult) error {
 	conf := &CedarConfig{}
 	conf.Setup(t.env)
 	if err := conf.Find(); err != nil {
-		return errors.Wrap(err, "problem getting application configuration")
+		return errors.Wrap(err, "getting application configuration")
 	}
 	bucket, err := t.Artifact.Type.Create(
 		ctx,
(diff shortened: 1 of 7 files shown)


model/test_results.go
```diff
@@ -173,50 +171,43 @@ func (t *TestResults) Append(ctx context.Context, results []TestResult) error {
...
+			return errors.Wrap(err, "reading uploaded test results")
+		}
+
+		if err = bson.Unmarshal(data, &allResults); err != nil {
+			return errors.Wrap(err, "unmarshalling uploaded test results")
+		}
+	}
+	allResults.Results = append(allResults.Results, results...)
+
+	data, err := bson.Marshal(&allResults)
```
ybrill, line 196: How large could the file get? If it's really big, would we consider using [gob](https://golang.org/pkg/encoding/gob/) to encode the results? Some [people on the internet](https://stackoverflow.com/a/41180680) seem to think it's more efficient.

bsamek:
@julianedwards @ybrill I saw a couple proposals in this PR and wanted to push back against doing them, or at least against doing them soon.

> There will be a follow up ticket to create a migration job for existing results stored the old way.

Is this important? Users are much less likely to view test results that are older than a few days. Since (I think) those test results will still load, but slowly, I think it's not important that we migrate.

> How large could the file get? If it's really big, would we consider using gob to encode the results? Some people on the internet seem to think it's more efficient.

Given that we've optimized away the current bottleneck (many small s3 network requests), is this additional optimization really important? I agree it could be useful to research this, but I'd suggest waiting until after this is deployed. If it's fast(ish) to do it this way, I think it's better to get the work deployed. I'd caution against trying to speed up a PR that is already speeding something up, since the additional work on the PR can slow delivery of value.

Kimchelly:
@bsamek We could just not migrate them, leave this old code in for a while as a fall-back case when the test results aren't in the new format, and make a TODO + Jira ticket to remove this code once the TTL expires on the last test results stored in the old format.

It seems like we're not certain exactly what format we want test results to be in and it could change. Can we version them (e.g. add a simple version suffix to the object key or something) and use the version field in the `TestResultsArtifactInfo` to keep track of which one it is?

model/test_results.go
Kimchelly: I have this same question. The serialize/update/deserialize roundtrip seems okay but only if it won't take up a ton of memory for large test results.

bsamek:
@Kimchelly That sounds reasonable. We would need to support both formats for some period of time even if we were doing a migration soon. My goal here is to avoid blocking work on something we need not do right away.

model/test_results.go
julianedwards: Given the size of a single test result is < 500B, even if a task has 10k or even 20k results that will be no more than 10 MB (the same size as our logging buffer which is held in memory).

julianedwards:
@bsamek I am fine with not doing the migration, especially if this old data will not really be used.

model/test_results.go
```diff
@@ -173,50 +171,38 @@ func (t *TestResults) Append(ctx context.Context, results []TestResult) error {
...
+	}
+
+	return nil
 }
 
 // Download returns a TestResult slice with the corresponding results stored in
 // the offline blob storage. The TestResults should be populated and the
 // environment should not be nil.
-func (t *TestResults) Download(ctx context.Context) (TestResultsIterator, error) {
+func (t *TestResults) Download(ctx context.Context) ([]TestResult, error) {
```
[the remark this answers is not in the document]
julianedwards: Yeah, since the results should cap around 10MB (most being much smaller) it will actually be very efficient. The current bottleneck is making thousands of individual requests to s3 when fetching each test result stored separately. We are not currently paginating, but that may occur in a follow-up ticket. I need to continue working with UI as they develop new features.

Kimchelly (commented):
If we're not migrating the old test results, don't we still have to support the old chunked test results format? Just because we don't migrate the old data doesn't mean that they should be inaccessible. That's why I suggested incrementing the version.

julianedwards:
Download falls back to using the Iterator (which uses the old chunk methods) if it can't find the single test results doc in s3.

Kimchelly (approved):

Kimchelly (approved):
Bless versioned data.

PR #365, opened 2021-06-24
EVG-14386 Error on missing cert depot

bsamek:


Kimchelly (commented):
The certificate depot should no longer be required in Cedar since it's no longer managing its own TLS credentials.

bsamek:
@Kimchelly Can this ticket close as won't fix then?

Kimchelly:
Yeah, we should make a ticket to remove certdepot at some point, but in the meantime, it's fine as it is.

PR #371, opened 2021-08-20
EVG-15254: Add REST route to fetch failing tests sample

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-15254

### DIFF
--- a/rest/data/test_results.go
+++ b/rest/data/test_results.go
@@ -190,7 +190,7 @@ func (mc *MockConnector) findTestResultsByDisplayTaskID(ctx context.Context, opt
 		testResultsDocs = filteredDocs
 	}
 
-	if testResultsDocs == nil {
+	if len(testResultsDocs) == 0 {
 		return nil, gimlet.ErrorResponse{
 			StatusCode: http.StatusNotFound,
 			Message:    "test results not found",
@@ -263,7 +263,7 @@ func importTestResults(ctx context.Context, results []dbModel.TestResult) ([]mod
 
 func extractFailedTestResultsSample(results ...dbModel.TestResults) []string {
 	var sample []string
-	for i := 0; i < len(results) && len(sample) < 10; i++ {
+	for i := 0; i < len(results) && len(sample) < dbModel.FailedTestsSampleSize; i++ {
 		sample = append(sample, results[i].FailedTestsSample...)
 	}
 
--- a/rest/data/test_results.go
+++ b/rest/data/test_results.go
@@ -33,6 +33,23 @@ func (dbc *DBConnector) FindTestResults(ctx context.Context, opts TestResultsOpt
 	return importTestResults(ctx, results)
 }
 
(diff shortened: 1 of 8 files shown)


rest/data/test_results.go
```diff
@@ -194,6 +261,15 @@ func importTestResults(ctx context.Context, results []dbModel.TestResult) ([]mod
 	return apiResults, nil
 }
 
+func extractFailedTestResultsSample(results ...dbModel.TestResults) []string {
+	var sample []string
+	for i := 0; i < len(results) && len(sample) < 10; i++ {
```
Kimchelly, line 266: Should the `failedTestsSampleSize` be exported and used here?

rest/data/test_results.go
```diff
@@ -56,81 +73,131 @@ func (dbc *DBConnector) FindTestResultByTestName(ctx context.Context, opts TestR
...
 	}
 
-	for i := range testResults {
-		if !opts.EmptyExecution || testResults[i].Info.Execution == latestExecution {
-			results, err := testResults[i].Download(ctx)
-			if err != nil {
-				return nil, err
-			}
-			combinedResults = append(combinedResults, results...)
+	if testResultsDocs == nil {
```
Kimchelly, line 193: Should this check if `len(testResultsDocs) == 0` to cover both nil and empty slice cases?

rest/test_results_routes.go
```diff
@@ -174,3 +176,68 @@ func (h *testResultGetByTestNameHandler) Run(ctx context.Context) gimlet.Respond
...
+// Factory returns a pointer to a new testResultsGetFailedSampleHandler.
+func (h *testResultsGetFailedSampleHandler) Factory() gimlet.RouteHandler {
+	return &testResultsGetFailedSampleHandler{
+		sc: h.sc,
+	}
+}
+
+// Parse fetches the task_id from the http request.
+func (h *testResultsGetFailedSampleHandler) Parse(_ context.Context, r *http.Request) error {
+	var err error
```
Kimchelly, line 204: nit: I would declare `err` closer to where you use it (L214).

rest/test_results_routes_test.go
```diff
@@ -316,6 +318,96 @@ func (s *TestResultsHandlerSuite) TestTestResultGetByTestNameHandlerCtxErr() {
 	s.Equal(http.StatusInternalServerError, resp.Status())
 }
 
+func (s *TestResultsHandlerSuite) TestTestResultsGetFailedSampleFound() {
+	rh := s.rh["failed_tests_sample"]
```
Kimchelly, line 322: In these tests, it seems like you always type assert that it's a testResultsGetFailedSampleHandler. It seems like you could just type assert it here instead of doing repeated type asserts.

Kimchelly (changes requested):

Kimchelly (approved):

PR #372, opened 2021-08-23
EVG-15269: Ensure that the update result is not nil before checking it

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-15269


model/perf_data.go
```diff
@@ -169,9 +169,10 @@ func tryUpdate(ctx context.Context, collection *mongo.Collection, id string, r P
 					},
 				},
 			},
-		}))
+		}),
+	)
 
-	return res.MatchedCount == 1, errors.WithStack(err)
+	return updateResult != nil && updateResult.MatchedCount == 1, errors.WithStack(err)
```
Kimchelly, line 175: nit: would it be simpler to do this?
```go
if err != nil {
    return nil, errors.WithStack(err)
}
return updateResult.MatchedCount == 1, nil
```

Kimchelly (approved):
LGTM mod suggestion.

PR #378, opened 2021-08-30
EVG-15354: Make sure time range is valid when fetching grouped logs

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-15354

This occurs when a test log was not able to close out so the `completed_at` field is zero, for example, when a resmoke test timeouts.

### DIFF
--- a/rest/buildlogger_routes.go
+++ b/rest/buildlogger_routes.go
@@ -555,6 +555,9 @@ func (h *logGroupByTestNameHandler) Run(ctx context.Context) gimlet.Responder {
 				h.opts.TimeRange.EndAt = time.Time(log.CompletedAt)
 			}
 		}
+		if h.opts.TimeRange.EndAt.IsZero() {
+			h.opts.TimeRange.EndAt = time.Now()
+		}
 	}
 
 	data, next, paginated, err := h.sc.FindGroupedLogs(ctx, h.opts)


ybrill (commented):
could an alternative be to only set the `logCreatedAtKey` in [the query](https://github.com/evergreen-ci/cedar/blob/5946f953bacce9d7604820e6fefec361a66db92b/model/buildlogger.go#L478) if  `!opts.TimeRange.EndAt.IsZero()`

julianedwards:
> could an alternative be to only set the `logCreatedAtKey` in [the query](https://github.com/evergreen-ci/cedar/blob/5946f953bacce9d7604820e6fefec361a66db92b/model/buildlogger.go#L478) if `!opts.TimeRange.EndAt.IsZero()`

The find function checks that the time range is either zero or valid, which requires `EndAt` to be greater than `StartAt`. I think it is clearer to explicitly set it to a valid time range.

ybrill (approved):
sgtm!

julianedwards:
Also, this query actually doesn't use the time range for finding the logs in the DB, rather it uses it for the reader (filtering logs before and after the range).

When `taskID` is set, the range query is removed: https://github.com/evergreen-ci/cedar/blob/main/model/buildlogger.go#L493-L496

ybrill:
oh interesting. 
So the fix here is really just to get the query through the [validation](https://github.com/evergreen-ci/cedar/blob/32e8b7a079bd8ee011244017ef027e6d36efa2fa/model/buildlogger.go#L425) you mentioned, then.

PR #380, opened 2021-09-02
EVG-15255 new update time series job

ybrill:
[EVG-15255](https://jira.mongodb.org/browse/EVG-15255)

If the report contains rollups with no artifacts we need to enqueue a time series update right away.

### DIFF
--- a/rpc/internal/perf_service.go
+++ b/rpc/internal/perf_service.go
@@ -56,6 +56,14 @@ func (srv *perfService) CreateMetricSeries(ctx context.Context, result *ResultDa
 	if err := record.SaveNew(ctx); err != nil {
 		return resp, newRPCError(codes.Internal, errors.Wrap(err, "problem saving record"))
 	}
+	if len(record.Rollups.Stats) > 0 {
+		processingJob := units.NewUpdateTimeSeriesJob(record.Info.ToPerformanceResultSeriesID())
+		err := amboy.EnqueueUniqueJob(ctx, srv.env.GetRemoteQueue(), processingJob)
+
+		if err != nil {
+			return nil, newRPCError(codes.Internal, errors.Wrapf(err, "problem creating signal processing job for perf result '%s'", record.ID))
+		}
+	}
 
 	if err := srv.addFTDCRollupsJob(ctx, record.ID, record.Artifacts); err != nil {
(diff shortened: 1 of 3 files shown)


rpc/internal/perf_service.go
```diff
@@ -56,6 +56,14 @@ func (srv *perfService) CreateMetricSeries(ctx context.Context, result *ResultDa
 	if err := record.SaveNew(ctx); err != nil {
 		return resp, newRPCError(codes.Internal, errors.Wrap(err, "problem saving record"))
 	}
+	if len(record.Rollups.Stats) > 0 && len(record.Artifacts) == 0 {
```
julianedwards, line 59: In theory someone could add a non-ftdc artifact and rollups, which, with this if statement, would never create an update-time-series job. It's fine to call signal processing more than once with updated data, so this should just be `if len(record.Rollups.Stats) > 0`.

julianedwards (changes requested):

julianedwards (approved):

rpc/internal/perf_service.go
ybrill: done.
Something is confusing me about how it works: what prevents us from enqueueing the FTDC rollups job for non-FTDC artifacts? Is it [this schema check](https://github.com/evergreen-ci/cedar/blob/32e8b7a079bd8ee011244017ef027e6d36efa2fa/rpc/internal/perf_service.go#L255)? In other words, is the raw-events schema particular to FTDC artifacts?

julianedwards: The schema `raw-events` represents a (performance point)[https://github.com/mongodb/ftdc/blob/main/events/performance.go] which cedar understands in [perf/ftdc_rollups.go](https://github.com/evergreen-ci/cedar/blob/main/perf/ftdc_rollups.go). Any data with another schema passed to the rollups job would fail. 

Though, I just realized that we do not check the file format in (addFTDCRollupsJob)[https://github.com/evergreen-ci/cedar/blob/main/rpc/internal/perf_service.go#L249]—this currently is not a problem cause no one is attaching non-ftdc files. But there should be a check that `artifact.Format == model.FormatFTDC`.

ybrill: I opened https://jira.mongodb.org/browse/EVG-15372

PR #382, opened 2021-09-03
EVG-15372 add FTDC format check

ybrill:
Implement https://github.com/evergreen-ci/cedar/pull/380#discussion_r701953667

### DIFF
--- a/model/perf.go
+++ b/model/perf.go
@@ -648,7 +648,7 @@ func (r *PerformanceResults) FindOutdatedRollups(ctx context.Context, name strin
 
 	search := bson.M{
 		perfCreatedAtKey: bson.M{"$gt": after},
-		bsonutil.GetDottedKeyName(perfArtifactsKey, artifactInfoFormatKey): FileFTDC,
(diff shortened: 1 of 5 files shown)


julianedwards (approved):
LGTM % fixing the tests (probably just need to set the artifact format to FTDC).

julianedwards:
Actually, let's hold off on merging this. It looks like artifact formats are getting set incorrectly to `text` instead of `ftdc`.

ybrill:
> artifact formats are getting set incorrectly to text instead of ftdc

@julianedwards Should I open a TIPS ticket for the file type to be set correctly? Or should we just leave it

julianedwards:
I spoke to Alex, I think the best thing to do for now is change this [line](https://github.com/evergreen-ci/cedar/blob/main/model/perf.go#L651) to check for raw events schema instead of type, as done [here](https://github.com/evergreen-ci/cedar/blob/main/rpc/internal/perf_service.go#L263).

julianedwards (approved):

PR #383, opened 2021-09-15
EVG-14548 summarize stats

ybrill:
[EVG-14548](https://jira.mongodb.org/browse/EVG-14548)

Log counts to Splunk once a minute for buildlogger, testresults, and perf.


julianedwards (changes requested):
I am nervous that the version and project counts will blow up the logging, though they are probably useful. Maybe we deploy with them and see what happens?

Also, I wonder if this will slow down requests since there will be lots of contention to get the mutex. Maybe it would be more beneficial to use buffered channel?

model/stats_cache.go
```diff
@@ -0,0 +1,148 @@
...
+		LinesByVersion: make(map[string]int),
+		LinesByProject: make(map[string]int),
+	}
+}
+
+func (b *buildloggerStatsCache) LogStats() {
+	b.Lock.Lock()
+	defer b.Lock.Unlock()
+
+	grip.InfoWhen(b.TotalLines > 0, message.Fields{
```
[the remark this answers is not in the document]
ybrill: that makes sense.
done

model/stats_cache.go
```diff
@@ -0,0 +1,148 @@
...
+
+	CacheRegistry = append(CacheRegistry, buildLoggerCache)
+	CacheRegistry = append(CacheRegistry, testResultsCache)
+	CacheRegistry = append(CacheRegistry, perfCache)
+}
+
+type buildloggerStatsCache struct {
+	Lock sync.Mutex
+
+	TotalLines     int
```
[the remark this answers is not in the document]
ybrill: You're right.
I un-exported it

model/stats_cache.go
```diff
@@ -0,0 +1,148 @@
...
+	testResultsCache = newTestResultsStatsCache()
+	perfCache = newPerfStatsCacheStatsCache()
+
+	CacheRegistry = append(CacheRegistry, buildLoggerCache)
+	CacheRegistry = append(CacheRegistry, testResultsCache)
+	CacheRegistry = append(CacheRegistry, perfCache)
+}
+
+type buildloggerStatsCache struct {
+	Lock sync.Mutex
```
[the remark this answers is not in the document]
ybrill: done

model/stats_cache.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/mongodb/grip/message"
+)
+
+var CacheRegistry []StatsCache
+
+type StatsCache interface {
+	LogStats()
+}
+
+var buildLoggerCache *buildloggerStatsCache
```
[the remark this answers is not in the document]
ybrill: done

model/stats_cache.go
```diff
@@ -0,0 +1,148 @@
...
+import (
+	"sync"
+
+	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/message"
+)
+
+var CacheRegistry []StatsCache
+
+type StatsCache interface {
```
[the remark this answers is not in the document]
ybrill: done

model/stats_cache.go
```diff
@@ -0,0 +1,148 @@
+package model
+
+import (
+	"sync"
+
+	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/message"
+)
+
+var CacheRegistry []StatsCache
```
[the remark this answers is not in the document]
ybrill: done

units/stats_counters.go
```diff
@@ -0,0 +1,55 @@
...
+func init() {
+	registry.AddJobType(statsCacheLoggerJobName,
+		func() amboy.Job { return makeStatsCacheLogger() })
+}
+
+type statsCacheLoggerJob struct {
+	job.Base `bson:"job_base" json:"job_base" yaml:"job_base"`
+}
+
+// NewLocalAmboyStatsCollector reports the status of only the local queue
```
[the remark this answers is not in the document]
ybrill: done

cedar.go
```diff
@@ -35,6 +35,11 @@ const (
 	TriggerRequester            = "trigger_request"
 	MergeTestRequester          = "merge_test"
 	AdHocRequester              = "ad_hoc"
+
+	// stats cache names
```
julianedwards, line 39: `Stats cache names.`

environment.go
```diff
@@ -232,6 +234,9 @@ type Environment interface {
 	GetServerCertVersion() int
 	SetServerCertVersion(i int)
 
+	// GetStatsCache returns the cache corresponding to the string
```
julianedwards, line 237: Missing period.

model/buildlogger.go
```diff
@@ -226,6 +227,19 @@ func (l *Log) appendLogChunkInfo(ctx context.Context, logChunk LogChunkInfo) err
...
 		"op":           "append log chunk info to buildlogger log",
 	})
+
+	if err = l.env.GetStatsCache(cedar.StatsCacheBuildlogger).AddStat(cedar.Stat{
+		Count:   logChunk.NumLines,
+		Project: l.Info.Project,
+		Version: l.Info.Version,
+		Task:    l.Info.TaskID,
+	}); err != nil {
+		grip.ErrorWhen(sometimes.Percent(10), message.WrapError(err, message.Fields{
```
julianedwards, line 237: Since we do not expect these errors often, I think it's better to always log when an error occurs. It may help us diagnose issues or, if this is erroring frequently, realize that our stats probably aren't very useful since so many are missing.

stats_cache.go
```diff
@@ -0,0 +1,125 @@
+package cedar
```
julianedwards, line 1: Add documentation in this file.

stats_cache.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	return registry
+}
+
+type Stat struct {
+	Count   int
+	Project string
+	Version string
+	Task    string
```
julianedwards, line 33: I would change this to `TaskID` to be more specific, since we also have `TaskName`. It may be useful to add `TaskName`.

stats_cache.go
```diff
@@ -0,0 +1,125 @@
...
+func (s *statsCache) cacheStat(newStat Stat) {
+	s.calls++
+	s.total += newStat.Count
+	s.byProject[newStat.Project] += newStat.Count
+	s.byVersion[newStat.Version] += newStat.Count
+	s.byTask[newStat.Task] += newStat.Count
+}
+
+func (s *statsCache) startConsumerLoop(ctx context.Context) {
+	for {
```
julianedwards, line 73: Add a defer statement to recover from panics here.

units/crons.go
```diff
@@ -68,7 +68,10 @@ func StartCrons(ctx context.Context, env cedar.Environment, rpcTLS bool) error {
 			return nil
 		}
 
-		return queue.Put(ctx, NewRemoteAmboyStatsCollector(env, utility.RoundPartOfMinute(0).Format(tsFormat)))
+		catcher := grip.NewBasicCatcher()
+		catcher.Add(queue.Put(ctx, NewRemoteAmboyStatsCollector(env, utility.RoundPartOfMinute(0).Format(tsFormat))))
+		catcher.Add(queue.Put(ctx, NewStatsCacheLogger(env, utility.RoundPartOfMinute(0).Format(tsFormat))))
```
julianedwards, line 73: The issue with creating this as an amboy job is that any of the 12 servers can pick it up. So it becomes very easy where we run into a case where these stats are not being flushed frequently enough on each server and we start dropping many useful stats. I would propose instead of using an amboy job to create a background thread. We could also then just move the stats collection entirely into the `rpc` package (since it is specific to that). What are your thoughts?

environment.go
ybrill: done

units/crons.go
ybrill: That's a good point. I moved it into a loop that gets started together with the consumer loop.
I like the location of the stats cache itself: I was aiming to consolidate global state to just the environment. I'm agnostic as to adding the stats from the rpc service vs. the model.

stats_cache.go
```diff
@@ -0,0 +1,163 @@
...
+			s.cacheStat(nextStat)
+		}
+	}
+}
+
+func (s *statsCache) startLoggerLoop(ctx context.Context) {
+	defer func() {
+		if err := recovery.HandlePanicWithError(recover(), nil, "stats cache logger"); err != nil {
+			grip.Error(message.WrapError(err, message.Fields{
+				"message": "panic in statsCache logger loop",
```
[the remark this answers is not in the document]
ybrill: done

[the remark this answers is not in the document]
ybrill: you're right! done

model/buildlogger.go
```diff
@@ -196,7 +196,35 @@ func (l *Log) Append(ctx context.Context, lines []LogLine) error {
...
+
+	l.addToStatsCache(lines)
+
+	return nil
+}
+
+func (l *Log) addToStatsCache(lines []LogLine) {
+	linesLen := 0
+	for _, line := range lines {
+		linesLen += len(line.Data)
```
julianedwards, line 211: Nit: `linesSize` or something to not confuse the reader with the number of log lines since `len` can be used to get the number lines in the `LogLine` slice.

julianedwards (changes requested):
LGTM % nit

PR #389, opened 2021-09-28
EVG-15263: Return test results with stats and base status (if applicable)

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-15263

I tested that this is backwards compatible in staging.


rest/model/test_results.go
```diff
@@ -61,8 +71,9 @@ func (a *APITestResult) Import(i interface{}) error {
 
 // APITTestResultsStats describes basic stats for a group of test results.
 type APITestResultsStats struct {
-	TotalCount  int `json:"total_count"`
-	FailedCount int `json:"failed_count"`
+	TotalCount    int  `json:"total_count"`
+	FailedCount   int  `json:"failed_count"`
+	FilteredCount *int `json:"filtered_count,omitempty"`
```
Kimchelly, line 76: Is this omitempty because it only applies in the pagination case?

Kimchelly (approved):
LGTM mod question.

rest/model/test_results.go
julianedwards: Yeah, if using the `/stats` route it doesn't make sense to have a `FilteredCount`.

PR #390, opened 2021-09-30
EVG-14837: Create cedar route to determine if perf data exists for a task

hhoke:
This sends a shortened message iff perf results exists, otherwise returning a 404 error.

### DIFF
--- a/rest/perf_routes.go
+++ b/rest/perf_routes.go
@@ -191,7 +191,7 @@ func (h *perfCountByTaskIdHandler) Parse(_ context.Context, r *http.Request) err
 }
 
 // Run calls the data FindPerformanceResults function and returns the
-// number of PerformanceResults fount
+// number of PerformanceResults found
 func (h *perfCountByTaskIdHandler) Run(ctx context.Context) gimlet.Responder {
 	perfResults, err := h.sc.FindPerformanceResults(ctx, h.opts)
 	if err != nil {
--- a/rest/perf_routes.go
+++ b/rest/perf_routes.go
@@ -163,36 +163,36 @@ func (h *perfGetByTaskIdHandler) Run(ctx context.Context) gimlet.Responder {
 
 ///////////////////////////////////////////////////////////////////////////////
 //
-// GET /perf/task_id/{task_id}/exists
+// GET /perf/task_id/{task_id}/count
 
-type perfExistsByTaskIdHandler struct {
+type perfCountByTaskIdHandler struct {
 	opts data.PerformanceOptions
 	sc   data.Connector
 }
 
-func makeExistsPerfByTaskId(sc data.Connector) gimlet.RouteHandler {
-	return &perfExistsByTaskIdHandler{
+func makeCountPerfByTaskId(sc data.Connector) gimlet.RouteHandler {
+	return &perfCountByTaskIdHandler{
 		sc: sc,
 	}
 }
 
-// Factory returns a pointer to a new perfGetByTaskIdHandler.
-func (h *perfExistsByTaskIdHandler) Factory() gimlet.RouteHandler {
-	return &perfExistsByTaskIdHandler{
+// Factory returns a pointer to a new perfCountByTaskIdHandler.
+func (h *perfCountByTaskIdHandler) Factory() gimlet.RouteHandler {
+	return &perfCountByTaskIdHandler{
 		sc: h.sc,
 	}
 }
 
 // Parse fetches the task_id from the http request.
-func (h *perfExistsByTaskIdHandler) Parse(_ context.Context, r *http.Request) error {
+func (h *perfCountByTaskIdHandler) Parse(_ context.Context, r *http.Request) error {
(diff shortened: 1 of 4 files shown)


rest/perf_routes.go
```diff
@@ -161,6 +161,60 @@ func (h *perfGetByTaskIdHandler) Run(ctx context.Context) gimlet.Responder {
...
+		err = errors.Wrapf(err, "problem getting performance results by task id '%s'", h.opts.TaskID)
+		grip.Error(message.WrapError(err, message.Fields{
+			"request": gimlet.GetRequestID(ctx),
+			"method":  "GET",
+			"route":   "/perf/task_id/{task_id}/exists",
+			"task_id": h.opts.TaskID,
+		}))
+		return gimlet.MakeJSONErrorResponder(err)
+	}
+	if len(perfResults) > 0 {
```
Kimchelly, line 207: nit: normally, early return paths are for the unhappy case (i.e. perf results don't exist) and the "main code path" (i.e. the one without the if statements) is for the happy path. Can you invert L207-215 so that it reads more like this?
```
if len(perfResults) == 0 {
    // Return because we've hit the unhappy path.
    return <error response>
}

// Return the real result in the happy path.
return gimlet.NewJSONResponse(result)
```

rest/perf_routes.go
```diff
@@ -161,6 +161,60 @@ func (h *perfGetByTaskIdHandler) Run(ctx context.Context) gimlet.Responder {
...
+		existResult := model.APIPerformanceResultExists{
+			Name:   h.opts.TaskID,
```
Kimchelly, line 209: I don't think this is necessarily wrong, but why does this have to return the task ID when the caller already includes the task ID in the route parameter? It just seems superfluous.

Kimchelly (changes requested):
1. I think you forgot to register your route in `rest/service.go`.
2. Can you write tests?
3. nit: the PR title needs a space after the ticket number.

rest/perf_routes.go
hhoke: I'll return the number of results instead

rest/perf_routes.go
```diff
@@ -204,15 +204,16 @@ func (h *perfExistsByTaskIdHandler) Run(ctx context.Context) gimlet.Responder {
 		}))
 		return gimlet.MakeJSONErrorResponder(err)
 	}
-	if len(perfResults) > 0 {
-		existResult := model.APIPerformanceResultExists{
-			Name:   h.opts.TaskID,
-			Exists: true,
-		}
-		return gimlet.NewJSONResponse(existResult)
+	if len(perfResults) == 0 {
```
Kimchelly, line 207: The way this is written to return the number of results, it seems like 0 is just a special case and not necessarily an error. Would it be more reasonable to just return no perf results in the same way as non-zero perf results and let the client (in this case, Evergreen) deal with the 0 results case however it wants?

rest/service.go
```diff
@@ -305,6 +305,7 @@ func (s *Service) addRoutes() {
 	s.app.AddRoute("/perf/{id}").Version(1).Delete().Wrap(checkUser).RouteHandler(makeRemovePerfById(s.sc))
 	s.app.AddRoute("/perf/children/{id}").Version(1).Get().RouteHandler(makeGetPerfChildren(s.sc))
 	s.app.AddRoute("/perf/task_id/{task_id}").Version(1).Get().RouteHandler(makeGetPerfByTaskId(s.sc))
+	s.app.AddRoute("/perf/task_id/{task_id}/exists").Version(1).Get().RouteHandler(makeExistsPerfByTaskId(s.sc))
```
Kimchelly, line 308: Since the route now returns a count instead of an exist boolean, should this route be renamed to be /count instead of /exists?

Kimchelly (changes requested):

rest/perf_routes.go
hhoke: Agreed, that would be more general

rest/service.go
hhoke: absolutely

rest/perf_routes.go
```diff
@@ -163,36 +163,36 @@ func (h *perfGetByTaskIdHandler) Run(ctx context.Context) gimlet.Responder {
...
+func (h *perfCountByTaskIdHandler) Parse(_ context.Context, r *http.Request) error {
 	h.opts.TaskID = gimlet.GetVars(r)["task_id"]
 	h.opts.Tags = r.URL.Query()[tags]
 	return nil
 }
 
 // Run calls the data FindPerformanceResults function and returns the
-// PerformanceResults from the provider.
-func (h *perfExistsByTaskIdHandler) Run(ctx context.Context) gimlet.Responder {
+// number of PerformanceResults fount
```
Kimchelly, line 194: number of PerformanceResults found.

Kimchelly (approved):
LGTM mod nit.

PR #391, opened 2021-10-01
EVG-15514: Marshal the PerformanceArguments type to the BSON null value when empty

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-15514

This is causing a bug where we try to look up perf results with no args but instead pass in `info.args: {}` to the query.

### DIFF
--- a/model/perf.go
+++ b/model/perf.go
@@ -348,7 +348,7 @@ func (args PerformanceArguments) MarshalBSON() ([]byte, error) {
 	// If the arguments map is empty or only has one item, we can just use
 	// the default marshaler.
 	if len(args) < 2 {
-		return bson.Marshal(args)
+		return bson.Marshal(map[string]int32(args))
 	}
 
 	sortedArgs := bson.D{}
--- a/model/perf.go
+++ b/model/perf.go
@@ -350,7 +350,8 @@ func (args PerformanceArguments) MarshalBSONValue() (bsontype.Type, []byte, erro
 	// implementation of the bson.ValueMarshaler (see
 	// `https://github.com/mongodb/mongo-go-driver/blob/v1.7.2/bson/bsoncodec/default_value_encoders.go#L767`
 	// for more information), when it is nil we need to first cast it to a
-	// map[string]int32 and then call bson.MarshalValue.
+	// map[string]int32 and then call bson.MarshalValue to get the desired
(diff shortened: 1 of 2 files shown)


model/perf.go
```diff
@@ -345,6 +345,12 @@ var (
 type PerformanceArguments map[string]int32
 
 func (args PerformanceArguments) MarshalBSON() ([]byte, error) {
+	// If the arguments map is empty or only has one item, we can just use
+	// the default marshaler.
+	if len(args) < 2 {
+		return bson.Marshal(map[string]int32(args))
+	}
+
 	sortedArgs := bson.D{}
```
ybrill, line 354: would an alternative be to 
```
var sortedArgs bson.D
```
My thinking is that this way sortedArgs will be nil (unless it's appended to) and get marshaled to null.

ybrill (commented):
is it possible to test this case?

model/perf.go
julianedwards: I believe, in that case, it would have to be `var sortedArgs *bson.D`, `var sortedArgs bson.D` would still marshal to `{}`.

julianedwards: Actually, `bson.D` wraps a slice. So this would work.

julianedwards:
Turns out, the trick is using the `bson.ValueMarshaler` interface. It is a somewhat complicated, but maps are treated a bit special by the go driver and, despite actually being pointers, the `bsoncodec` library doesn't treat them as such. When checking if the `bson.Marshaler` interface's impl is `nil` [here](https://github.com/mongodb/mongo-go-driver/blob/v1.7.2/bson/bsoncodec/default_value_encoders.go#L556-L560), [this](https://github.com/mongodb/mongo-go-driver/blob/v1.7.2/bson/bsoncodec/default_value_encoders.go#L767-L773) func evaluates to `false` since the type is not a `relflect.Ptr` even though it is `nil`. So this always ends up writing an empty object instead of writing null to the `bson.ValueWriter`. With the `bson.ValueMarshaler` interface, we can instead return `bsontype.Null` by casting the `nil` `PerformanceArguments` object to a `map[string]int32` and then calling `bson.ValueMarshal`.

ybrill (approved):
🚀

PR #393, opened 2021-10-05
EVG-15538: Add GUID and temporary info logging to time series update requests

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-15538

To aid in our investigation of missing time series updates to SPS.

### DIFF
--- a/perf/time_series_updater.go
+++ b/perf/time_series_updater.go
@@ -38,6 +38,7 @@ type TimeSeriesModel struct {
 	Measurement string                `json:"measurement"`
 	Arguments   []ArgumentsModel      `json:"args"`
 	Data        []TimeSeriesDataModel `json:"data"`
+	GUID        string                `json:"guid"`
 }
 
 // PerformanceAnalysisService is the interface for the Performance Analysis Service.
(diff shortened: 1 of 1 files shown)


perf/time_series_updater.go
```diff
@@ -59,12 +60,13 @@ func NewPerformanceAnalysisService(baseURL, user string, token string) Performan
...
 func (spc *performanceAnalysisAndTriageClient) ReportUpdatedTimeSeries(ctx context.Context, series TimeSeriesModel) error {
 	startAt := time.Now()
+	series.GUID = utility.RandomString()
 
 	if err := spc.doRequest(http.MethodPost, spc.baseURL+"/time_series/update", ctx, series); err != nil {
 		return errors.WithStack(err)
 	}
 
-	grip.Debug(message.Fields{
+	grip.Info(message.Fields{
```
bsamek, line 69: Log `series.GUID` in this message. That way we can correlate the cedar and SPS logs.

bsamek (changes requested):

perf/time_series_updater.go
julianedwards: It's logged with `"update": series,` line since it is part of the `TimeSeriesUpdate` struct.

ybrill: separately, do we also want to log the error case (atm we're [returning on error](https://github.com/evergreen-ci/cedar/blob/7141126a0841e2a49ba7dde72b1dbfa101679bfd/perf/time_series_updater.go#L64) before we hit this)? For example, if we get `resp.StatusCode != 200` from SPS?

bsamek: Oh, I missed that, thanks. LGTM, but not resolving this thread since it has @ybrill 's question in it.

bsamek (approved):

perf/time_series_updater.go
julianedwards: That will get logged [here](https://github.com/evergreen-ci/cedar/blob/7141126a0841e2a49ba7dde72b1dbfa101679bfd/units/time_series_update.go#L130-L134).

PR #402, opened 2021-10-19
EVG-15411 use maxtime instead of context cancel

ybrill:
[EVG-15411](https://jira.mongodb.org/browse/EVG-15411)

tl;dr We had some discussion [here](https://github.com/mongodb/anser/pull/43#issuecomment-897110393) about how cancelling the context during an operation can cause connection churn. Instead we opted there to use maxtime which will killOp the operation on the server side.

I also removed the `SetReturnDocument(options.After)` option to the findOneAndUpdate. [The docs](https://docs.mongodb.com/manual/reference/method/db.collection.findOneAndUpdate/#mongodb-method-db.collection.findOneAndUpdate) don't explain why this is the default and [the internet](https://stackoverflow.com/questions/55087482/mongodb-does-returning-the-modified-document-after-update-affect-performance) is inconclusive about whether it matters for performance. I figured since we're not actually using the returned document we may as well go with the default (Before). I didn't switch the findOneAndUpdate for an updateOne since updateOne doesn't take a maxTime argument.

### DIFF
--- a/model/historical_test_data.go
+++ b/model/historical_test_data.go
@@ -169,6 +169,7 @@ func (d *HistoricalTestData) Update(ctx context.Context, result TestResult) erro
 		query,
 		pipeline,
 		options.FindOneAndUpdate().SetUpsert(true),
+		options.FindOneAndUpdate().SetReturnDocument(options.After),
 		options.FindOneAndUpdate().SetMaxTime(time.Minute),
 	).Decode(d)
 	grip.DebugWhen(err == nil, message.Fields{


model/historical_test_data.go
```diff
@@ -168,8 +168,8 @@ func (d *HistoricalTestData) Update(ctx context.Context, result TestResult) erro
 		ctx,
 		query,
 		pipeline,
-		options.FindOneAndUpdate().SetReturnDocument(options.After),
```
julianedwards, line 171: I think we need to keep this in order to update the in-memory stat doc, otherwise `d` will still have the outdated information.

ybrill: my thinking was that it doesn't look like we don't actually use the `d` for anything after the update.
I think I see what you're saying, though: if we start using it we'd expect d to be changed after the update. I'll change it back.

ybrill: done

PR #407, opened 2021-11-01
EVG-15628: use Go modules

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15628

I don't think Evergreen will be able to test this PR because it's such a large diff. I tried testing out most of the targets locally and they succeeded.

* Use Go modules and remove vendored dependencies. Instead, the dependencies for our repositories will be dynamically fetched during runtime using `go mod tidy`.
* Upgrade dependencies to the newest minor version. I'm going to assume that repos are obeying semantic versioning, so minor version upgrades are supposed to be backwards compatible. This makes the next deploy slightly riskier, but I'm willing to deploy and monitor it. I think the alternative of trying to make Go modules properly pin all the dependencies to the exact versions from the `glide.lock` is actually riskier, which is why I opted not to do that. I tried multiple times to the module system to use the exact combination of pinned versions of dependencies that we have, but there were multiple problems:
    * Some dependencies have such old dependencies that the Go module system can't even recognize them. In these cases, we have no choice but to upgrade them.
    * Since we `vendor-clean`  a bunch of the transitive vendored dependencies but not all of them, I can't be 100% confident that the version listed in the top-level `glide.lock` is accurate for all dependencies that use that package.
    * Minimal version selection will always try to pick the highest dependency version so when you downgrade a dependency to a specific pinned version, it'll automatically attempt to upgrade all dependencies to the highest minor version possible. You can try downgrading the minimal-version-selection upgrade, but you have to remember to do it after the downgrade occurs and deal with the possibility that downgrade _also_ tries to upgrade other packages.
    * I plan to upgrade these dependencies to the latest minor version anyways. This is basically front-loading the risk that something breaks unintentionally by upgrading to newer packages instead of deferring it to later.
* Fix breaking changes to APIs for upgraded repos.
* Fix usage of testify suite framework. I don't think it's valid to use a testify suite before `suite.Run` has been called, since the assumption is that all setup/teardown is put in `SetupSuite/SetupTest/TearDownSuite/TearDownTest`, so I moved them to use the framework methods more.
* Update and standardize the Evergreen YAML file. The `GOMODCACHE` is set because for CI tests, it's better to not write outside of the agent working directory for Evergreen since it won't clean it up.


julianedwards (approved):
Thanks for fixing the assert errors!

PR #408, opened 2021-11-08
EVG-15729 filter endpoint

ybrill:
Takes a list of (TaskSpecifier, []TestNameRegex) (where a task specifier is an ID, execution, and whether it's a display task) and returns a list of (TaskID, execution, []TestName).

- If the task is a display task we return all the tests that match in any of its execution tasks. 
- If the task regexes is empty we return all the test names in the cache.
- If more than one regex is specified we return test names that match any one of them.


rest/data/interface.go
```diff
@@ -159,6 +162,15 @@ type TestResultsFilterAndSortOptions struct {
 	BaseResults  *TestResultsOptions
 }
 
+// TestSampleOptions specifies the tasks to get the sample for
+// regexes to filter the test names by.
+type TestSampleOptions struct {
+	TaskID          string
```
julianedwards, line 168: The route should accept a list of task ids.

julianedwards (changes requested):

rest/data/interface.go
ybrill: My thinking here was that they'd provide a list of these TestSampleOptions [here](https://github.com/ybrill/cedar/blob/302f05143eb34e5ad2a48995a8434a00aa9e00dd/rest/test_results_routes.go#L177).
Alternatively, I could wrap the list in an object (like I do [on the DB side](https://github.com/ybrill/cedar/blob/302f05143eb34e5ad2a48995a8434a00aa9e00dd/model/test_results.go#L589-L593))

rest/test_results_routes.go
```diff
@@ -167,6 +168,80 @@ func (h *testResultsGetByTaskIDHandler) Run(ctx context.Context) gimlet.Responde
...
+			"method":  "GET",
+			"route":   "/test_results/filtered_samples",
+		}))
+		return gimlet.MakeJSONInternalErrorResponder(err)
+	}
+
+	return gimlet.NewJSONResponse(samples)
+}
+
+// TODO I copy/pasted this out of the evergreen/util package.
```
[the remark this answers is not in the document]
ybrill: I opened https://github.com/evergreen-ci/utility/pull/28

rest/model/test_results.go
```diff
@@ -91,3 +91,25 @@ func (a *APITestResultsStats) Import(i interface{}) error {
 
 	return nil
 }
+
+// APITestResultsSample is a sample of test names for a given task and execution
```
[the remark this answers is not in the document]
ybrill: done

rest/data/interface.go
julianedwards: I guess that is fine. I think they were expecting a many-to-many mapping of task ids to regular expressions. In this case they would have to duplicate the regex list for each task ID.

ybrill: Ah, that would make more sense.
Let me clarify that.

ybrill: I've changed the interface to be as you suggested: a list of tasks and a single list of filters.

rest/data/interface.go
```diff
@@ -159,6 +162,19 @@ type TestResultsFilterAndSortOptions struct {
...
 }
 
+// TestSampleOptions specifies the tasks to get the sample for
+// and regexes to filter the test names by.
+type TestSampleOptions struct {
+	Tasks           []TaskInfo
+	TestNameRegexes []string
+}
+
+type TaskInfo struct {
```
julianedwards, line 172: Documentation.

model/test_results.go
```diff
@@ -579,6 +586,117 @@ func FindTestResults(ctx context.Context, env cedar.Environment, opts FindTestRe
...
+}
+
+func consolidateSamples(testResults []TestResults) []TestResultsSample {
+	type taskExecutionPair struct {
+		taskID    string
+		execution int
+	}
+	taskPairMap := make(map[taskExecutionPair][]string)
+	for _, result := range testResults {
+		taskID := result.Info.DisplayTaskID
```
julianedwards, line 630: Are we sure that we want to always default to using the display task id? What if an execution task id was passed in? In that case, I think we would want to return the execution task id along with the filtered failed test names.

model/test_results.go
```diff
@@ -579,6 +586,117 @@ func FindTestResults(ctx context.Context, env cedar.Environment, opts FindTestRe
...
+
+	return samples
+}
+
+func filterTestNames(samples []TestResultsSample, regexStrings []string) []TestResultsSample {
+	regexes := []*regexp.Regexp{}
+	for _, regexString := range regexStrings {
+		testNameRegex, err := regexp.Compile(regexString)
+		if err != nil {
+			continue
```
julianedwards, line 659: Should we not error here? It seems like a valid case in which we should return an error to the user.

rest/test_results_routes.go
```diff
@@ -167,6 +167,55 @@ func (h *testResultsGetByTaskIDHandler) Run(ctx context.Context) gimlet.Responde
...
+		sc: h.sc,
+	}
+}
+
+func (h *testResultsGetFilteredSamplesHandler) Parse(_ context.Context, r *http.Request) error {
+	body := utility.NewRequestReader(r)
+	defer body.Close()
+
+	if err := utility.ReadJSON(body, h.options); err != nil {
+		return errors.Wrap(err, "Argument read error")
```
julianedwards, line 197: Nit: `argument`

model/test_results.go
```diff
@@ -579,6 +586,117 @@ func FindTestResults(ctx context.Context, env cedar.Environment, opts FindTestRe
...
+
+	return samples
+}
+
+func consolidateSamples(testResults []TestResults) []TestResultsSample {
+	type taskExecutionPair struct {
+		taskID    string
+		execution int
+	}
+	taskPairMap := make(map[taskExecutionPair][]string)
```
julianedwards, line 628: Nit: we can clean this up a bit by creating the samples slice in the same for loop.
```
taskPairMap := make(map[taskExecutionPair]int) // where int is the index of the task/exec pair in the samples slice
samples := []TestResultsSample{}
for _, result := range testResults {
	pair := taskExecutionPair{
		taskID:    taskID,
		execution: result.Info.Execution,
	}
        idx, ok := taskPairMap[pair]
        if ok {
                samples[idx].FailedTestNames = append(samples[idx].FailedTestNames, results.FailedTestsSamples...)
        } else {
                samples = append(samples, TestResultsSample{
			TaskID:          pair.taskID,
			Execution:       pair.execution,
			FailedTestNames: names,
		})
                taskPairMap[pair] = len(samples) - 1
        }
}
```

model/test_results.go
```diff
@@ -446,6 +446,13 @@ type TestResultsStats struct {
 	FailedCount int `bson:"failed_count"`
 }
 
+// TestResultsSample contains test names culled from a test result's FailedTestsSample.
+type TestResultsSample struct {
```
julianedwards, line 450: I remember the UI team also wanting the stats for these so that they can display `X/Y failing tests match` for each task id on the task history page. Do they still want that?

julianedwards (changes requested):

model/test_results.go
khelif96: Yes we do I forgot I mentioned this to Julian and not Jonathan. Sorry

ybrill: I updated it to also return the total length of the cache (and the aggregate length of all the execution task caches in the case of a display task).

rest/test_results_routes.go
ybrill: done

model/test_results.go
ybrill: fixed. this will return an error to the caller now.

rest/data/interface.go
ybrill: done

model/test_results.go
ybrill: I rejiggered `consolidateSamples` to return what was asked for. So now if you ask for an execution task you'll get just that execution task.

model/test_results.go
ybrill: this is neat.
I refactored the `consolidateSamples` function (for the execution task comment above) and I'm not sure this comment is still applicable.

rest/model/test_results.go
```diff
@@ -91,3 +91,27 @@ func (a *APITestResultsStats) Import(i interface{}) error {
 
 	return nil
 }
+
+// APITestResultsSample is a sample of test names for a given task and execution.
+type APITestResultsSample struct {
```
julianedwards, line 96: Should this have JSON tags?

ybrill: done

julianedwards (approved):

PR #410, opened 2021-11-11
EVG-14837: return 0 instead of 404 when task is not found

hhoke:
This was created to ameliorate the problem designed in https://mongodb.slack.com/archives/C0V896UV8/p1636395702304700 . Only the last commit is new. When I run `go mod-tidy` it adds the following to my go.mod: `gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c`. I assumed I should keep it there.


rest/perf_routes.go
```diff
@@ -203,7 +203,8 @@ func (h *perfCountByTaskIdHandler) Parse(_ context.Context, r *http.Request) err
 // number of PerformanceResults found
 func (h *perfCountByTaskIdHandler) Run(ctx context.Context) gimlet.Responder {
 	perfResults, err := h.sc.FindPerformanceResults(ctx, h.opts)
-	if err != nil {
+	errResp, ok := err.(gimlet.ErrorResponse)
+	if err != nil && (!ok || errResp.StatusCode != http.StatusNotFound) {
```
julianedwards, line 207: I would push this logic down to the `data` package.

julianedwards (changes requested):

rest/perf_routes.go
Kimchelly: I think pushing the logic down would require modifying the other routes that also depend on the 404 to handle the "no results" case. Is it okay if the other routes (e.g. `perfGetByTaskNameHandler`) return an empty response when there's no results?

In any case, I think it's probably better to not 404 in the case of no results since a 404 usually means that the REST route is not a valid route, not that it returned empty results. But I don't know if any users are depending on 404 meaning "no results" instead of "no such REST route".

hhoke: @julianedwards what do you think about Kim's arguments.

julianedwards: Ok that makes sense, I am fine with this then.

julianedwards (approved):

Kimchelly (commented):
LGTM % mod tidy.

Kimchelly (approved):

PR #412, opened 2021-11-16
EVG-15612: set mongo driver max pool size to 1000

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-15612

The atlas cluster is currently an m200 tier, which has a max connection limit of 128k. Since certain loads can be write-heavy, we should increase the max connection pool size. The app servers each have a ulimit of 256k so they can also safely handle this. I picked 1000 kind of arbitrarily (the default is 100), but I think its enough, we can always increase if needed.

### DIFF
--- a/environment.go
+++ b/environment.go
@@ -57,6 +57,7 @@ func NewEnvironment(ctx context.Context, name string, conf *Configuration) (Envi
 			SetConnectTimeout(conf.MongoDBDialTimeout).
 			SetSocketTimeout(conf.SocketTimeout).
 			SetServerSelectionTimeout(conf.SocketTimeout).
+			SetMaxPoolSize(1000).
 			SetMonitor(apm.NewLoggingMonitor(ctx, time.Minute, apm.NewBasicMonitor(&apm.MonitorConfig{AllTags: true})).DriverAPM())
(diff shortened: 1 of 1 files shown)


Kimchelly (approved):
LGTM mod questions for understanding.

Does it already hit the 100 default limit today when writes are heavy? Do we have a way of knowing if/when is the next time hits the max connection pool limit?

julianedwards:
> LGTM mod questions for understanding.
> 
> Does it already hit the 100 default limit today when writes are heavy? Do we have a way of knowing if/when is the next time hits the max connection pool limit?

In load testing I saw that it the default limit was hit, but I am not sure because when we have write-heavy load there are often obscure context canceled messages (sometimes deadline exceeded, sometimes just canceled messages).  We can always look out for the explicit "timeout while waiting for client pool connection" message. What is interesting is that, lately, when we have write heavy-load—say, a large set of test results comes in—the database actually seems healthy but connections still get canceled. I am not 100% sure that this will help solve our issues, it may just be writes slow down everything, in which case sharding should do the trick. But I am hoping that this will help utilize the existing resources more since we are severely underusing at this point.

Kimchelly (approved):

PR #415, opened 2021-11-17
EVG-14303 adjust log level conditionally

ybrill:
https://jira.mongodb.org/browse/EVG-14303

Wrap all the rest/route level `grip.Error`s in a function that lowers the log level for 404s.
I figured we still need to return the 404s (like Kim's https://github.com/evergreen-ci/cedar/pull/410#discussion_r747814999) but it's not really an _error_ so the `Info` level makes more sense.

### DIFF
--- a/rest/buildlogger_routes.go
+++ b/rest/buildlogger_routes.go
@@ -76,12 +76,12 @@ func (h *logGetByIDHandler) Run(ctx context.Context) gimlet.Responder {
 	data, next, paginated, err := h.sc.FindLogByID(ctx, h.opts)
 	if err != nil {
 		err = errors.Wrapf(err, "problem getting log by id '%s'", h.opts.ID)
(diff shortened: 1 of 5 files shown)


ybrill:
I wonder, though, if this whole PR is worthwhile given [EVG-15517](https://jira.mongodb.org/browse/EVG-15517).
In other words, if we're going to be changing the routes to return 200 for an empty result set then this would be unnecessary. Though I don't know if we're planning to do that for _every_ route.

julianedwards (approved):

julianedwards:
> I wonder, though, if this whole PR is worthwhile given [EVG-15517](https://jira.mongodb.org/browse/EVG-15517). In other words, if we're going to be changing the routes to return 200 for an empty result set then this would be unnecessary. Though I don't know if we're planning to do that for _every_ route.

I think that it is worth it, cause we may keep it for certain routes.

PR #428, opened 2022-01-12
EVG-15699: Pin protobuf compiler dependencies to a specific version

julianedwards:
EVG-15699: https://jira.mongodb.org/browse/EVG-15699

Once this is approved, I will put up the same changes for jasper, juniper, and poplar.


makefile
```diff
@@ -96,9 +96,19 @@ htmlCoverageOutput := $(foreach target,$(testPackages),$(buildDir)/output.$(targ
...
-proto:
-	@mkdir -p rpc/internal
-	protoc --go_out=plugins=grpc:rpc/internal *.proto
+protoOS := $(shell uname -s | tr A-Z a-z)
+ifeq ($(protoOS),darwin)
+protoOS := osx
+endif
+protoOS := $(protoOS)-$(shell uname -m | tr A-Z a-z)
+$(buildDir)/protoc:
+	curl --retry 10 --retry-max-time 60 -L0 https://github.com/protocolbuffers/protobuf/releases/download/v3.6.1/protoc-3.6.1-$(protoOS).zip --output protoc-3.6.1-$(protoOS).zip
```
Kimchelly, line 105: Can you put the protoc and protoc-gen-go versions in separate variables so we can easily change/upgrade them?

makefile
```diff
@@ -96,9 +96,21 @@ htmlCoverageOutput := $(foreach target,$(testPackages),$(buildDir)/output.$(targ
...
+endif
+protoOS := $(protoOS)-$(shell uname -m | tr A-Z a-z)
+$(buildDir)/protoc:
+	curl --retry 10 --retry-max-time 60 -L0 https://github.com/protocolbuffers/protobuf/releases/download/v$(protocVersion)/protoc-$(protocVersion)-$(protoOS).zip --output protoc.zip
+	unzip -q protoc.zip -d $(buildDir)/protoc
+	rm -f protoc.zip
+	GOBIN=$(abspath $(buildDir)) $(gobin) install github.com/golang/protobuf/protoc-gen-go@v$(protocGenGoVersion)
+proto: $(buildDir)/protoc
+	mkdir -p rpc/internal
+	PATH=$(PATH):$(abspath $(buildDir)) $(buildDir)/protoc/bin/protoc --go_out=plugins=grpc:rpc/internal *.proto
```
Kimchelly, line 113: The PATH has to be `PATH="$(abspath $(buildDir)):$(PATH)"` or else any other `protoc-gen-go` or `protoc` executable already present in the `PATH` will be used. It still needs double quotes during assignment so that the entire `PATH` is considered as a single string by shell parsing rules.

makefile
```diff
@@ -96,9 +96,21 @@ htmlCoverageOutput := $(foreach target,$(testPackages),$(buildDir)/output.$(targ
...
+protoOS := $(shell uname -s | tr A-Z a-z)
+ifeq ($(protoOS),darwin)
+protoOS := osx
+endif
+protoOS := $(protoOS)-$(shell uname -m | tr A-Z a-z)
+$(buildDir)/protoc:
+	curl --retry 10 --retry-max-time 60 -L0 https://github.com/protocolbuffers/protobuf/releases/download/v$(protocVersion)/protoc-$(protocVersion)-$(protoOS).zip --output protoc.zip
+	unzip -q protoc.zip -d $(buildDir)/protoc
+	rm -f protoc.zip
+	GOBIN=$(abspath $(buildDir)) $(gobin) install github.com/golang/protobuf/protoc-gen-go@v$(protocGenGoVersion)
```
Kimchelly, line 110: `GOBIN="$(abspath $(buildDir))"`

Kimchelly (changes requested):
It would be preferable to make `protoc-gen-go`'s binary version (you set it to v1.3.2) agree with the version of the `github.com/golang/protobuf` that we're using for Go modules (v1.5.2), since having an older binary version may not be forward compatible with the newer Go module version, and it would be difficult to remember to update both. Can we make `protocGenGoVersion` (v1.3.2) agree with the Go module `github.com/google/protobuf` (v1.5.2) and add a comment saying they should be the same?

julianedwards:
The problem is, using the binary from v1.5.2 already is incompatible with our pb files :(. I have [EVG-15720](https://jira.mongodb.org/browse/EVG-15720) to update to a new version, but I'd rather wait until then to do it cause it's going to be a headache.

julianedwards:
The problem is, using the binary from v1.5.2 already is incompatible with our pb files :(. I have [EVG-15720](https://jira.mongodb.org/browse/EVG-15720) to update to a new version, but I'd rather wait until then to do it cause it's going to be a headache.

Kimchelly (approved):

PR #429, opened 2022-01-18
EVG-15720: Upgrade to newer version of protobuf

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-15720

This is finally happening! 

I tested that this is compatible with .pb client files generated by the older version of protobuf we are using.


buildlogger.proto
```diff
@@ -2,7 +2,7 @@ syntax = "proto3";
 
 package cedar;
 
-option go_package = "internal";
+option go_package = "rpc/internal";
```
Kimchelly, line 5: This doesn't require a `./` to make it a relative path, but `./gopb` is necessary in juniper?

Kimchelly (changes requested):

buildlogger.proto
julianedwards: Yeah, it doesn't make sense. This is the error message: 
```
The import path must contain at least one period ('.') or forward slash ('/') character.

See https://developers.google.com/protocol-buffers/docs/reference/go-generated#package for more information.
```

go.mod
```diff
@@ -28,6 +28,7 @@ require (
 	golang.org/x/sys v0.0.0-20211123173158-ef496fb156ab // indirect
 	google.golang.org/genproto v0.0.0-20211118181313-81c1377c94b1 // indirect
 	google.golang.org/grpc v1.43.0
+	google.golang.org/protobuf v1.27.1
```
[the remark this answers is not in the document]
julianedwards: Done.

Kimchelly (approved):

PR #430, opened 2022-01-21
EVG-16037: Only send time series IDs to signal process service

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-16072

The SPS is now using the analytics node to run aggregations that fetch performance result time series data so we only need to send the time series ID rather than running the aggregation in amboy jobs and sending all of the time series data.

### DIFF
--- a/model/perf.go
+++ b/model/perf.go
@@ -64,7 +64,7 @@ var (
 )
 
 // CreateUnanalyzedSeries converts the PerformanceResult into an
-// UnanalyzedPerformanceSeries for comminication with the signal processing
+// UnanalyzedPerformanceSeries for communication with the signal processing
 // service.
 func (result PerformanceResult) CreateUnanalyzedSeries() UnanalyzedPerformanceSeries {
 	measurements := make([]string, len(result.Rollups.Stats))
--- a/model/perf.go
+++ b/model/perf.go
@@ -7,6 +7,7 @@ import (
 	"hash"
 	"io"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/evergreen-ci/cedar"
@@ -392,6 +393,18 @@ func (args PerformanceArguments) MarshalBSONValue() (bsontype.Type, []byte, erro
 	})
 
 	return bson.MarshalValue(&sortedArgs)
+}
+
+func (args PerformanceArguments) String() string {
+	var sortedArgs []string
+	for key, value := range args {
+		sortedArgs = append(sortedArgs, fmt.Sprintf("%s:%d", key, value))
+	}
+	sort.Slice(sortedArgs, func(i, j int) bool {
+		return sortedArgs[i] < sortedArgs[j]
+	})
+
+	return fmt.Sprintf("[%s]", strings.Join(sortedArgs, " "))
 }
 
 // ID creates a unique hash for a performance result.
--- a/model/perf.go
+++ b/model/perf.go
@@ -73,11 +73,12 @@ func (result PerformanceResult) CreateUnanalyzedSeries() UnanalyzedPerformanceSe
 	}
 
 	return UnanalyzedPerformanceSeries{
-		Project:   result.Info.Project,
-		Variant:   result.Info.Variant,
-		Task:      result.Info.TaskName,
-		Test:      result.Info.TestName,
-		Arguments: result.Info.Arguments,
+		Project:      result.Info.Project,
+		Variant:      result.Info.Variant,
+		Task:         result.Info.TaskName,
+		Test:         result.Info.TestName,
+		Arguments:    result.Info.Arguments,
+		Measurements: measurements,
 	}
 }
 
--- a/model/perf.go
+++ b/model/perf.go
@@ -63,12 +63,21 @@ var (
 	perfVersionKey           = bsonutil.MustHaveTag(PerformanceResult{}, "Version")
 )
 
-func (info *PerformanceResultInfo) ToPerformanceResultSeriesID() PerformanceResultSeriesID {
-	return PerformanceResultSeriesID{
-		Project: info.Project,
-		Variant: info.Variant,
-		Task:    info.TaskName,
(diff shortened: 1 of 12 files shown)


units/time_series_update_test.go
```diff
@@ -165,111 +20,53 @@ func (m *MockPerformanceAnalysisService) ReportUpdatedTimeSeries(ctx context.Con
...
 	env := cedar.GetEnvironment()
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	defer func() {
-		require.NoError(t, tearDown(env))
-	}()
-
 	t.Run("ReportsTimeSeries", func(t *testing.T) {
 		_ = env.GetDB().Drop(ctx)
```
Kimchelly, line 29: Should this (and other places where the DB gets dropped) `require.NoError`?

model/time_series_update.go
```diff
@@ -128,25 +167,32 @@ func GetPerformanceResultSeriesIdsNeedingTimeSeriesUpdate(ctx context.Context, e
...
 	}
 	return res, nil
 }
 
-// GetPerformanceResultSeriesIDs finds all performance result series IDs in the
-// the database.
-func GetPerformanceResultSeriesIDs(ctx context.Context, env cedar.Environment) ([]PerformanceResultSeriesID, error) {
+// GetAllPerformanceResultSeriesIDs finds all performance result series in the
+// database.
+func GetAllPerformanceResultSeriesIDs(ctx context.Context, env cedar.Environment) ([]UnanalyzedPerformanceSeries, error) {
```
Kimchelly, line 195: * Should this be tested?
* This query is pretty similar to `GetUnanalyzedPerformanceSeries`, which has limits to avoid excessive query runtime. Wouldn't this be more expensive since it doesn't limit the results?

rest/data/perf.go
```diff
@@ -165,27 +164,28 @@ func (dbc *DBConnector) FindPerformanceResultWithChildren(ctx context.Context, i
...
+		return gimlet.ErrorResponse{StatusCode: http.StatusInternalServerError, Message: errors.Wrap(err, "failed to get time series ids").Error()}
 	}
 
 	catcher := grip.NewBasicCatcher()
 
-	for _, id := range ids {
-		job := units.NewUpdateTimeSeriesJob(id)
+	for _, series := range allSeries {
+		job := units.NewUpdateTimeSeriesJob(series)
 		err := amboy.EnqueueUniqueJob(ctx, queue, job)
```
Kimchelly, line 179: L180-181 can be combined.

Kimchelly (changes requested):

model/perf.go
```diff
@@ -62,12 +63,21 @@ var (
...
+// CreateUnanalyzedSeries converts the PerformanceResult into an
+// UnanalyzedPerformanceSeries for communication with the signal processing
+// service.
+func (result PerformanceResult) CreateUnanalyzedSeries() UnanalyzedPerformanceSeries {
+	measurements := make([]string, len(result.Rollups.Stats))
+	for i, stat := range result.Rollups.Stats {
+		measurements[i] = stat.Name
+	}
+
+	return UnanalyzedPerformanceSeries{
```
ybrill, line 75: do we need to set the `Measurements` field?

model/perf.go
```diff
@@ -394,6 +403,18 @@ func (args PerformanceArguments) MarshalBSONValue() (bsontype.Type, []byte, erro
 	return bson.MarshalValue(&sortedArgs)
 }
 
+func (args PerformanceArguments) String() string {
+	var sortedArgs []string
+	for key, value := range args {
+		sortedArgs = append(sortedArgs, fmt.Sprintf("%s:%d", key, value))
+	}
+	sort.Slice(sortedArgs, func(i, j int) bool {
```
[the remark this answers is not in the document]
julianedwards: It is to make the string consistent to avoid duplicate amboy jobs since this is part of the ID string.

model/time_series_update.go
julianedwards: Yeah this is aggregation would only be used in an emergency situation where we would want to rerun all of the SPS notifications, it has never been used AFAIK. It has to be manually called via a protected REST route. It is tested in the `rest` package.

units/time_series_update_periodic_test.go
```diff
@@ -35,6 +39,136 @@ func tearDownPeriodicTest(env cedar.Environment) error {
...
+				break
+			}
+		}
+	}
+	return newVals
+}
+
+func makePerfResultsWithChangePoints(unique string, seed int64) ([]testResultsAndRollups, [][]int) {
+	// Deterministic testing on failure.
+	rand.Seed(seed)
```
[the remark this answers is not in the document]
julianedwards: I guess not, I didn't write these tests so I am not sure the logic behind this move. I also do not think it's worth to do too much refactoring since we are going to move away from this code in the nearish future.

Kimchelly (approved):

units/time_series_update.go
```diff
@@ -81,75 +91,45 @@ func (j *timeSeriesUpdateJob) Run(ctx context.Context) {
...
-			Measurement: perfData.PerformanceResultId.Measurement,
-		}
-		series.Arguments = []perf.ArgumentsModel{}
-		for k, v := range perfData.PerformanceResultId.Arguments {
-			series.Arguments = append(series.Arguments, perf.ArgumentsModel{
-				Name:  k,
-				Value: v,
-			})
-		}
-		filteredData := filterOldExecutions(perfData.TimeSeries)
```
ybrill, line 120: is the idea that we no longer need this because the stat names are the same for every execution of the task and they're only going to fetch the values from the latest executions of each task?

julianedwards: We can remove this because it is now done on the SPS side—they take care of filtering out older executions.

ybrill (approved):

PR #439, opened 2022-02-24
EVG-16137, EVG-16136: Store test results in Parquet and fetch and compare

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-16137, https://jira.mongodb.org/browse/EVG-16136

Store test results in both BSON and Parquet, when fetching results pull both Parquet and BSON and compare.

### DIFF
--- a/model/test_results.go
+++ b/model/test_results.go
@@ -219,6 +219,12 @@ func (t *TestResults) Append(ctx context.Context, results []TestResult) error {
 	if err != nil {
 		return errors.Wrap(err, "creating Presto bucket writer")
 	}
+	var closedWriter bool
+	defer func() {
+		if !closedWriter {
+			_ = w.Close()
+		}
+	}()
 	pw, err := writer.NewParquetWriterFromWriter(w, new(ParquetTestResults), 1)
 	if err != nil {
 		return errors.Wrap(err, "creating new Parquet writer")
@@ -229,6 +235,7 @@ func (t *TestResults) Append(ctx context.Context, results []TestResult) error {
 	if err = pw.WriteStop(); err != nil {
 		return errors.Wrap(err, "stopping Parquet writer")
 	}
+	closedWriter = true
 	if err := w.Close(); err != nil {
 		return errors.Wrap(err, "closing Presto bucket writer")
 	}
@@ -323,7 +330,6 @@ func (t *TestResults) Download(ctx context.Context) ([]TestResult, error) {
 
 	switch t.Artifact.Version {
 	case 0:
-		catcher := grip.NewBasicCatcher()
 		bucket, err := t.GetBucket(ctx)
 		if err != nil {
 			return nil, err
@@ -334,6 +340,8 @@ func (t *TestResults) Download(ctx context.Context) ([]TestResult, error) {
 		for iter.Next(ctx) {
 			results = append(results, iter.Item())
 		}
+
+		catcher := grip.NewBasicCatcher()
 		catcher.Wrap(iter.Err(), "iterating test results")
 		catcher.Wrap(iter.Close(), "closing test results iterator")
 
@@ -380,8 +388,8 @@ func (t *TestResults) downloadParquet(ctx context.Context) ([]TestResult, error)
 	data, err := ioutil.ReadAll(r)
 	catcher.Wrap(err, "reading Parquet test results")
 	catcher.Wrap(r.Close(), "closing Presto bucket reader")
-	if err = catcher.Resolve(); err != nil {
-		return nil, err
+	if catcher.HasErrors() {
+		return nil, catcher.Resolve()
 	}
 
 	pr, err := reader.NewParquetReader(buffer.NewBufferFileFromBytes(data), new(ParquetTestResults), 1)
--- a/model/test_results.go
+++ b/model/test_results.go
@@ -376,9 +376,12 @@ func (t *TestResults) downloadParquet(ctx context.Context) ([]TestResult, error)
 		return nil, errors.Wrap(err, "getting Parquet test results")
 	}
 
+	catcher := grip.NewBasicCatcher()
 	data, err := ioutil.ReadAll(r)
-	if err != nil {
-		return nil, errors.Wrap(err, "reading Parquet test results")
+	catcher.Wrap(err, "reading Parquet test results")
+	catcher.Wrap(r.Close(), "closing Presto bucket reader")
+	if err = catcher.Resolve(); err != nil {
+		return nil, err
 	}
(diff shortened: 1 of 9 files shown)


model/test_results.go
```diff
@@ -186,40 +199,38 @@ func (t *TestResults) Append(ctx context.Context, results []TestResult) error {
...
+	if err != nil {
+		return errors.Wrap(err, "creating new Parquet writer")
+	}
+	if err = pw.Write(t.convertToParquet(allResults)); err != nil {
+		return errors.Wrap(err, "writing Parquet test results")
+	}
+	if err = pw.WriteStop(); err != nil {
+		return errors.Wrap(err, "stopping Parquet writer")
+	}
+	if err := w.Close(); err != nil {
```
Kimchelly, line 232: Should this be closed in a defer on L222?

model/test_results.go
```diff
@@ -295,42 +321,115 @@ func (t *TestResults) Download(ctx context.Context) ([]TestResult, error) {
...
-	var results testResultsDoc
-	catcher := grip.NewBasicCatcher()
-	bucket, err := t.GetBucket(ctx)
-	if err != nil {
-		return nil, err
-	}
-
 	switch t.Artifact.Version {
 	case 0:
+		catcher := grip.NewBasicCatcher()
```
Kimchelly, line 326: Move this initialization down to L336 since it's not used until then.

model/test_results.go
```diff
@@ -41,6 +45,13 @@ const (
 	parquetDateFormat     = "2006-01-02"
 )
 
+// invalidParquetAlertCount is a temporary count to avoid over-logging when we
+// detect invalid Parquet test results.
```
Kimchelly, line 49: What is the parquet alert intended to do? Is it supposed to tell us to fix the mismatched BSON/Parquet test results or is it just to make us aware that it happened at least a few times?

model/test_results.go
```diff
@@ -295,42 +321,115 @@ func (t *TestResults) Download(ctx context.Context) ([]TestResult, error) {
...
 
-		catcher.Wrap(bson.Unmarshal(data, &results), "unmarshalling test results")
+		return bsonResults, nil
 	default:
-		catcher.Errorf("unsupported test results artifact version '%d'", t.Artifact.Version)
+		return nil, errors.Errorf("unsupported test results artifact version '%d'", t.Artifact.Version)
+	}
+}
+
+// TODO (EVG-16140): Move this to Download (above) once we do the BSON to
```
Kimchelly, line 366: Does this mean that version 1 test results will only download from Presto eventually?

model/test_results.go
```diff
@@ -295,42 +321,115 @@ func (t *TestResults) Download(ctx context.Context) ([]TestResult, error) {
...
+			return nil, err
 		}
-
-		data, err := ioutil.ReadAll(r)
-		catcher.Add(r.Close())
-		if err != nil {
-			catcher.Wrap(err, "reading test results")
-			break
+		parquetResults, err := t.downloadParquet(ctx)
+		if err != nil && !pail.IsKeyNotFoundError(err) {
```
Kimchelly, line 347: I'm wondering about the backward compatibility of this change. Won't this error when downloading any previous test results since they're not stored in Presto?

model/test_results.go
```diff
@@ -295,42 +321,115 @@ func (t *TestResults) Download(ctx context.Context) ([]TestResult, error) {
...
+	r, err := prestoBucket.Get(ctx, t.PrestoPartitionKey())
+	if err != nil {
+		return nil, errors.Wrap(err, "getting Parquet test results")
+	}
+
+	catcher := grip.NewBasicCatcher()
+	data, err := ioutil.ReadAll(r)
+	catcher.Wrap(err, "reading Parquet test results")
+	catcher.Wrap(r.Close(), "closing Presto bucket reader")
+	if err = catcher.Resolve(); err != nil {
```
Kimchelly, line 383: This could be rewritten as:
```
if catcher.HasErrors() {
    return nil, catcher.Resolve()
}
```

Kimchelly (changes requested):

model/test_results.go
julianedwards: Just to make us aware that it is occurring so we can revert, but it is most likely that if it happens to one file its happening to all since they are serialized in the same way. We don't want to bombard splunk with these messages.

model/test_results.go
julianedwards: Nope because we ignore the `IsKeyNotFoundError`, so 404s are ignored. Anything else (like auth or network) we would want to be error on.

model/test_results.go
julianedwards: Yeah, we have a backfill ticket to migrate all data within our TTL range to Presto so everything will be either in Presto or deleted.

model/test_results.go
julianedwards: I am going to add a defer func with a conditional because I think it is worth keeping this here since `w.Close()` will flush the data to S3 and if that errors we should definitely know about it.

model/test_results.go
Kimchelly: Ah yeah, missed the negation on `IsKeyNotFound`.

Kimchelly (approved):

model/test_results.go
```diff
@@ -295,42 +328,116 @@ func (t *TestResults) Download(ctx context.Context) ([]TestResult, error) {
...
-		data, err := ioutil.ReadAll(r)
-		catcher.Add(r.Close())
-		if err != nil {
-			catcher.Wrap(err, "reading test results")
-			break
+		parquetResults, err := t.downloadParquet(ctx)
+		if err != nil && !pail.IsKeyNotFoundError(err) {
+			return nil, err
+		} else if err == nil && invalidParquetAlertCount < 5 {
+			if reflect.DeepEqual(parquetResults, bsonResults) {
```
rtimmons, line 358: shouldnt' this be `if ! equal`?

rtimmons: Is there any way to see what the (deep) diff actually is?

julianedwards: Good catch!

julianedwards: I am nervous to log the diff because these test results can be really large for certain projects (sometimes on the order of 100k results).

julianedwards: But since we have the taskid/execution, we can retroactively download the test BSON and Parquet results and manually evaluate them.

rtimmons (approved):
Let's find a time to deploy this together and watch it live while querying presto.

PR #442, opened 2022-03-02
EVG-16138: Production-Ready BSON to Parquet Backfill 

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-16138

- Create a new Amboy job that migrates BSON test result data from our S3 bucket to Parquet format and writes it to the Presto bucket.
- Add a new `BatchJobController` model to enable controlling batch jobs without requiring app restarts/deploys.
- Add a new `MigrationStats` model, used as a sub-document in the `TestResults` docs in the temporary backfill collection for keeping track of the batch job stats and enabling versioning (in case we need to rerun the migration).
- Clean up some test files in the `units` package.

Testing
- I added unit tests, and we will test in this staging. The job itself will not be run yet in production.

### DIFF
--- a/model/test_results.go
+++ b/model/test_results.go
@@ -81,6 +81,11 @@ type TestResults struct {
 	// This is an optimization for Evergreen's UI features that display a
 	// limited number of failing tests for a task.
 	FailedTestsSample []string `bson:"failed_tests_sample"`
+	// Migration holds statistics for the BSON to Parquet backfill, it is
+	// only present in the temporary test results migration collection.
+	//
+	// TODO (EVG-16140): Remove once we do the  BSON to Parquet cutover.
+	Migration *MigrationStats `bson:"migration,omitempty"`
 
 	env                cedar.Environment
 	bucket             string
@@ -97,6 +102,8 @@ var (
 	testResultsArtifactKey          = bsonutil.MustHaveTag(TestResults{}, "Artifact")
 	testResultsStatsKey             = bsonutil.MustHaveTag(TestResults{}, "Stats")
 	testResultsFailedTestsSampleKey = bsonutil.MustHaveTag(TestResults{}, "FailedTestsSample")
+	// TODO (EVG-16140): Remove once we do the BSON to Parquet cutover.
+	TestResultsMigrationKey = bsonutil.MustHaveTag(TestResults{}, "Migration")
 )
 
 // CreateTestResults is an entry point for creating a new TestResults.
@@ -228,33 +235,8 @@ func (t *TestResults) Append(ctx context.Context, results []TestResult) error {
 		return errors.Wrap(err, "appending BSON test results")
 	}
 
-	bucket, err := t.GetPrestoBucket(ctx)
-	if err != nil {
-		return err
-	}
-	w, err := bucket.Writer(ctx, t.PrestoPartitionKey())
-	if err != nil {
-		return errors.Wrap(err, "creating Presto bucket writer")
-	}
-	var closedWriter bool
-	defer func() {
-		if !closedWriter {
-			_ = w.Close()
-		}
-	}()
-	pw, err := writer.NewParquetWriterFromWriter(w, new(ParquetTestResults), 1)
-	if err != nil {
-		return errors.Wrap(err, "creating new Parquet writer")
-	}
-	if err = pw.Write(t.convertToParquet(allResults)); err != nil {
-		return errors.Wrap(err, "writing Parquet test results")
-	}
-	if err = pw.WriteStop(); err != nil {
-		return errors.Wrap(err, "stopping Parquet writer")
-	}
-	closedWriter = true
-	if err := w.Close(); err != nil {
-		return errors.Wrap(err, "closing Presto bucket writer")
+	if err = t.uploadParquet(ctx, t.convertToParquet(allResults)); err != nil {
+		return errors.Wrap(err, "appending Parquet test results")
 	}
 
 	if err = t.env.GetStatsCache(cedar.StatsCacheTestResults).AddStat(cedar.Stat{
@@ -285,6 +267,36 @@ func (t *TestResults) uploadBSON(ctx context.Context, results []TestResult) erro
 	}
 
 	return errors.Wrap(bucket.Put(ctx, testResultsCollection, bytes.NewReader(data)), "uploading BSON test results")
+}
+
+func (t *TestResults) uploadParquet(ctx context.Context, results *ParquetTestResults) error {
+	bucket, err := t.GetPrestoBucket(ctx)
+	if err != nil {
+		return err
+	}
+	w, err := bucket.Writer(ctx, t.PrestoPartitionKey())
+	if err != nil {
+		return errors.Wrap(err, "creating Presto bucket writer")
+	}
+	var closedWriter bool
+	defer func() {
+		if !closedWriter {
+			_ = w.Close()
+		}
+	}()
+	pw, err := writer.NewParquetWriterFromWriter(w, new(ParquetTestResults), 1)
+	if err != nil {
+		return errors.Wrap(err, "creating new Parquet writer")
+	}
+	if err = pw.Write(results); err != nil {
+		return errors.Wrap(err, "writing Parquet test results")
+	}
+	if err = pw.WriteStop(); err != nil {
+		return errors.Wrap(err, "stopping Parquet writer")
+	}
+
+	closedWriter = true
+	return errors.Wrap(w.Close(), "closing Presto bucket writer")
 }
 
 func (t *TestResults) updateStatsAndFailedSample(ctx context.Context, results []TestResult) error {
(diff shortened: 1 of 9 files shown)


model/test_results.go
```diff
@@ -81,6 +81,11 @@ type TestResults struct {
 	// This is an optimization for Evergreen's UI features that display a
 	// limited number of failing tests for a task.
 	FailedTestsSample []string `bson:"failed_tests_sample"`
+	// Migration holds statistics for the BSON to Parquet backfill, it is
+	// only present in the temporary test results migration collection.
+	//
+	// TODO (EVG-16140): Remove once we do the  BSON to Parquet cutover.
+	Migration *MigrationStats `bson:"migration,omitempty"`
```
rtimmons, line 88: Note to self: do we need this or would a separate collection be easier?

units/bson_to_parquet_backfill.go
```diff
@@ -0,0 +1,183 @@
...
+	// the batch size number of UpdateOne operations.
+	updates := make([]mongo.WriteModel, j.controller.BatchSize)
+	for i := 0; i < j.controller.BatchSize; i++ {
+		updates[i] = mongo.NewUpdateOneModel().
+			SetFilter(bson.M{"$or": []bson.M{
+				{model.TestResultsMigrationKey: bson.M{"$exists": false}},
+				{bsonutil.GetDottedKeyName(model.TestResultsMigrationKey, model.MigrationStatsVersionKey): bson.M{"$lt": j.controller.Version}},
+				{"$and": []bson.M{
+					{bsonutil.GetDottedKeyName(model.TestResultsMigrationKey, model.MigrationStatsStartedAtKey): bson.M{"$and": []bson.M{
+						{"gt": time.Time{}},
```
rtimmons, line 110: this gets rows with started at > now?
and now is evaluated for each item?

units/bson_to_parquet_backfill.go
```diff
@@ -0,0 +1,183 @@
...
+	for i, result := range results {
+		if err = result.DownloadConvertAndWriteParquet(ctx); err != nil {
+			return errors.Wrap(err, "downloading, converting, and writing Parquet test results")
+		}
+
+		toUpdate[i] = result.ID
+	}
+
+	// TODO: Should this be a bulk update? Or should we update each
+	// individual test result doc after a successful write to the Presto
```
rtimmons, line 166: after each would be more correct but slower. The worst case scenario is that we double-migrate the remaining parts of batches. Either way is fine. Batch like here is fine imho.

units/bson_to_parquet_backfill.go
```diff
@@ -0,0 +1,183 @@
...
+
+	var results []model.TestResults
+	if err = cur.All(ctx, &results); err != nil {
+		return errors.Wrap(err, "decoding test results from temporary collection")
+	}
+
+	toUpdate := make([]string, len(results))
+	for i, result := range results {
+		if err = result.DownloadConvertAndWriteParquet(ctx); err != nil {
+			return errors.Wrap(err, "downloading, converting, and writing Parquet test results")
```
rtimmons, line 159: this will cause the whole batch to fail, ya? Then the other jobs will end up in an "orphaned" state?

rtimmons (commented):
Looks like a good first part, but I don't think I'm understanding the queue-table behavior. 

Most of my comments are primarily rhetorical, so no need to address unless you agree.

Can you add a comment about the overall flow?

E.g.

1. Create a new job. When it starts it gives itself an id in `batch_job_controller`.
2. Update N rows in test_results having no migration.id. Do one-by-one because updateMany is wrong. (Not sure what the time field is here?)
3. Do a find to get the test_results items we just updated (having our id).
4. For each, perform the migration.
5. Update the batch_job_controller with the stats for the batch.
6. Repeat. Goto 2.

Also not sure if super detailed automated testing is necessary here. This is temporary code, and we'll prove it works by comparing aggregations and spot-checks. Of course if there are any tricky parts please add testing until you're confident things are 👍🏽.

model/test_results.go
julianedwards: I was thinking the same thing, but adding this as a sub-doc to the test results docs in the temp collection is programmatically simpler and I don't see any major benefit of putting it in a separate collection.

units/bson_to_parquet_backfill.go
julianedwards: That is zero time, so basically this is looking for orphaned test results (test results that were marked for migration but the job failed before they could be marked complete).

units/bson_to_parquet_backfill.go
julianedwards: Yep! But I have a clause in the update filter above to check for test results that find themselves in this state.

units/bson_to_parquet_backfill.go
julianedwards: Yeah, if we keep the batches a reasonable size I think this should be fine.

units/bson_to_parquet_backfill.go
```diff
@@ -0,0 +1,184 @@
...
+	"github.com/mongodb/anser/bsonutil"
+	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/message"
+	"github.com/pkg/errors"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+// TODO (EVG-16138): Remove this job once we do the BSON to Parquet cutover.
```
Kimchelly, line 22: EVG-16138 is this ticket. Should this be EVG-16140 instead of EVG-16138?

units/bson_to_parquet_backfill.go
```diff
@@ -0,0 +1,184 @@
...
+		},
+	}
+}
+
+// NewBSONToParquetBackfillJob returns a new Amboy job that does batched
+// migrations of existing BSON test results to Apache Parquet test results
+// stored in the Presto bucket.
+func NewBSONToParquetBackfillJob(id string) amboy.Job {
+	j := makeBSONToParquetBackfillJob()
+	j.SetID(fmt.Sprintf("%s.%s", bsonToParquetBackfillJobName, id))
```
Kimchelly, line 55: Does this job need scopes?

units/bson_to_parquet_backfill.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	if controller.Iterations <= 0 {
+		controller.Iterations = 1
+	}
+	if controller.Timeout <= 0 {
+		controller.Timeout = time.Minute
+	}
+	j.controller = controller
+
+	var cancel context.CancelFunc
```
Kimchelly, line 83: nit: I don't think you need to explicitly declare `cancel` as a variable anymore since we removed vetshadow from the linters.

units/bson_to_parquet_backfill.go
```diff
@@ -0,0 +1,184 @@
...
+				{model.TestResultsMigrationKey: bson.M{"$exists": false}},
+				{bsonutil.GetDottedKeyName(model.TestResultsMigrationKey, model.MigrationStatsVersionKey): bson.M{"$lt": j.controller.Version}},
+				{"$and": []bson.M{
+					{bsonutil.GetDottedKeyName(model.TestResultsMigrationKey, model.MigrationStatsStartedAtKey): bson.M{"$lte": time.Now().Add(-j.controller.Timeout)}},
+					{bsonutil.GetDottedKeyName(model.TestResultsMigrationKey, model.MigrationStatsCompletedAtKey): nil},
+				}},
+			}}).
+			SetUpdate(bson.M{"$set": bson.M{
+				model.TestResultsMigrationKey: &model.MigrationStats{
+					MigratorID: j.ID(),
```
Kimchelly, line 120: Is this migration ID/started at staleness check for retrying the backfill when it gets interrupted? If so, can it be made a retryable job and set the `WaitUntil` time instead? Seems like it would cut down on some of the querying complexity and bookkeeping on the start time.

units/bson_to_parquet_backfill.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil {
+		j.AddError(errors.Wrap(err, "getting batch job controller"))
+		return
+	}
+	if controller.Iterations <= 0 {
+		controller.Iterations = 1
+	}
+	if controller.Timeout <= 0 {
+		controller.Timeout = time.Minute
+	}
```
Kimchelly, line 80: Is there a default batch size, or is it assumed that it must be set?

Kimchelly (changes requested):

units/bson_to_parquet_backfill.go
julianedwards: Good catch!

units/bson_to_parquet_backfill.go
julianedwards: Maybe, but not yet. The idea is that since there will be a lot if I/O wait time (reading and writing to the DB and S3) we can run these jobs in parallel across the app servers and adjust the batch size/ iterations accordingly to not overwhelm the DB.

units/bson_to_parquet_backfill.go
julianedwards: It is also so we can run these jobs in parallel, so there may be multiple migrator IDs doing work at the same time and we need to be able to distinguish when fetching the docs in the next step, and updating them as completed in the last step.

units/bson_to_parquet_backfill.go
```diff
@@ -0,0 +1,177 @@
...
+	return j
+}
+
+func (j *bsonToParquetBackfillJob) Run(ctx context.Context) {
+	defer j.MarkComplete()
+
+	if j.env == nil {
+		j.env = cedar.GetEnvironment()
+	}
+	if j.queue == nil {
```
Kimchelly, line 66: This is unused.

Kimchelly (approved):
LGTM mod one comment.

rtimmons (approved):

PR #443, opened 2022-03-07
EVG-16463: Pipe through Presto bucket IAM Role ARN

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-16463

Since the AWS client session will be assuming a role to access to the Presto bucket, we need to pipe through the ARN of the role.

### DIFF
--- a/model/pail_storage.go
+++ b/model/pail_storage.go
@@ -24,7 +24,7 @@ const (
 	defaultS3Region = "us-east-1"
 )
 
-// Create returns a pail Bucket backed by PailType.
+// Create returns a Pail Bucket backed by PailType.
 func (t PailType) Create(ctx context.Context, env cedar.Environment, bucket, prefix, permissions string, compress bool) (pail.Bucket, error) {
 	var b pail.Bucket
 	var err error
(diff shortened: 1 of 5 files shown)


model/pail_storage.go
```diff
@@ -82,6 +82,42 @@ func (t PailType) Create(ctx context.Context, env cedar.Environment, bucket, pre
...
+	}
+
+	switch t {
+	case PailS3:
+		opts := pail.S3Options{
+			Name:          conf.Bucket.PrestoBucket,
+			Prefix:        prefix,
+			Region:        defaultS3Region,
+			Permissions:   pail.S3Permissions(permissions),
+			Credentials:   pail.CreateAWSCredentials(conf.Bucket.AWSKey, conf.Bucket.AWSSecret, ""),
```
Kimchelly, line 101: Since the IAM role is associated with the app server host, is it still necessary to set these credentials?

model/pail_storage.go
```diff
@@ -82,6 +82,42 @@ func (t PailType) Create(ctx context.Context, env cedar.Environment, bucket, pre
...
+			AssumeRoleARN: conf.Bucket.PrestoRoleARN,
+			MaxRetries:    10,
```
Kimchelly, line 103: Should this be a default constant since it's also used for `Create`?

Kimchelly (changes requested):

model/pail_storage.go
julianedwards: The IAM role is associated with the cedar IAM user, not the app server host, in this case.

Kimchelly (approved):

PR #449, opened 2022-03-16
EVG-16139 Add DDL/Schema for cedar_test_results_landing

rtimmons:
The README seems like a fine enough place to start with this. If there's a better place in the future we can move it.

### DIFF
--- a/model/README.md
+++ b/model/README.md
@@ -1,0 +1,38 @@
+Presto SQL DDL Log
+==================
+
+Executed in March 2022:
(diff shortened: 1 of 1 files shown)


julianedwards:
@rtimmons can we close this PR since we decided not to have Cedar interact directly with Presto?

julianedwards:
@rtimmons can we close this PR since we decided not to have Cedar interact directly with Presto?

PR #465, opened 2022-07-06
EVG-17255: Add display task name to Parquet test results

julianedwards:
EVG-17255: https://jira.mongodb.org/browse/EVG-17255

We need to support display test names for daily test stats and future-proofing Presto test results (for instance, when we move off Cedar).

- Add two new optional fields, `display_task_name` and `display_task_id` to the Parquet test results struct.
- Clean up optional fields in Parquet test results struct and add tests that we correctly set them depending on if they are empty or not.
- Clean up some `omitempty` tags for BSON.


model/test_results.go
```diff
@@ -1212,13 +1234,13 @@ type ParquetTestResult struct {
 	TestName        string  `parquet:"name=test_name, type=BYTE_ARRAY, convertedtype=UTF8"`
 	DisplayTestName *string `parquet:"name=display_test_name, type=BYTE_ARRAY, convertedtype=UTF8"`
 	GroupID         *string `parquet:"name=group_id, type=BYTE_ARRAY, convertedtype=UTF8"`
-	Trial           *int32  `parquet:"name=trial, type=INT32"`
-	Status          string  `parquet:"name=status, type=BYTE_ARRAY, convertedtype=UTF8"`
+	Trial           int32   `parquet:"name=trial, type=INT32"`
+	Status          string  `parquet:"name=status, type=BYTE_ARRAY, convertedtype=UTF7"`
```
ybrill, line 1238: Is it `UTF-7`?

ybrill (commented):
what's behind the omitempty changes?

model/test_results.go
julianedwards: Good catch!

julianedwards:
I ended up switching the go parquet library to one that allows for schema evolution. I tested that this works in staging with the new and old schema.

 > what's behind the omitempty changes?

I wanted to be explicit on what are optional fields and fields that we always expect (and thus should be empty values if they are not present).

model/test_results.go
```diff
@@ -348,16 +392,35 @@ func (t *TestResults) downloadParquet(ctx context.Context) ([]TestResult, error)
...
-		return nil, errors.Wrap(err, "unmarshalling Parquet test results")
+		parquetResults = append(parquetResults, row)
+	}
+
+	if err := pr.Err(); err != nil {
+		return nil, errors.Wrap(err, "reading Parquet test results rows")
+	}
+	readerClosed = true
+	if err := pr.Close(); err != nil {
+		return nil, errors.Wrap(err, "closing Parquet reader")
```
ybrill, line 422: could it make sense to just log the error in this case? We already have the results

model/test_results.go
```diff
@@ -478,18 +486,18 @@ func (t *TestResults) GetPrestoBucket(ctx context.Context) (pail.Bucket, error)
 
 // TestResultsInfo describes information unique to a single task execution.
 type TestResultsInfo struct {
-	Project                string `bson:"project,omitempty"`
-	Version                string `bson:"version,omitempty"`
```
johndaniels, line 482: What's the purpose of dropping all the "omitempty"?

model/test_results.go
```diff
@@ -581,20 +589,32 @@ func (t TestResult) getDuration() time.Duration {
...
+		Status:         t.Status,
+		LineNum:        int32(t.LineNum),
+		TaskCreateTime: types.TimeToTIMESTAMP_MILLIS(t.TaskCreateTime.UTC(), true),
+		TestStartTime:  types.TimeToTIMESTAMP_MILLIS(t.TestStartTime.UTC(), true),
+		TestEndTime:    types.TimeToTIMESTAMP_MILLIS(t.TestEndTime.UTC(), true),
+	}
+	if t.DisplayTestName != "" {
+		result.DisplayTestName = utility.ToStringPtr(t.DisplayTestName)
 	}
+	if t.GroupID != "" {
```
johndaniels, line 666: What's this change about as well?

johndaniels (changes requested):

model/test_results.go
julianedwards: These are "required" fields, it is invalid for the to be empty.

model/test_results.go
julianedwards: Yeah these fields are optional and not always populated.

model/test_results.go
julianedwards: Good idea! I went overboard here, fixed.

model/test_results.go
```diff
@@ -231,19 +271,23 @@ func (t *TestResults) uploadParquet(ctx context.Context, results *ParquetTestRes
...
 	}()
-	pw, err := writer.NewParquetWriterFromWriter(w, new(ParquetTestResults), 1)
-	if err != nil {
-		return errors.Wrap(err, "creating new Parquet writer")
-	}
+
+	pw := floor.NewWriter(goparquet.NewFileWriter(w, goparquet.WithSchemaDefinition(parquetTestResultsSchemaDef)))
+	defer func() {
+		if !closedWriter {
+			_ = pw.Close()
```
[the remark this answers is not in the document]
julianedwards: Fixed it, I went overboard. We usually just log any errors.

model/test_results.go
johndaniels: Thanks!

johndaniels (approved):

model/test_results.go
```diff
@@ -388,10 +384,6 @@ func (t *TestResults) downloadParquet(ctx context.Context) ([]TestResult, error)
 	if err := pr.Err(); err != nil {
 		return nil, errors.Wrap(err, "reading Parquet test results rows")
 	}
-	readerClosed = true
-	if err := pr.Close(); err != nil {
```
johndaniels: Does it matter that we don't return an error if we fail to close?

johndaniels:
Does it matter that we don't return the errors on failure to close anymore? Not sure if it does.

julianedwards:
> Does it matter that we don't return the errors on failure to close anymore? Not sure if it does.

No, at that point we have either read or written the data so logging any errors is fine (errors when closing readers/writers are also extremely rare and often not very consequential).

PR #485, opened 2022-12-19
EVG-18423: update Dependabot reviewers

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-18423

Change the Dependabot PR reviewers to team review. I've historically maintained Dependabot upgrades for PLT repos. The way it works is that, once per month on the first day of the month, Dependabot will search all our dependencies in our `go.mod` that use Go semantic versioning. It would then open PRs for those modules that have an updated version available, which are then manually reviewed and merged. I used a script to manage quickly merging those that passed all CI tests.

Once this is merged, I'm also going to apply this change to all the other Go repos (aside from evergreen-ci/evergreen, which isn't using Dependabot yet) without review, since these settings are basically the same across all our repos.

Checking and merging Dependabot PRs is a fairly smooth and easy process. The only catch is that right now, some of the modules we depend on (such as aws-sdk-go and gopsutil) have released new versions that rely on go1.17 features. Even though it's no longer officially supported, we have to compile for go1.16 until [EVG-18584](https://jira.mongodb.org/browse/EVG-18584) is done. For now, I just close PRs for dependencies that require go1.17.

* Update Dependabot PR reviewers.
* Update timezone to be less ambiguous (it uses UTC by default, but it seemed better to make it explicit).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,9 +5,10 @@ updates:
     directory: "/"
     schedule:
       interval: monthly
-      time: "10:00"
+      time: "06:00"
+      timezone:  "America/New_York"
     open-pull-requests-limit: 99
     commit-message:
       prefix: "CHORE: "
     reviewers:
-      - kimchelly
+      - evg-plt


ybrill:
> I used a script to manage quickly merging those that passed all CI tests

Given we're okay with this do we need a human in the loop at all? (i.e. could running the script be automated)

Kimchelly:
I think there's value in letting devs manually review changes if they so desire. The Dependabot PRs include patch notes so you know what's changed. Also, if you're about to do a major version upgrade, I feel like that should be manually okay'd by a person, even if they pass tests.

The script for the successful ones takes me a couple minutes at most to run across all repos. I usually spend more time than that reviewing the ones that don't pass (because of flaky tests, using go1.17, etc.)

ybrill:
If it's just tagging the team it seems possible (likely?) that no one will take it up.

Maybe if we had a way to integrate it into the current process it would help it get done? Like if we had a recurring ticket in Jira to review the dependabot findings? 
Is there a way to ensure the plt channel isn't flooded by PR reminders?

(CC @johndaniels)

Kimchelly:
For the PR reminder, I don't control the Slack notification, but I imagine that GitHub doesn't have enough fine-grained options to silence particular notifications.

A couple of ideas to reduce the PR volume are 1. turning off Dependabot entirely and using our own special process to bump dependencies, 2. restricting which version updates are of interest (e.g. only update if there's a new major/minor version), or 3. reducing the `open-pull-requests-limit` (which, in my experience, might not help much given how many repos we have and how many depend on the same dependencies). I would like if it Dependabot could have "once every N months" cadence or something, but I've already put Dependabot on the least frequent cadence that it allows, which is once per month.

I imagine if we want it to be reviewed at our leisure with no notifications, we would have to turn off Dependabot and make a separate process, such as using Jira tickets and manual upgrades.

julianedwards:
> If it's just tagging the team it seems possible (likely?) that no one will take it up.

@ybrill I think that your follow-up concern would mitigate this first concern:

> Is there a way to ensure the plt channel isn't flooded by PR reminders?

The Depandabot only runs once a month and, in my experience, that the slack reminders make it hard to ignore/miss open PRs.

Another idea is to use the existing support rotation to facilitate this process—whoever is on call for PLT would be responsible for going through all the Dependabot PRs, should the bot create them.

julianedwards:
Yeah, it seems pretty low-commitment. 10-15 minutes once a month is nothing, especially if its the responsibility of the person on rotation.

johndaniels:
> If it's just tagging the team it seems possible (likely?) that no one will take it up.
> 
> Maybe if we had a way to integrate it into the current process it would help it get done? Like if we had a recurring ticket in Jira to review the dependabot findings? Is there a way to ensure the plt channel isn't flooded by PR reminders?
> 
> (CC @johndaniels)

I think these will badger us in Slack daily if they are assigned to evg-plt, right?

ybrill:
@johndaniels, that's my understanding.

johndaniels (approved):
There's some debate about how best to handle this, but switching to plt as the reviewer seems absolutely like the right thing to do for now.

PR #488, opened 2023-01-12
EVG-18728: Remove historical test data code

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-18728

With the completion of [PM-2940](https://jira.mongodb.org/browse/PM-2940) we no longer calculate historical test data on the application nor do we support the historical test data route. This removes that code.

I tested this in staging, mainly that the removal of the relevant fields from the protobuf message is backwards compatible.


makefile
```diff
@@ -99,9 +99,10 @@ htmlCoverageOutput := $(foreach target,$(testPackages),$(buildDir)/output.$(targ
 protocVersion := 3.19.3
 protoOS := $(shell uname -s | tr A-Z a-z)
 ifeq ($(protoOS),darwin)
-protoOS := osx
-endif
+protoOS := osx-x86_64
```
julianedwards, line 102: This is for M1 users.

julianedwards: It doesn't exist for this version of protoc so we have to use the `x86_64` flavor.

ybrill: ah I get it.
Why are we stuck on such an old version?

ybrill (approved):

makefile
julianedwards: We tried updating and it broke everything, I think there is a ticket tracking it. But I think the original reason was resmoke couldn't upgrade gRPC if I remember correctly (related to updating Python)—this is no longer an issue obviously but I think we shouldn't upgrade until we figure out what we are going to do with Cedar.

PR #494, opened 2023-02-08
EVG-18798: Add multi-task test results rest routes

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-18798

- Add three new routes to fetch test results, test result stats, and failed test sample by multiple (task ID, execution) pairs
- Fix pagination bug that was caused by not sorting test results that were downloaded in parallel (the order needs to be consistent for pagination to work properly)
- Write new tests for both the routes and the bug fix

### DIFF
--- a/rest/test_results_routes.go
+++ b/rest/test_results_routes.go
@@ -2,6 +2,7 @@ package rest
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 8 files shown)


rest/test_results_routes.go
```diff
@@ -26,6 +27,146 @@ const (
...
+		logFindError(err, message.Fields{
+			"request":         gimlet.GetRequestID(ctx),
+			"method":          "GET",
+			"route":           "/test_results/tasks",
+			"request_payload": h.payload,
+		})
+		return gimlet.MakeJSONErrorResponder(err)
+	}
+
+	return gimlet.NewJSONResponse(testResults)
```
ybrill, line 91: is it necessary to set `pages` on the response?

julianedwards: Not anymore: since we are now taking in a request payload (which isn't really what GET requests should do) the pagination params are part of the request body rather than the query params. The `pages` only support setting the query parameters, we would also need to set the body. After speaking with @khelif96 this isn't a problem since the Evergreen code that talks to Cedar takes care of pagination itself via the graphql queries rather than using the `next` header from the response.

ybrill: ah, I understand

khelif96:
Assuming @ybrill  thinks this makes sense it lgtm. I don't know much about cedar so I will remove my review request

PR #495, opened 2023-02-11
EVG-18181: Fix how we calculate totalWallTime.

johndaniels:
We weren't including the duration of the first operation in the totalWallTime calculation, which leads DurationTotal to be wrong in the trend charts / rollups.

### DIFF
--- a/perf/rollup_factory.go
+++ b/perf/rollup_factory.go
@@ -145,7 +145,7 @@ type documentThroughput struct{}
 
 const (
 	documentThroughputName    = "DocumentThroughput"
-	documentThroughputVersion = 2
+	documentThroughputVersion = 1
 )
 
 func (f *documentThroughput) Type() string    { return documentThroughputName }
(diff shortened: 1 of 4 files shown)


julianedwards (approved):
LGTM!

perf/rollup_factory.go
```diff
@@ -145,7 +145,7 @@ type documentThroughput struct{}
 
 const (
 	documentThroughputName    = "DocumentThroughput"
-	documentThroughputVersion = 0
+	documentThroughputVersion = 2
```
julianedwards, line 148: Nit: can't this just changed to `1`?

julianedwards (approved):

perf/rollup_factory.go
johndaniels: TBH, I would have been fine to reject the review on this. Will fix.

johndaniels: Also good catch.

PR #501, opened 2023-03-14
EVG-18798: clean up TODOs and remove old routes

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-18798

- Remove old test results fetching routes
- Clean up any other TODOs

Tested that the fetching routes work as expected and the legacy + Spruce UIs test fetching works in staging.

### DIFF
--- a/model/test_results.go
+++ b/model/test_results.go
@@ -30,7 +30,6 @@ import (
 	"github.com/pkg/errors"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
-	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
 const (
@@ -653,10 +652,6 @@ type TestResultsTaskOptions struct {
 	Execution int
 }
 
-func (opts *TestResultsTaskOptions) createFindOptions() *options.FindOptions {
-	return options.Find().SetSort(bson.D{{Key: bsonutil.GetDottedKeyName(testResultsInfoKey, testResultsInfoExecutionKey), Value: -1}})
-}
-
 func (opts *TestResultsTaskOptions) createFindQuery() bson.M {
 	return bson.M{
 		bsonutil.GetDottedKeyName(testResultsInfoKey, testResultsInfoTaskIDKey):    opts.TaskID,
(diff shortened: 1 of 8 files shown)


model/test_results.go
```diff
@@ -649,52 +649,18 @@ func (t TestResult) convertToParquet() ParquetTestResult {
...
 }
 
 func (opts *TestResultsTaskOptions) createFindOptions() *options.FindOptions {
-	findOpts := options.Find().SetSort(bson.D{{Key: bsonutil.GetDottedKeyName(testResultsInfoKey, testResultsInfoExecutionKey), Value: -1}})
-	if !opts.DisplayTask {
-		findOpts = findOpts.SetLimit(1)
-	}
-
-	return findOpts
+	return options.Find().SetSort(bson.D{{Key: bsonutil.GetDottedKeyName(testResultsInfoKey, testResultsInfoExecutionKey), Value: -1}})
```
ybrill, line 657: Given the execution will be given to us explicitly do we still need to do a sort?

rest/service.go
```diff
@@ -321,11 +321,6 @@ func (s *Service) addRoutes() {
 	s.app.AddRoute("/test_results/tasks").Version(1).Get().RouteHandler(makeGetTestResultsByTasks(s.sc))
 	s.app.AddRoute("/test_results/tasks/stats").Version(1).Get().RouteHandler(makeGetTestResultsStatsByTasks(s.sc))
 	s.app.AddRoute("/test_results/tasks/failed_sample").Version(1).Get().RouteHandler(makeGetTestResultsFailedSampleByTasks(s.sc))
-	// TODO (EVG-18798): Remove these test results routes once Spruce and
```
ybrill, line 324: Are we certain we're good to remove these? They're [documented](https://github.com/evergreen-ci/cedar/wiki/Rest-V1-Usage#id6) as being part of the API, for instance.

julianedwards: They aren't used directly by users (they go through Evergreen rest V2 to fetch test results). Since the app is not longer using them we can just remove them, I will update the documentation as well.

model/test_results.go
julianedwards: Good catch!

ybrill (approved):

PR #511, opened 2023-05-03
EVG-19750: fix time-series-update job race condition when there is both user-submitted rollups and FTDC rollups

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-19750

- Only enqueue the time-series-update job once for a performance result upon the final call to `CloseMetrics`
- Add the `measurements` to the scope for each job
- Add more better testing to the perf results RPC service code

### DIFF
--- a/rpc/internal/perf_service.go
+++ b/rpc/internal/perf_service.go
@@ -54,20 +54,7 @@ func (srv *perfService) CreateMetricSeries(ctx context.Context, result *ResultDa
 	resp.Id = record.ID
 
 	if err := record.SaveNew(ctx); err != nil {
-		return resp, newRPCError(codes.Internal, errors.Wrap(err, "saving record"))
-	}
-
-	if record.Info.Mainline && len(record.Rollups.Stats) > 0 {
-		processingJob := units.NewUpdateTimeSeriesJob(record.CreateUnanalyzedSeries())
-		err := amboy.EnqueueUniqueJob(ctx, srv.env.GetRemoteQueue(), processingJob)
-
-		if err != nil {
-			return nil, newRPCError(codes.Internal, errors.Wrapf(err, "creating signal processing job for perf result '%s'", record.ID))
-		}
-	}
-
-	if err := srv.addFTDCRollupsJob(ctx, record.ID, record.Artifacts); err != nil {
-		return resp, errors.Wrap(err, "creating ftdc rollups job")
+		return nil, newRPCError(codes.Internal, errors.Wrap(err, "saving record"))
 	}
 
 	grip.Info(message.Fields{
@@ -96,8 +83,8 @@ func (srv *perfService) AttachArtifacts(ctx context.Context, artifactData *Artif
 	resp := &MetricsResponse{}
 	resp.Id = record.ID
 
-	if err := srv.addArtifacts(ctx, record, artifactData.Artifacts); err != nil {
-		return resp, err
+	for _, a := range artifactData.Artifacts {
+		record.Artifacts = append(record.Artifacts, *a.Export())
 	}
 
 	record.Setup(srv.env)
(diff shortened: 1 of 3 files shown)


units/time_series_update.go
```diff
@@ -57,7 +58,7 @@ func NewUpdateTimeSeriesJob(series model.UnanalyzedPerformanceSeries) amboy.Job
 		series.Arguments,
 	)
 	j.SetID(fmt.Sprintf("%s.%s", baseID, time.Now().UTC()))
-	j.SetScopes([]string{baseID})
+	j.SetScopes([]string{baseID, strings.Join(series.Measurements, ",")})
```
anjani-bhat, line 61: Is there any limit to how large this string can be? Joining all metrics would probably result in a very large string.

rpc/internal/perf_service.go
```diff
@@ -228,21 +215,28 @@ func (srv *perfService) CloseMetrics(ctx context.Context, end *MetricsSeriesEnd)
...
+	// Only enqueue a new update time series job for mainline results if
+	// there are user-submitted rollups AND no event data (FTDC artifact).
+	// In the latter case, the FTDC rollups job will take care of
+	// enqueueing the update time series job.
+	if record.Info.Mainline && len(record.Rollups.Stats) > 0 && !hasEventData {
+		processingJob := units.NewUpdateTimeSeriesJob(record.CreateUnanalyzedSeries())
+		if err = amboy.EnqueueUniqueJob(ctx, srv.env.GetRemoteQueue(), processingJob); err != nil {
+			return nil, newRPCError(codes.Internal, errors.Wrapf(err, "creating signal processing job for perf result '%s'", record.ID))
+		}
 	}
```
anjani-bhat, line 232: Why is this moved from the CreateMetrics function to CloseMetrics function?

rpc/internal/perf_service.go
```diff
@@ -228,21 +215,28 @@ func (srv *perfService) CloseMetrics(ctx context.Context, end *MetricsSeriesEnd)
...
+	}
 
-func (srv *perfService) addArtifacts(ctx context.Context, record *model.PerformanceResult, artifacts []*ArtifactInfo) error {
-	for _, a := range artifacts {
-		record.Artifacts = append(record.Artifacts, *a.Export())
+	// Only enqueue a new update time series job for mainline results if
+	// there are user-submitted rollups AND no event data (FTDC artifact).
+	// In the latter case, the FTDC rollups job will take care of
+	// enqueueing the update time series job.
+	if record.Info.Mainline && len(record.Rollups.Stats) > 0 && !hasEventData {
```
johndaniels, line 227: Just to confirm, this is a deliberate decision to only include Mainline since that's all that the Signal Processing Service handles, right?

rpc/internal/perf_service.go
```diff
@@ -228,21 +215,28 @@ func (srv *perfService) CloseMetrics(ctx context.Context, end *MetricsSeriesEnd)
...
+		}
 	}
 
-	return errors.Wrap(srv.addFTDCRollupsJob(ctx, record.ID, record.Artifacts), "creating FTDC rollups job")
+	resp.Success = true
+	return resp, nil
 }
 
-func (srv *perfService) addFTDCRollupsJob(ctx context.Context, id string, artifacts []model.ArtifactInfo) error {
+func (srv *perfService) addFTDCRollupsJob(ctx context.Context, id string, artifacts []model.ArtifactInfo) (bool, error) {
```
johndaniels, line 238: Probably worth a doc comment, since it's not at all obvious what the bool here is.

rpc/internal/perf_service_test.go
```diff
@@ -345,8 +384,8 @@ func TestAttachResultData(t *testing.T) {
 			defer func() {
 				require.NoError(t, tearDownEnv(env, false))
 			}()
-			port := getPort()
 
+			port := getPort()
```
johndaniels, line 388: ?

units/time_series_update.go
johndaniels, line 61: What's the purpose of this change?

johndaniels: Does it matter that series.Measurements might have variable order?

rpc/internal/perf_service.go
```diff
@@ -96,8 +83,8 @@ func (srv *perfService) AttachArtifacts(ctx context.Context, artifactData *Artif
 	resp := &MetricsResponse{}
 	resp.Id = record.ID
 
-	if err := srv.addArtifacts(ctx, record, artifactData.Artifacts); err != nil {
-		return resp, err
+	for _, a := range artifactData.Artifacts {
```
johndaniels, line 86: It's fine for AttachArtifacts to not send updates now, since it will all be sent together in CloseMetrics. Right?

johndaniels (commented):
Mostly just want to make sure I understand the series.Measurements scope change.

units/time_series_update.go
johndaniels: FWIW it's already in an array in memory  (series.Measurements), so this is probably fine from a memory usage perspective.

rpc/internal/perf_service_test.go
julianedwards: Just moved it down a line lol. It felt like it belonged with this code block.

rpc/internal/perf_service.go
```diff
@@ -251,19 +245,19 @@ func (srv *perfService) addFTDCRollupsJob(ctx context.Context, id string, artifa
 		}
 
 		if hasEventData {
-			return newRPCError(codes.InvalidArgument, errors.New("cannot have more than one raw events artifact"))
+			return false, newRPCError(codes.InvalidArgument, errors.New("cannot have more than one raw events artifact"))
```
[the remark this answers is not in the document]
julianedwards: `hasEventData` just means that it has a time series FTDC artifact, what is considered a`RawEvents` artifact schema.

units/time_series_update.go
julianedwards: Yeah this is fine!

julianedwards: This is to prevent race conditions where new rollups are processed async from the same perf result but with different sets of measurements (usually one will be a subset of the other). If this happens, the job with the fuller set of measurements may not run since the scope is less selective without adding the measurements to it and the `analysis.processed_at` field on the perf result document would get set to a time greater than that of the `rollups.processed_at` field preventing the backfill aggregation from picking up the perf result. 

Adding this to the job scope enable two jobs with the same base ID but different measurements to get enqueued simultaneously—we will do some extra work by sending some measurements to SPS twice, but that is better than not sending them at all. I also mitigated the duplicate sending in the perf results RPC service code by enqueueing the FTDC rollups job/time-series-updated job when `CloseMetrics` is called.

julianedwards: The measurements are stored as arrays in the DB so order wouldn't be variable.

rpc/internal/perf_service.go
julianedwards: Yes, the existing behavior already does this. I only added the last condition which is to check if the FTDC job was already enqueued.

rpc/internal/perf_service.go
julianedwards: To ensure we only send it once per result and don't send duplicate notifications to SPS.

rpc/internal/perf_service.go
julianedwards: Exactly! We only want to do this once to prevent duplicating the work here.

rpc/internal/perf_service.go
julianedwards: Done!

rpc/internal/perf_service_test.go
johndaniels: Ok so it was intentional, good.

units/time_series_update.go
johndaniels: Thanks for the explanation.

johndaniels (approved):

anjani-bhat (approved):

PR #514, opened 2023-05-26
EVG-19981: Upgrade go, distros, and  mongo, remove Arch Linux

johndaniels:


cmd/init-mongo/wait_for_mongo.js
```diff
@@ -0,0 +1,31 @@
...
+	try {
+		connection = new Mongo("localhost:27017");
+	} catch(error) {
+		lastError = error;
+	}
+
+	const diff = (new Date()).getTime() - start.getTime();
+
+	if (connection) {
+		console.log("waitForMongo.js: Successfully connected to local MongoDB instance:")
```
ybrill, line 20: here the file is called `wait_for_mongo.js`

makefile
```diff
@@ -176,27 +176,29 @@ phony += mod-tidy verify-mod-tidy
...
+	rm -rf mongosh
+	mkdir -p mongosh
+	cd mongosh && curl "$(MONGOSH_URL)" -o mongosh.tgz && $(MONGOSH_DECOMPRESS) mongosh.tgz && chmod +x ./mongosh-*/bin/*
+	cd mongosh && mv ./mongosh-*/bin/* .
 get-mongodb: mongodb/.get-mongodb
 	@touch $<
+get-mongosh: mongodb/.get-mongosh
+	@touch $<
 start-mongod: mongodb/.get-mongodb
-	./mongodb/mongod --dbpath ./mongodb/db_files --port 27017 --replSet evg --oplogSize 10
```
ybrill, line 192: Is it known why the `--oplogSize` option was originally specified?

model/pail_storage.go
```diff
@@ -52,19 +52,6 @@ func (t PailType) Create(ctx context.Context, env cedar.Environment, bucket, pre
...
-
-		opts := pail.GridFSOptions{
-			Database: conf.DatabaseName,
-			Name:     bucket,
-			Prefix:   prefix,
-		}
-		b, err = pail.NewGridFSBucketWithClient(ctx, client, opts)
-		if err != nil {
-			return nil, errors.WithStack(err)
-		}
```
ybrill, line 67: Why is this being removed?

johndaniels: Because we removed GridFS from pail, so this breaks the build if you use an updated version.

ybrill: Ah, I get it.
Should we remove [these](https://github.com/evergreen-ci/cedar/blob/8e0bf9facd32bfe09460d0b84f5801204bdf87be/model/pail_storage.go#L21-L22) too?

ybrill (approved):
lgtm mod the protobuf stuff

julianedwards (approved):
LGTM protofbuf changes

PR #520, opened 2023-07-31
EVG-20455: enable the exclusion of display test names from the test name regex filter

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-20455

We need to enable not only regex matches on display test names but also on the unique test names to support fetching test logs with the graphql resolver [EVG-20443](https://jira.mongodb.org/browse/EVG-20443). This adds a new option to "exclude display names" when matching on the test name regex, the default behavior will continue to favor matching on display test name—the changes are backwards compatible.

### DIFF
--- a/model/test_results_test.go
+++ b/model/test_results_test.go
@@ -1053,12 +1053,6 @@ func TestFilterAndSortTestResults(t *testing.T) {
 			expectedCount:   4,
 		},
 		{
-			name:            "TestNameExactMatchFilter",
-			opts:            &TestResultsFilterAndSortOptions{TestName: "B test"},
-			expectedResults: results[1:2],
-			expectedCount:   1,
-		},
-		{
 			name: "TestNameRegexFilter",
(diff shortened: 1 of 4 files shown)


model/test_results_test.go
```diff
@@ -1067,9 +1061,18 @@ func TestFilterAndSortTestResults(t *testing.T) {
 			},
 			expectedCount: 2,
 		},
+		{
+			name: "TestNameExcludeDisplayNamesFilter",
+			opts: &TestResultsFilterAndSortOptions{
+				TestName:            "B test",
```
ybrill, line 1067: Could it make sense to give another one of the results a DisplayName that would have matched this filter? (as is the result will be the same whether or not `ExcludeDisplayNames` is specified)

julianedwards: Done.

ybrill (approved):

PR #523, opened 2023-11-03
DEVPROD-1950: only check for and mark analyzed perf results for the most recent 7 days

julianedwards:
[DEVPROD-1950](https://jira.mongodb.org/browse/DEVPROD-1950)

- Update `GetUnanalyzedPerformanceSeries` to look only check for unanalyzed perf results from the past 7 days
- Update `MarkPerformanceResultsAsAnalyzed` to only update results for mainline perf results from the most recent 7 days

### DIFF
--- a/model/time_series_update.go
+++ b/model/time_series_update.go
@@ -11,6 +11,8 @@ import (
 	"go.mongodb.org/mongo-driver/bson"
 )
 
+const maxTimeSeriesUpdateTime = -7 * 24 * time.Hour
+
 // PerfAnalysis contains information about when the associated performance
 // result was analyzed for change points.
 type PerfAnalysis struct {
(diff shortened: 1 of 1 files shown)


model/time_series_update.go
```diff
@@ -37,17 +37,19 @@ func (p PerformanceResultSeriesID) String() string {
...
+// timestamp.
 func MarkPerformanceResultsAsAnalyzed(ctx context.Context, env cedar.Environment, performanceResultId PerformanceResultSeriesID) error {
 	filter := bson.M{
 		bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoProjectKey):   performanceResultId.Project,
 		bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoVariantKey):   performanceResultId.Variant,
 		bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoTaskNameKey):  performanceResultId.Task,
 		bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoTestNameKey):  performanceResultId.Test,
 		bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoArgumentsKey): performanceResultId.Arguments,
+		bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoMainlineKey):  true,
+		perfCreatedAtKey: bson.M{"$gt": time.Now().Add(-7 * 24 * time.Hour)},
```
anjani-bhat, line 51: `-7 * 24 * time.Hour` could be a constant that can be reused for both cases.

anjani-bhat (approved):
LGTM - I'm wondering if it would be helpful to create a MongoDB chart that gives us info about the number of results that has `perfCreatedAtKey ` older than last 7 days. This will give us overall info about results that we are missing. [Not as part of this PR]

julianedwards:
> LGTM - I'm wondering if it would be helpful to create a MongoDB chart that gives us info about the number of results that has `perfCreatedAtKey ` older than last 7 days. This will give us overall info about results that we are missing. [Not as part of this PR]

That sounds like a good idea. cc @johndaniels something to think about for the DB.

johndaniels (approved):
LGTM

PR #527, opened 2024-02-14
DEVPROD-4679: Add env variable to turn off logging to stdout

zamj:


operations/service.go
```diff
@@ -55,6 +56,10 @@ func Service() cli.Command {
 					Name:  localQueueFlag,
 					Usage: "uses a locally-backed queue rather than MongoDB",
 				},
+				cli.BoolFlag{
+					Name:  disableLocalLoggingFlag,
+					Usage: "disable logging to stdout",
```
ybrill, line 61: Did you want to be able to set it from the environment too? (if so you could specify an `EnvVar`)

operations/setup.go
```diff
@@ -41,15 +42,17 @@ func (c *serviceConf) export() *cedar.Configuration {
...
 
-	if c.interactive {
-		senders = append(senders, send.MakeNative())
-	} else {
-		sender, err := send.MakeDefaultSystem()
-		if err != nil {
-			return nil, errors.WithStack(err)
-		}
+	if !c.disableLocalLogging {
+		if c.interactive {
```
julianedwards, line 46: I think we can ignore local logging when in interactive mode—this is used for the interactive CLI commands run manually (like loading/dumping the cedar DB config to/from a file). In which case, sending logs to stdout is useful always.

```
if c.interactive {
        ...
} else if !c.disableLocalLogging {
        ...
}
```

julianedwards (changes requested):
Do we want to capture these logs _somewhere_? They can be useful diagnosing startup errors and other types of logs that may not make it to splunk. Not sure what type of logging infra Kanopy provides in this respect.

cc @ybrill

ybrill:
It seemed to me that everything up to where the senders are configured would make it out to stdout and kanopy splunk.

julianedwards:
>It seemed to me that everything up to where the senders are configured would make it out to stdout (so like kubectl logs <pod_name>) and kanopy splunk and everything after where they're configured goes to our Splunk index.

That makes sense, it also looks like there is a fallback error logger. I think this is fine then.

operations/service.go
zamj: Good point, added

operations/setup.go
zamj: Ah gotcha. Just separated these out.

operations/setup.go
```diff
@@ -43,7 +44,7 @@ func (c *serviceConf) getSenders(ctx context.Context, conf *model.CedarConfig) (
 
 	if c.interactive {
 		senders = append(senders, send.MakeNative())
-	} else {
+	} else if !c.disableLocalLogging {
```
ybrill, line 47: It doesn't look like we're going to hit this at the moment, but if `!c.interactive` and `c.disableLocalLogging` then line 57 would panic.

zamj: Good catch! Just handled that more explicitly - an error is returned now if no senders were configured and no conf was given.

ybrill (approved):

julianedwards (approved):

PR #541, opened 2024-05-06
DEVPROD-7252 Add override info to the cedar data model

SimonEismann:
This PR adds an OverrideInfo field to the cedar data model that is empty per default and will only be set if patches are manually blessed from SPS. The OverrideInfo indicates if the patch should override mainline results, the reason behind the override, and the user that configured the override.

Things we do not need to update:
- The buildlogger, as it is deprecated
- The protobuf definitions as these only cover ResultData/ResultID and not PerformanceResultInfo

In a follow-up PR, I'll add an index on override_info.override_mainline, as this will be the only field we'll query for.

### DIFF
--- a/model/perf.go
+++ b/model/perf.go
@@ -744,14 +744,14 @@ func (r *PerformanceResults) FindOutdatedRollups(ctx context.Context, name strin
 // Also stores the information where in the timeseries the patch fits.
 type OverrideInfo struct {
 	OverrideMainline bool   `bson:"override_mainline"`
-	Order            int    `bson:"order,omitempty"`
+	BaseOrder        int    `bson:"base_order,omitempty"`
 	Reason           string `bson:"reason,omitempty"`
 	User             string `bson:"user,omitempty"`
 }
 
 var (
-	overrideInfoOverrideKey = bsonutil.MustHaveTag(OverrideInfo{}, "OverrideMainline")
-	overrideInfoOrderKey    = bsonutil.MustHaveTag(OverrideInfo{}, "Order")
-	overrideInfoReasonKey   = bsonutil.MustHaveTag(OverrideInfo{}, "Reason")
-	overrideInfoUserKey     = bsonutil.MustHaveTag(OverrideInfo{}, "User")
+	overrideInfoOverrideKey  = bsonutil.MustHaveTag(OverrideInfo{}, "OverrideMainline")
+	overrideInfoBaseOrderKey = bsonutil.MustHaveTag(OverrideInfo{}, "BaseOrder")
+	overrideInfoReasonKey    = bsonutil.MustHaveTag(OverrideInfo{}, "Reason")
+	overrideInfoUserKey      = bsonutil.MustHaveTag(OverrideInfo{}, "User")
 )


model/perf.go
```diff
@@ -334,37 +334,39 @@ func (result *PerformanceResult) Close(ctx context.Context, completedAt time.Tim
...
+	TaskName     string               `bson:"task_name,omitempty"`
+	TaskID       string               `bson:"task_id,omitempty"`
+	Execution    int                  `bson:"execution"`
+	TestName     string               `bson:"test_name,omitempty"`
+	Trial        int                  `bson:"trial"`
+	Parent       string               `bson:"parent,omitempty"`
+	Tags         []string             `bson:"tags,omitempty"`
+	Arguments    PerformanceArguments `bson:"args,omitempty"`
+	Mainline     bool                 `bson:"mainline"`
+	OverrideInfo OverrideInfo         `bson:"override_info,omitempty"`
```
SimonEismann, line 350: This line is the real change, the others are just indentation.

model/perf.go
```diff
@@ -334,37 +334,39 @@ func (result *PerformanceResult) Close(ctx context.Context, completedAt time.Tim
...
+	perfResultInfoTaskNameKey     = bsonutil.MustHaveTag(PerformanceResultInfo{}, "TaskName")
+	perfResultInfoTaskIDKey       = bsonutil.MustHaveTag(PerformanceResultInfo{}, "TaskID")
+	perfResultInfoExecutionKey    = bsonutil.MustHaveTag(PerformanceResultInfo{}, "Execution")
+	perfResultInfoTestNameKey     = bsonutil.MustHaveTag(PerformanceResultInfo{}, "TestName")
+	perfResultInfoTrialKey        = bsonutil.MustHaveTag(PerformanceResultInfo{}, "Trial")
+	perfResultInfoParentKey       = bsonutil.MustHaveTag(PerformanceResultInfo{}, "Parent")
+	perfResultInfoTagsKey         = bsonutil.MustHaveTag(PerformanceResultInfo{}, "Tags")
+	perfResultInfoArgumentsKey    = bsonutil.MustHaveTag(PerformanceResultInfo{}, "Arguments")
+	perfResultInfoMainlineKey     = bsonutil.MustHaveTag(PerformanceResultInfo{}, "Mainline")
+	perfResultInfoOverrideInfoKey = bsonutil.MustHaveTag(PerformanceResultInfo{}, "OverrideInfo")
```
SimonEismann, line 368: This line is the real change, the others are just indentation.

model/perf_test.go
```diff
@@ -501,6 +501,11 @@ func getTestPerformanceResults() (*PerformanceResult, *PerformanceResult) {
 			Tags:      []string{"tag1", "tag2", "tag3"},
 			Arguments: map[string]int32{"threads": 64},
 			Mainline:  true,
+			OverrideInfo: OverrideInfo{
+				OverrideMainline: true,
+				Reason:           "test_reason",
+				User:             "test_user",
+			},
```
SimonEismann, line 508: I've added the OverrideInfo to the test data here and as expected, this does not break anything.

SimonEismann:
I've updated the code slightly to make it write the override_mainline:false per default, keeping the other two fields optional. 

Example entry with only override_mainline:
image

Example entry with everything set:
image

zamj (approved):

rpc/internal/perf_service_test.go
```diff
@@ -284,6 +284,12 @@ func TestCreateMetricSeries(t *testing.T) {
 				assert.Equal(t, test.expectedResp.Id, resp.Id)
 				assert.Equal(t, test.expectedResp.Success, resp.Success)
 
+				conf, sess, err := cedar.GetSessionWithConfig(env)
```
julianedwards, line 287: You should be able to just use the `GetDB` func here:
```
var result model.PerformanceResult{}
require.NoError(t, env.GetDB().Collection("perf_results").FindId(resp.Id).One(&result))
assert.False(t, result.Info.OverrideInfo.OverrideMainline)
```

I would also do this check after `CloseMetrics` is called and after we check the rollups (L299) just the field doesn't get set in any of those functions.

julianedwards (changes requested):
You will probably want to update the [perf API model](https://github.com/evergreen-ci/cedar/blob/main/rest/model/perf.go#L59-L91) unless you do want expose the new field via the REST end points. If you do decide to add it, make sure to update the wiki documentation with the new fields.

model/perf.go
```diff
@@ -736,3 +738,20 @@ func (r *PerformanceResults) FindOutdatedRollups(ctx context.Context, name strin
 
 	return errors.WithStack(it.Close(ctx))
 }
+
+// OverrideInfo describes if a patch build should override mainline results,
+// the reason for the override and the person that issued the override.
+// Also stores the information where in the timeseries the patch fits.
+type OverrideInfo struct {
+	OverrideMainline bool   `bson:"override_mainline"`
+	Order            int    `bson:"order,omitempty"`
```
julianedwards, line 747: Can we make this a bit more explicit? Something like `BaseOrder`.

julianedwards (changes requested):

julianedwards:
> You will probably want to update the [perf API model](https://github.com/evergreen-ci/cedar/blob/main/rest/model/perf.go#L59-L91) unless you do want expose the new field via the REST end points. If you do decide to add it, make sure to update the wiki documentation with the new fields.

Discussed offline with @SimonEismann and he does not think it is necessary to surface this information via the REST endpoints.

model/perf.go
SimonEismann: Good catch, having two fields called order in the same db entry seems questionable :)

julianedwards (approved):

PR #544, opened 2024-05-21
DEVPROD-7253 Add a new index to the cedar DB for the overrides field

SimonEismann:
This PR adds a new index to the list of required cedar indexes.

I've tested this by adding pushing this to staging, where no indexes exist. This results in the following warning in the log:
  ```
  [cedar] 2024/05/21 14:19:30 [p=error]: missing expected DB indexes: 
  ...
  expected index '[{info.override_info.override_mainline 1}]' is missing from collection 'perf_results'
  ...
  ```
   With the index, the above line disappears.
   
Before we deploy this change, I'll manually create the index in production as well. Once the change goes in, I'll monitor that the newly deployed version does not print the above warning.

### DIFF
--- a/model/indexes.go
+++ b/model/indexes.go
@@ -114,6 +114,7 @@ func GetRequiredIndexes() []SystemIndexes {
 		{
 			Keys: bson.D{
 				{Key: bsonutil.GetDottedKeyName(perfInfoKey, perfResultInfoOverrideInfoKey, overrideInfoOverrideKey), Value: 1},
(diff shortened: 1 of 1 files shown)


julianedwards (approved):

julianedwards:
Does this need to be a compound index with `version_id`?

SimonEismann:
Updated this as a compound index with `info.override_info.override_mainline` and `info.version` based on a discussion with Julian, so we can directly get the patches with override using `distinct('info.version', {'info.override_info.override_mainline': true})`.

PR #550, opened 2024-06-21
DEVPROD-7257 Add index to support new pipeline

SimonEismann:
This PR adds the new index required for the overrides pipeline to the indexes.go file.

I've verified on staging that this matches the index we are planning to create.


zamj (approved):
LGTM - I just kicked off the index build

PR #553, opened 2024-11-20
DEVPROD-12862: Remove the performance endpoints on Cedar

zamj:


model/pail_storage_test.go
```diff
@@ -1,104 +0,0 @@
...
-	"strings"
-	"testing"
-
-	"github.com/evergreen-ci/pail"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-)
-
-func TestGetDownloadURL(t *testing.T) {
-	if skip, _ := strconv.ParseBool(os.Getenv("SKIP_INTEGRATION_TESTS")); skip {
```
johndaniels, line 19: Should pail_storage.go also be deleted?

rest/data/buildlogger_test.go
```diff
@@ -10,9 +10,43 @@ import (
...
 	"github.com/evergreen-ci/cedar/model"
 	"github.com/evergreen-ci/pail"
+	"github.com/mongodb/amboy/queue"
+	"github.com/pkg/errors"
 	"github.com/stretchr/testify/suite"
 )
 
+const testDBName = "cedar_rest_data_test"
+
+func init() {
```
johndaniels, line 20: Worrying that this needed to be moved. I'm guessing this test was depending on the setup from the other test?

johndaniels (approved):
I think I have the one suggested change (deleting one extra file) but otherwise looks good. Let's be very careful when we ship this one and make sure we don't blow up the other results stuff.

rest/data/buildlogger_test.go
zamj: Yes, exactly

model/pail_storage_test.go
zamj: It's still in use by the buildlogger stuff, but I'm not sure if that's in use at all? 

The reason I just deleted this test is because it was storing performance results in pail/s3 and it didn't seem worth it to try to recreate the test for the buildlogger system, when I wasn't even sure that was being used or actively developed on.

